PR #30, opened 2015-01-23
Simplify if() conditions on check result variables

jcfr:
Remove use of an old hack that takes advantage of the auto-dereference
behavior of the if() command to detect if a variable is defined.  The
hack has the form:

 if("${VAR} MATCHES "^${VAR}$")

where "${VAR}" is a macro argument reference.  Use if(NOT DEFINED) instead.
This also avoids warnings for CMake Policy CMP0054.

### DIFF
--- a/CMake/sitkCheckCXX11.cmake
+++ b/CMake/sitkCheckCXX11.cmake
@@ -21,7 +21,7 @@ function(sitkCXX11Test VARIABLE)
   # cache the results.
   string(MD5 cmake_cxx_flags_hash "${CMAKE_CXX_FLAGS}")
   set(cache_var "${VARIABLE}_${hash_cmake_cxx_flags_hash}")
-  if("${cache_var}" MATCHES "^${cache_var}$")
+  if(NOT DEFINED "${cache_var}")
     message(STATUS "Performing Test ${VARIABLE}")
     set(requred_definitions "${CMAKE_REQUIRED_DEFINITIONS} -D${VARIABLE}")
     try_compile(${VARIABLE}


blowekamp:
I just made a note to look into this yesterday. I created an JIRA Ticket:
https://issues.itk.org/jira/browse/SIMPLEITK-614

Thank you very much for contributing and beginning to look into the issue!

Re-running CMake 3.1.1 in my SimpleITK build dir I see the following warnings related to CMP0054:

CMake Warning (dev) at CMake/sitkCheckCXX11.cmake:24 (if):
  Policy CMP0054 is not set: Only interpret if() arguments as variables or
  keywords when unquoted.  Run "cmake --help-policy CMP0054" for policy
  details.  Use the cmake_policy command to set the policy and suppress this
  warning.

  Quoted variables like "SITK_HAS_CXX11_STATIC_ASSERT_" will no longer be
  dereferenced when the policy is set to NEW.  Since the policy is not set
  the OLD behavior will be used.

CMake Warning (dev) at CMake/UseSWIGLocal.cmake:260 (IF):
  Policy CMP0054 is not set: Only interpret if() arguments as variables or
  keywords when unquoted.  Run "cmake --help-policy CMP0054" for policy
  details.  Use the cmake_policy command to set the policy and suppress this
  warning.

  Quoted variables like "r" will no longer be dereferenced when the policy is
  set to NEW.  Since the policy is not set the OLD behavior will be used.

CMake Warning (dev) at CMake/UseSWIGLocal.cmake:220 (IF):
  Policy CMP0054 is not set: Only interpret if() arguments as variables or
  keywords when unquoted.  Run "cmake --help-policy CMP0054" for policy
  details.  Use the cmake_policy command to set the policy and suppress this
  warning.

  Quoted variables like "r" will no longer be dereferenced when the policy is
  set to NEW.  Since the policy is not set the OLD behavior will be used.

* and more from UseSWIGLocal.cmake *

I'll likely create a branch and address the other on-topic warnings too.

jcfr:
Warning like the one associated with:
```
CMake Warning (dev) at CMake/UseSWIGLocal.cmake:260 (IF):
```
 and 

```
CMake Warning (dev) at CMake/UseSWIGLocal.cmake:220 (IF):
```

are false positive and won't be an issue as soon as you set the `CMP0054` policy to `NEW`


On the other hand, the warning associated with `CMake Warning (dev) at CMake/sitkCheckCXX11.cmake:24 (if):` should be fixed by commit. That is strange.



On a side note, I also notice that in function `sitkCXX11Test` you are not setting the variable into the parent scope. See https://github.com/SimpleITK/SimpleITK/blob/0e77cf8f4163ee6362257f4a76ae6ed3b0c58872/CMake/sitkCheckCXX11.cmake#L44

Is this the expected behavior ?

jcfr:
Closing the PR. Discussion is moved into Jira.

PR #32, opened 2015-03-12
Add support for 4D images via CMake option

kaspermarstal:
This pull request adds 4D support to Image, ImageFileReader, ImageFileWriter, ImportImageFilter and implements corresponding 4D tests. All 2D/3D image tests now have a 4D counterpart except for test that use HashImageFilter and BinaryFunctors since this involves extending these filters themselves which is beyond the scope of this issue. Tests that do not involve dimension and already covered by ImageTests are also not implemented (type tests on pixel buffers etc).

Most changes are wrapped in #ifdefs. In the future, we might consider only wrapping external interface and allocators in #ifdefs and leave the internal API as clean as possible. However, a code review at this point is desirable before tighter integration.

Note that
- DualMemberFunctionFactory is not extended to 4D as only MemberFunctionFactory are used by Image and I/O classes.
- Language specific interface optimizations (e.g. extract slice/volume from 4D image in python) are not covered by this topic.
- 4D data is not available from data repo so tests uses only 4D images generated from the library itself. In the future, 4D data should be added and tests rewritten.


blowekamp:
Great! This looks like its well along the right track.

A couple line made me wonder if it'd compile and test OK, with out the 4D flag enabled.

1) The first thing I am pondering is the public Image interface. I find the name "Length" very odd for the 4th dimension. As there is already a GetSize which returns a vector, and a public constructor which takes a vector for size, I am thinking that there is not a need for additional public methods.

2) Remove the usage if the define in the Image header. Call the argument "dim4" and not length, in the non-public methods. Always have the same declaration of AllocateInternal, just conditional check the arguments in the implementation, and have the ifdef just there.

3) This is the tricky one! There are too many ifdef around RegisterMemberFunctions calls. The conditionality need to be handles in the MemberFunctionFactory. One way to do this is to use the EnableIf idiom to only actually do the registration if the dimension is supported.

kaspermarstal:
Thank you for your suggestions! They are on point.

1/2) Coming up with a name with the fourth name was indeed a struggle. Simplifying the external interface is an elegant solution.

3) Aaahhh clever. Will have a look.

kaspermarstal:
The suggestions for 1) and 2) are now incorporated. There is only _one_ ifdef now which guards the registration of 4D MemberFunctions in the member function factory. Using this approach the user is gracefully presented with a "Pixel type is not supported in 4D by " << typeid(ObjectType).name() << " **or SimpleITK compiled with SITK_4D_IMAGES set to OFF.**" simply because the function lookup fails if she tries to invoke a 4D function call from an SimpleITK installation without 4D support. Text in bold was added to error message to convey this possibility to the user.

I had a hard time coming up with a EnableIf-approach that was simpler than this, but maybe that is because I don't understand how you thought it could implemented. Could you elaborate?

blowekamp:
Looking good. There are still the additional methods in sitk::Image class which I think should be removed for 4D images, and only provide the array based methods for creating a 4D image and obtaining it's size.

I'll make a pass through this patch this week and remove those methods add verify that 4D methods aren't getting instantiated, to see if there is a need for the EnableIf with the factory.

Thanks
Brad

kaspermarstal:
Great!

kaspermarstal:
There is still one ifdef left in sitkImageFileReader.cxx related handling of error messages. One possibility is to remove this and subsequent HasMemberFunction check entirely and let the MemberFunctionFactory throw the 4D error. I leave that choice to you though.

blowekamp:
I updated the branch here:
https://github.com/blowekamp/SimpleITK/tree/SimpleITK-621_AddSupportFor4DImagesViaCMakeOption

kaspermarstal:
Looking good. The PR has been updated.

blowekamp:
You said you wanted to add some test images for IO? If there are some small ones you wish to add, please just add them to the data directory and update the tests as needed. I'll then upload the files and squash the commit.

Thanks this is almost there!

Are there any filters or other features that you are going to try to add in 4d? I am thinking the ExtractImageFilter and the SliceImageFilter would be the first ones to support the python sliced based indexing.

kaspermarstal:
Great, will add some images to the data directory and 4D I/O tests.

As for other filters, it would be very handy if the JoinSeriesImageFilter supported 4D as well. A typical use case is a user having a bunch of 3D images on which they want to do groupwise registration. If 3D images could be combined to 4D within SimpleITK we have a very complete solution. I imagine it is just a question of registering a 4-dim JoinSeriesImageFilter with the member function factory. If you are okay with this I will go ahead and add it myself.

blowekamp:
While using a 4D image for group wise registration seems a bit odd to me. It sounds reasonable to get the JoinSeries filter working for this too.

Please create a new JIRA ticket, and I'll discuss it further there.

kaspermarstal:
Added 4D dataset from official nifti website http://nifti.nimh.nih.gov/nifti-1/data. The dataset is cropped in the time dimension from 180 timepoints to 4 timepoints to minimize binary file size (6.5MB -> 143KB). 

Also added 4D I/O tests. I took the liberty to extend the HashImageFilter to 4D as well in order to be able to properly compare images in tests. A 4D test for HashImageFilter was added as well.

blowekamp:
I have uploaded the data and fixed-up the commit and pushed it to my repo here:
https://github.com/blowekamp/SimpleITK/tree/SimpleITK-621_AddSupportFor4DImagesViaCMakeOption

1) Would you be able to provide a nightly dashboard build with this option enabled?
2) I think there may be an excessive amount to churn in the patch. That is commits that do it one way to be replaced by another. Do you think you are able to rebase with some fix-ups and squashes to clean up the history? Or perhaps just squash a lot of the comments into just a couple to clean up history?

Just about there.

Brad

kaspermarstal:
Great!

1) I do not have much experience with CTest and do not quite follow. Do you want me to provide a build server that builds SimpleITK with SITK_4D_IMAGES, or do you want me to write scripts for use in your dashboard?

2) Yes of course. Is it okay if you get just a single commit? Or do you prefer to break it up in a few commits that spell out the changes more clearly?

blowekamp:
1) I was asking if you could set up a nighty dashboard to build with your configuration. I thought you may have a computer around that was mostly idle over night, and could be put to use to contribute.

There is some basic information on how to do it here:
http://www.itk.org/SimpleITKDoxygen/html/DeveloperPage.html

You will likely also need to reference ITK and CMake/CTest documentation to get the full picture

2) I would think reducing it to just a few commits with good commit messages would be ideal.

kaspermarstal:
1) Okay, sure. Might as well put my BOINC machine to good use. Will probably never find those aliens anyway.  

2) No problem, will do.

kaspermarstal:
Git history was cleaned up. All tests pass on my machines which are Ubuntu 14.10 with GCC and Clang and Mac OS X Yosemite with Clang.

blowekamp:
Looks great! Rebased onto master, fix merge conflict. Will be merging into next today.

Issue #35, opened 2015-08-28
BMPImageIO supports unsigned char only


fyh:
Hi, when I am trying to run [this SimpleGaussian.cs example](https://github.com/SimpleITK/SimpleITK-MICCAI-2011-Tutorial/blob/master/Examples/AdvancedTutorial/SimpleGaussian/SimpleGaussian.cs), I got this:
```
System.ApplicationException: Exception thrown in SimpleITK ImageFileWriter_Execute: ..\..\..\..\..\ITK\Modules\IO\BMP\sr
c\itkBMPImageIO.cxx:727:
itk::ERROR: BMPImageIO(000000E0694600E0): BMPImageIO supports unsigned char only
   在 itk.simple.ImageFileWriter.Execute(Image arg0)
```
The image I set as an input has 3 channels, please tell me how to handle this. Thank you!

The image I used is attached here:
![th](https://cloud.githubusercontent.com/assets/6197117/9550945/dd60f548-4de0-11e5-9e6e-0d0331af494c.jpg)

PR #36, opened 2015-08-28
add CastImageFilter to support output writing

fyh:
to solve issue #35


blowekamp:
Great, I was about to comment that the output image was of type float and a cast just needed to be added!

Please only make the change to correct the issue and not the other off topic style changes.

Thanks you very much for your contribution.

PR #38, opened 2015-10-10
UseCSharp.cmake: fix list expansion to string

sim590:
I've noticed a problem in `UseCSharp.cmake` that doesn't let a list to expand properly at command execution after string slash replacement. Here's the fix.


blowekamp:
Do you have an example where this statement does not work?

PR #40, opened 2016-02-25
BUG: Fix Java 1.8 javadoc linting breaking the java build

kaspermarstal:
Java 1.8 javadoc treats linting errors as build errors by default so linting should be opt-in for this version.


kaspermarstal:
Disregard this PR (bug). Will open a new one.

Issue #42, opened 2016-04-21
Wish: Expose C++ code to R


zarquon42b:
I started playing with SimpleITK and R and currently try to create displacement images with each voxel being represented by a 3D vector.
Would it be possible to expose C++ code to R to allow overcoming the slow for loops in R. That way, you could pass the R-data via Rcpp and then call the implemented functions directly (after linking the created dynlib to SimpleITK)..
Best
Stefan

blowekamp:
Hello,

I experimented with something similar to what you request in Python. Please find the topic here [1].

Basically, Something like the following code would be added to the R library:
%wrapper %{
SWIGEXPORT itk::simple::Image * convertSwigSimpleITKImage( SEXP obj )
{
  itk::simple::Image *img;
  if (SWIG_ConvertPtr(obj, (void **) &img,  SWIGTYPE_p_itk__simple__Image, ??) == -1)
    return NULL;
  return img;
}
%}

You can look at the generated SimpleITKR_wrap.cxx file for more information.

I'd expect with this interface you should be able to work the Rcpp, but I haven't used it.

If you are able to give it a try, please share the results!

[1] https://github.com/SimpleITK/SimpleITK/compare/master...blowekamp:AddCXXConvertInterfaceToPythonLibrary

zarquon42b:
Hi, thanks for your answer.
I will check it out.

zarquon42b:
Hi,
It is actually quite easy. The `extern` declaration of the SimpleITK function does the trick. I started a package and added a simple function that calls the SimpleITK code on C++ level (getting the image dimensions; see: https://github.com/zarquon42b/RcppSimpleITK/blob/master/src/test.cpp). I will try to pass indices and values as matrices and run the image filling on C++ too. Theoretically, this should work.

blowekamp:
The problem with your approach is that you are dependent on the swig generated interface, which is not a public interface that is define and is subject to change.

Explicitly declaring a Public method, as I suggest, to get a C++ sitk object from a R Swigged object is better for the long term as it ensure compatibility. This would require just one method as oppose each member function in yours.

Is there an advantage to your approach that I am missing?

zarquon42b:
OK, now I got what you meant (I have no experience with SWIG yet). I tested it and it works on my local machine (I have not pushed it yet). To make things portable, this approach requires a more tedious setup for a dependent R-package, involving configure scripts and cmake and a complete install of SimpleITK (not only the created R-package) 
[...]

PR #43, opened 2016-04-22
added a wrapper to convert a Wrapped image to itk::simple::Image

zarquon42b:
, thus the C++ interface can be accessed easily via Rcpp. You can test it by installing the package RcppSimpleITK as outlined here:
https://github.com/zarquon42b/RcppSimpleITK
and then in *R* call
```r
require(RcppSimpleITK)
example("test")
```


zarquon42b:
Fixes #42

blowekamp:
This looks great! Nice job with the test project.

We are going to need to some home test this functionality in SimpleITK. I think we will should merge this, but we need to look at adding a minimal Example which runs through this code. Also we need to document this so that others can use it.

Lastly, I think the Transform class in another candidate for this interface.

zarquon42b:
I guess this should be done like this: 

1. adding some code providing  a minimum example  (without further requirements of packages such as Rcpp to keep it simple). Returning the dimensionality of an image would be fairly easy.

2. compile the code, passing the correct parameters for linking and includes (probably similar to my configure and src/CMakeLists.txt in my package)

3. run a script that dyn.loads() it an runs R code (but the SimpleITK R-package needs to be installed for this)

However, I have no idea yet how to add this to SimpleITK.

blowekamp:
I reorganized the Examples directory so that the languages are in separate directories. You will need to rebase the topic onto the current master. Please create a small simple standalone project to test this interface in "Examples/R/ExternalInterface". After you get the basic project working I'll integrate it into the testing/build of SimpleITK.

Thanks!

zarquon42b:
I added an example package. Could you have a look at it? Can I safely assume that at the moment the example code is executed that SimpleITK_DIR is set correctly? If not, I have add the option of passing command line args providing this variable.

blowekamp:
This look pretty cool! It looks like it depends on Rpp and tries to install packages into my environment. Can a simple example be created which doesn't use Rpp?

zarquon42b:
Yes,
no problem. I decided to use Rcpp because I wanted to showcase how easy
you can switch SEXP vectors to the appropriate class. But I can provide
a simpler case without Rcpp.

Best
Stefan

zarquon42b:
[Here](https://github.com/zarquon42b/SITKR_Example/blob/master/src/test.cpp) is an example of getting the image dimension without Rcpp (less elegant), if you prefer this code, I can use this function.

zarquon42b:
Here you go: same function without Rcpp: https://github.com/SimpleITK/SimpleITK/pull/43/commits/7b93ca52f788d67a19b0bd60e7b25bacee7a3dde
and with option to parse cmake options e.g. 
`Rscript --vanilla BuildAndTestSimpleITKCxx.R -DSimpleITK_DIR=~/GIT/SimpleITK-INSTALL/lib/cmake/SimpleITK-0.10/`

blowekamp:
Great! Thank you I'll take a closer look this afternoon.

blowekamp:
It looks like it's close to working for me when I manually try to run BuildAndTestSimpleITKCxx.R from the source tree.

Can you please use the Apache 2.0 license for this R module? We don't what to be obligated to change the license for the rest of the toolkits. Thanks!

zarquon42b:
Sorry, this GPL is the default in the DESCRIPTION file template and I had not given any thought to it. I changed it accordingly.

zarquon42b:
Hi,
is there still something missing?

blowekamp:
Thanks for updating the patch. I was working on the similar feature for Python, to figure out how the CMake stuff should behave. Unfortunately some other issue have required my attention. This patch seems about right, but it didn't work for me out of the box, so I'll need to figure a few things out before it's merged.

blowekamp:
I have squashed the commits and I'm merging it into next. It is still a little rough integrating into CMake and the regular build, but the R.i interface addition looks correct. We will improve the integration with CMake while the topic is in next before merging into master.

Thank you of this outstanding addition!

zarquon42b:
Your are very welcome. I have to thank you for the great job with SimpleITK.

What issues are there that are there still remaining?

BTW: this was my motivation for doing so; http://zarquon42b.github.io/2016/04/22/DisplacementImages/

PR #53, opened 2016-07-30
COMP: Using SimpleITK from C++ need operators

hjmjohnson:
After #include <SimpleITK.h>, the ability to manipulate
images with operators should be in place.

The missing headers were added to SimpleITK.h so that
the expected behavior is supported.

for i in SimpleITK/Code/BasicFilters/include/sitk*.h; do
  nn=$(basename $i);
  grep  $nn SimpleITK/Code/Common/include/SimpleITK.h;
done

### DIFF
--- a/Code/Common/include/SimpleITK.h
+++ b/Code/Common/include/SimpleITK.h
@@ -72,6 +72,20 @@
 
 #include "sitkAdditionalProcedures.h"
 
+/* 
+ * An intentional design choice to have to explicitly include the Image
+ * operators and not get them by with the basic "SimpleITK.h".
+ * The addition of these convenience mathematical operators
+ * in C++ can be a source of confusion.
+ *
+ * Either:
+ *   1) explicity include "sitkImageOperators.h" (Recommended)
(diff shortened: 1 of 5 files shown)


blowekamp:
Hans,

Thank you for the pull request and contribution.

I can't seem to find it documented any were, or any record of the discussion on the Wiki, but it was an intentional design choice to have to explicitly include the Image operators and not get them by with the basic "SimpleITK.h". Some of the initial contributors to the design were very admit about the confusion mathematical operators add in C++.

Perhaps can improve the documentation on this issue?  Perhaps in the common conventions page (https://itk.org/SimpleITKDoxygen/html/ConventionsPage.html), or anther place you think is most relevant and appropriate?

hjmjohnson:
@blowekamp Thanks for the comments.  I have changed this, and broken it apart into separate pull requests.  

The main change made based on your comments was to preserve backwards compatibility with behavior, but to provide documentation inline with the SimpleITK.h file as a pointer for other developers. to know why the operators are not there.

PR #56, opened 2016-08-05
ENH: Add GetNumberOfPixels to Images

hjmjohnson:
This can assist with integration with other libraries,
and in simple cases it allows for using the std algorithms
more easily.

float * imgdata = img.GetDataAsFloat();
float total = std::accumulate(imgdata, imgdata+img.GetNumberOfPixels, 0);

### DIFF
--- a/Code/Common/include/sitkImage.h
+++ b/Code/Common/include/sitkImage.h
@@ -149,6 +149,17 @@ namespace simple
      * number of components for each pixel is returned.
      */
     unsigned int GetNumberOfComponentsPerPixel( void ) const;
+    /** \brief Get the number of pixels in each image
+     *
+     *  Some pixels may have more than one component, so
+     *  GetNumberOfPixels should be multiplied by
+     *  GetNumberOfComponentsPerPixel in order to calculate the
+     *  the total number of values stored congituously for the image.
+     *
+     * The GetNumberOfPixels produces the same result
+     * as GetWidth()*GetHeight()*GetDepth().  This value
+     */
+    uint64_t GetNumberOfPixels( void ) const;
 
     /** Get/Set the Origin
      * @{
--- a/Code/Common/src/sitkPimpleImageBase.hxx
+++ b/Code/Common/src/sitkPimpleImageBase.hxx
@@ -261,16 +261,21 @@ namespace itk
           return 0;
           }
 
-        typename ImageType::RegionType largestRegion = this->m_Image->GetLargestPossibleRegion();
+        const typename ImageType::RegionType & largestRegion = this->m_Image->GetLargestPossibleRegion();
         return largestRegion.GetSize(dimension);
       }
 
     virtual std::vector<unsigned int> GetSize( void ) const
       {
-        typename ImageType::RegionType largestRegion = this->m_Image->GetLargestPossibleRegion();
+        const typename ImageType::RegionType & largestRegion = this->m_Image->GetLargestPossibleRegion();
         std::vector<unsigned int> size( ImageType::ImageDimension );
 
         return sitkITKVectorToSTL<unsigned int>( largestRegion.GetSize() );
+      }
+
+    virtual uint64_t GetNumberOfPixels( void ) const
+      {
+        return this->m_Image->GetLargestPossibleRegion().GetNumberOfPixels();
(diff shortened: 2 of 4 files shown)


Code/Common/src/sitkPimpleImageBase.hxx
```diff
@@ -261,18 +261,30 @@ namespace itk
...
       {
-        typename ImageType::RegionType largestRegion = this->m_Image->GetLargestPossibleRegion();
+        const typename ImageType::RegionType & largestRegion = this->m_Image->GetLargestPossibleRegion();
         std::vector<unsigned int> size( ImageType::ImageDimension );
 
         return sitkITKVectorToSTL<unsigned int>( largestRegion.GetSize() );
       }
 
+    virtual unsigned int GetNumberOfPixels( void ) const
+      {
```
blowekamp: Please use ImageRegion::GetNumberOfPixels

Code/Common/include/sitkImage.h
```diff
@@ -188,6 +188,7 @@ namespace simple
     std::vector< double > TransformContinuousIndexToPhysicalPoint( const std::vector< double > &index) const;
 
     std::vector< unsigned int > GetSize( void ) const;
+    unsigned int GetNumberOfPixels( void ) const;
```
blowekamp: Please add documentation to clarify how this interacts with the number of components and for vector images.

blowekamp:
This looks like a good idea.

The return type be int64_t or uint64_t type. Because not all languages ( Java ) support unsigned integers. (I'll get back to you on this).

The addition of "const ImageRegion" looks good. However, since I consider the ImageRegion type to be "small" since it only contains fixed arrays, I am not sure where there is any benefit to the "return to reference" added. Perhaps it still just looks little odd and I need to get use to it still.

Code/Common/include/sitkImage.h
hjmjohnson: Done.

Code/Common/src/sitkPimpleImageBase.hxx
hjmjohnson: Done.

hjmjohnson:
@blowekamp I am changing the signature to uint64_t, but the other member functions are still using unsigned integers.  GetWidth, GetDepth, GetHeight, GetNumberOfComponents.  Will those also need to be changed?

blowekamp:
No.

Indexes are 32-bits or unsigned int, but the total number of pixels should be uint64_t. IMHO, the indices should be the natural word of the system doing the computation where as the "size" or count should be the largest. But there is challenge of trying to provide a consistent interface across systems.

Currently, there is some inconsistent use of "unsigned int"  and "uint32_t" for the index types.

Code/Common/include/sitkImage.h
```diff
@@ -149,6 +149,17 @@ namespace simple
      * number of components for each pixel is returned.
      */
     unsigned int GetNumberOfComponentsPerPixel( void ) const;
+    /** \brief Get the number of pixels in each image
```
blowekamp: Please add a new line before the Doxygen comment.

Code/Common/include/sitkImage.h
```diff
@@ -149,6 +149,17 @@ namespace simple
...
+     *
+     *  Some pixels may have more than one component, so
+     *  GetNumberOfPixels should be multiplied by
+     *  GetNumberOfComponentsPerPixel in order to calculate the
+     *  the total number of values stored congituously for the image.
+     *
+     * The GetNumberOfPixels produces the same result
+     * as GetWidth()*GetHeight()*GetDepth().  This value
```
blowekamp: The above is incorrect. For 2D images as GetDepth() is 0, and SimpleITK also has the option to support 4D images too. And the last sentence appears to be unfinished.

blowekamp:
Also a couple simple tests should be added to "Testing/Unit/sitkImageTest.cxx" under the Image.Transform block.

Please let me know if you need to to finish up this patch.

Thanks for the contributions!

blowekamp:
This patch was updated and merged here: 2fcbf38189364d74b05acdbb05aedeeb1efcea1b

PR #57, opened 2016-08-05
ENH: Ease programming the C++ interface for SimpleITK

hjmjohnson:
The C++ interface for SimpleITK should provide typedefs
for common concepts used.

### DIFF
--- a/Code/Common/include/sitkImage.h
+++ b/Code/Common/include/sitkImage.h
@@ -57,6 +57,16 @@ namespace simple
   public:
     typedef Image              Self;
 
+    //Utility typedefs for commonly used concepts used
+    //in manipulating SimpleITK images
+    typedef std::vector<uint32_t> IndexType;
+    typedef std::vector<int64_t>  SignedIndexType;
+    typedef std::vector<uint32_t> SizeType;
+    typedef std::vector<double>   OriginType;
+    typedef std::vector<double>   PointType;
+    typedef std::vector<double>   SpacingType;
+    typedef std::vector<double>   DirectionType;
+
     virtual ~Image( );
 
     /** \brief Default constructor, creates an image of size 0 */


blowekamp:
The typedef are a wider issue. In some ways, many of the initial developers wanted to see the "real" type of these items to make them less opaque. In other, I know there are currently some inconsistency with these types.

This will need additional consideration and discussion.

hjmjohnson:
@blowekamp Should this old issue be closed, or put this on a backlog project?

blowekamp:
This is a reasonable idea and it does clarify the types that are used in SimpleITK. And point out some inconsistency.

It is not clear to me if these types should be member of the Image class or just types in the itk::simpleITK namespace.

The philosophical conflict with this, is that one of the design choices for SimpleITK to was have the types used for SimpleITK be readily apparent and not hidden by typedefs. The practical usefulness comes when getting the doc info for a member or procedure in Python, I believe seeing `std::vector<uint32_t>` is more useful that `IndexType`. However, for programming and maintaining consistency the typedef would be easier.

If these typedefs are added, what would be the guidance on their usage? Should they be used in the SimpleITK interfaces primarily?

I am going to leave this topic open for discussion...

zivy:
Personally I prefer the explicit types over the typedefs, reading the code is easier and it is clear what type is expected. 

Using typedefs becomes second nature in ITK as everything is a template and it makes the length of a type name manageable and often more readable. On the other hand it often obfuscates the code. 
   
In the current use case I don't think there is an advantage to using typedef.

hjmjohnson:
@zivy @blowekamp OK.  I wanted to have parity between ITK and SimpleITK, but this is fine.  I'll cancel this pull request.

blowekamp:
For the C++ programmer, I think it is reasonable to have an additional header file to "ease the c++ programming" to get more ITK style in C++. We already have an optional header file to add overloaded operators for C++.

PR #58, opened 2016-08-05
COMP: Using SimpleITK from C++ need operators

hjmjohnson:
Add documentation, instructions, and recommended
proceedures for adding math operators for simpleITK
so that C++ more closely mirrors python syntax
behaviors for math operations on images.

### DIFF
--- a/Code/Common/include/SimpleITK.h
+++ b/Code/Common/include/SimpleITK.h
@@ -72,6 +72,20 @@
 
 #include "sitkAdditionalProcedures.h"
 
+/* 
+ * An intentional design choice to have to explicitly include the Image
+ * operators and not get them by with the basic "SimpleITK.h".
+ * The addition of these convenience mathematical operators
+ * in C++ can be a source of confusion.
+ *
+ * Either:
+ *   1) explicity include "sitkImageOperators.h" (Recommended)
+ *   2) or define USE_SimpleITK_ImageOperators BEFORE including SimpleITK.h
+ */
+#ifdef USE_SimpleITK_ImageOperators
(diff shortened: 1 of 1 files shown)


Code/Common/include/SimpleITK.h
```diff
@@ -72,6 +72,20 @@
...
+ * An intentional design choice to have to explicitly include the Image
+ * operators and not get them by with the basic "SimpleITK.h".
+ * The addition of these convenience mathematical operators
+ * in C++ can be a source of confusion.
+ *
+ * Either:
+ *   1) explicity include "sitkImageOperators.h" (Recommended)
+ *   2) or define USE_SimpleITK_ImageOperators BEFORE including SimpleITK.h
+ */
+#ifdef USE_SimpleITK_ImageOperators
```
blowekamp: Is this intended to be a CMake option, propagated to a configuration file?

The following a comparison of the proposed usage:

```
#define USE_SimpleITK_ImageOperators
#include "SimpleITK.h"
```
vs
```
#include "SimpleITK.h"
#include "sitkImageOperators.h"
```

I am not sure the `define` works so well. Perhaps this comment could just be added to a Doxygen `\file` block?

hjmjohnson: Adding the comment to the doxygen block is probably the best option.  Just make sure that a git grep "sitkImageOperators.h" matches the doxygen message.

blowekamp:
I pushed an updated patch to gerrit:
http://review.source.kitware.com/#/c/21452/

blowekamp:
topic merged into next

PR #59, opened 2016-08-05
Add scalar vector math

hjmjohnson:
This patch includes #56, #57, & #58 the only new item is the last patch.

### DIFF
--- a/Code/BasicFilters/include/sitkImageOperators.h
+++ b/Code/BasicFilters/include/sitkImageOperators.h
@@ -29,8 +29,154 @@
 #include "sitkOrImageFilter.h"
 #include "sitkXorImageFilter.h"
 
+#include <algorithm>
+
 namespace itk {
 namespace simple {
+
+    /*
+     *   This family of math operators are designed to provide a more feature rich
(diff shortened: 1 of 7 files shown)


hjmjohnson:
@blowekamp I have a need for teaching many students how to migrate code from Matlab to C++,  SimpleITK is a very nice intermediary that provides rapid conversion to C++ for testing purposes (i.e. that the algorithm produces the same results as in Matlab).  Then further performance changes can be made slowly in the places that are needed.

I would appreciate a review of the many pull requests, I made today, but this is the one that I really need.

Thanks,
Hans

blowekamp:
This is a good feature to add. And it use to be there with the MultiplyByConstant and the DivideByConstant filters, until they were deprecated in ITK.

The proposed pixel wise implementation skips many of the differentiation features of ITK operators. They are multi-threaded, and ensure the physical location of the pixels match.

I'll look into adding this with the current Multiply and Divide ImageFilters. Any idea what the elected behavior for the *= and /= operators should be?

blowekamp:
Pull request #63 supersedes this.

PR #63, opened 2016-08-16
WIP: Add scalar op vector operations.

blowekamp:
Add support for multiplying a vector image by a constant scalar. This
patch changes the itk filter instantiated for all filters which use
the BinaryFunctorFilter template, from just the same binary type
for both images and scalars, to using a "ValueType" scalar image and
image type. This is the same as the pixel type for scalar images. For
VectorImages it is the same as the component type and for complex
images it is the same as the value_type.

WIP:  This has caused some problems for the complex pixel type with
the pow and add image filters. Specifically, the Add functor type use
the NumericTraits::AccumulatorType for std::complex<float> this is
std::complex<double>, but complex<double>+complex<float> is not
supported!!!

WIP: This now instantiates the BinaryOperator filters with the
combination of ScalarImages and VectorImages. The creates bigger
object files and libraries. But it does not exposed being able to
multiply scalar images with vector images.

Restores features that were provided by the MultipleByConstanst and
similarly named filters, before they were deprecated and removed.

Change-Id: I68143cf2a03d160136c030d5c7781d1f4db47308

### DIFF
--- a/Code/BasicFilters/json/AddImageFilter.json
+++ b/Code/BasicFilters/json/AddImageFilter.json
@@ -5,7 +5,7 @@
   "constant_type" : "double",
   "number_of_inputs" : 2,
   "doc" : "Some global documentation",
-  "pixel_types" : "NonLabelPixelIDTypeList",
+  "pixel_types" : "typelist::Append< BasicPixelIDTypeList, VectorPixelIDTypeList >::Type",
   "members" : [],
   "tests" : [
     {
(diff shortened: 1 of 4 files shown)


blowekamp:
@hjmjohnson This is what I have do far on this. I was working on getting the CircleCi working before this ( that is almost there ). Please let me know if this would meet your needs.

hjmjohnson:
@blowekamp I won't be able to test until next week.  I'm preparing for a very heavy teaching load this fall right now, and I have no free time to investigate this.

I do GREATLY appreciate all you assistance with this and the other requests.

Hans

bradking:
Do: check

PR #64, opened 2016-08-30
Travis

dave3d:


blowekamp:
The CI file looks good.
Please squash these commits into a single one.
Thanks!

PR #68, opened 2016-09-15
Separate R Array functions into separate file

blowekamp:


blowekamp:
@richardbeare  What do you think of this change?

richardbeare:
Where is sitkRArray? doesn't seem to be part of this change? Am I missing something?

PR #69, opened 2016-09-15
Remove CircleCI rsync and distributed testing

blowekamp:
It appears to take longer to distribute the data that is does to run
the tests. This may allow using CTest as a driver for the main
SimpleITK-core build.

Change-Id: I8c7cd16f027885aca09ef9f7d174b396b5447df7


blowekamp:
Not using this approach

PR #73, opened 2016-11-21
R transform tests

zivy:
TranslationTransform - in R, GetOffset method doesn't seem to be wrapped (invocation causes an error).
Transform (composite transform) - in R, AddTransform method returns Self (shouldn't return it).

### DIFF
--- a/Testing/Unit/AdditionalTests.cmake
+++ b/Testing/Unit/AdditionalTests.cmake
@@ -69,6 +69,9 @@ sitk_add_r_test( SwigVectorConversion
 sitk_add_r_test( ImageSlicing
   "--file=${SimpleITK_SOURCE_DIR}/Testing/Unit/RImageSlicingTests.R"
   )
+sitk_add_r_test( Transform
+  "--file=${SimpleITK_SOURCE_DIR}/Testing/Unit/RTransformTests.R"
+  )
 
 
 #
--- a/Testing/Unit/RTransformTests.R
+++ b/Testing/Unit/RTransformTests.R
@@ -1,0 +1,22 @@
+## test that the transformation interface works in R.
+
+library(SimpleITK)
+
+translationTest <- function()
+{
+  fname <- deparse(sys.call(0)[[1]])
+
+  dimension <- 3
+  tx <- c(1,2,3)
+  translation <- TranslationTransform(dimension, tx)
+  offset <- translation$GetOffset()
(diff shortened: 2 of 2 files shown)


blowekamp:
@zivy  Thanks for the pull request for this test failure case.

zivy:
@richardbeare Is the return self from the AddTransform an error in the R code generation or do you think this is coming from the C++ layer?

richardbeare:
There's something weird going on - GetOffset is partly wrapped, but isn't recognised as a method by swig. I can't see why. Other transforms seem OK.

The self return should not happen any more - we ought to be returning void instead. I think there is a macro that is meant to be doing that. Hope to check out in more detail soon

richardbeare:
I've discovered something very strange about swig/R (at least I presume it is only swig/R) - it seems that methods with "Offset" in the name result in different behaviour - specifically they generate a "$<-" and "[[<-" function pair instead of a "$" function - this means that it is setting up code to do stuff like:

transform$GetOffset() <- x

Which makes no sense. I've not found where in swig it happens. There are about 4 classes that have this problem!

blowekamp:
Sounds like you are making progress. What are the 4 classes? We should add some tests!

richardbeare:
OK,
I have this sorted out - it is down to a dodgy part of the R binding generator in swig.

First - the classes with problems are:

``` c++
GridImageSource
FastMarchingUpwindGradientImageFilter
OtsuMultipleThresholdsImageFilter
TranslationTransform
```
They all have methods ending with "set" -_SetGridOffset/GetGridOffset_, _SetTargetOffset/GetTargetOffset_, _SetLabelOffset/GetLabelOffset_ and _SetOffset/GetOffset_

The symptom is that these methods aren't accessible using the normal \$ notation:
``` r
tf <-TranslationTransform(3, c(1,2,3))
tf$GetOffset() # error
```

The problem is that the binding generation has been searching for methods ending in "set" instead of "_set", which is what the comments say it does. Methods ending in "_set" are created for structures and classes so that values can be set. i.e. if we had a publicly visible integer variable _x_ in a class then in R we could do

``` r
V <- TheClass()
V$x <- 10
```

I've submitted a pull request with the change - hopefully it will get through. I'm testing on sitk now

Available at:

github.com/richardbeare/swig/

c90239bce6164a771e07183f208e53d692c2e70f

blowekamp:
Resolved in #82

blowekamp:
The fix for this issue happened here swig/swig#839

PR #80, opened 2016-12-06
targetLinkLibrariesWithDynamicLookup: Backport module from scikit-build

blowekamp:
Note that the name of the function remains prefixed with "sitk_"

Change-Id: I9e3eba4f3de436b297a2547c4f2a1346e6ecfec3

### DIFF
--- a/CMake/sitkLanguageOptions.cmake
+++ b/CMake/sitkLanguageOptions.cmake
@@ -10,6 +10,11 @@
 
 include(sitkTargetLinkLibrariesWithDynamicLookup)
 
+sitk_check_dynamic_lookup(MODULE
+  SHARED
+  SITK_UNDEFINED_SYMBOLS_ALLOWED
+  )
+
 #
 # Macro to set "_QUIET" and "_QUIET_LIBRARY" based on the first
 # argument being defined and true, to either REQUIRED or QUIET.
--- a/CMake/sitkTargetLinkLibrariesWithDynamicLookup.cmake
+++ b/CMake/sitkTargetLinkLibrariesWithDynamicLookup.cmake
@@ -463,7 +463,7 @@ function(sitk_check_dynamic_lookup)
     ${link_flags_var}
     )
   set(${has_dynamic_lookup_var} ${${has_dynamic_lookup_var}} PARENT_SCOPE)
-  if(NOT link_flags_var STREQUAL "unused")
+  if(NOT "x${link_flags_var}x" MATCHES "^xunusedx$")
     set(${link_flags_var} ${${link_flags_var}} PARENT_SCOPE)
   endif()
 endfunction()
--- a/CMake/sitkTargetLinkLibrariesWithDynamicLookup.cmake
+++ b/CMake/sitkTargetLinkLibrariesWithDynamicLookup.cmake
@@ -1,89 +1,580 @@
+#.rst:
 #
-# - This module provides the function
-# sitk_target_link_libraries_with_dynamic_lookup which can be used to
-# "weakly" link loadable module.
+# Public Functions
+# ^^^^^^^^^^^^^^^^
 #
-# Link a library to a target such that the symbols are resolved at
-# run-time not link-time. This should be used when compiling a
-# loadable module when the symbols should be resolve from the run-time
-# environment where the module is loaded, and not a specific system
-# library.
+# The following functions are defined:
 #
-# Specifically, for OSX it uses undefined dynamic_lookup. This is
-# simular to using "-shared" on Linux where undefined symbols are
-# ignored.
+# .. cmake:command:: target_link_libraries_with_dynamic_lookup
 #
-# Additionally, the linker is checked to see if it supports undefined
+# ::
+#
+#     target_link_libraries_with_dynamic_lookup(<Target> [<Libraries>])
+#
+#
+# Useful to "weakly" link a loadable module. For example, it should be used
+# when compiling a loadable module when the symbols should be resolve from
+# the run-time environment where the module is loaded, and not a specific
+# system library.
+#
+# Like proper linking, except that the given ``<Libraries>`` are not necessarily
+# linked. Instead, the ``<Target>`` is produced in a manner that allows for
+# symbols unresolved within it to be resolved at runtime, presumably by the
+# given ``<Libraries>``.  If such a target can be produced, the provided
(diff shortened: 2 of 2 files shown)


blowekamp:
@jcfr 
Some system and some language's dev packages don't contain the language library, and require the usage of dynamic look up in the executable for the language symbols. The manylinux docker image is one case of this. Previously the CMake variable "SITK_UNDEFINED_SYMBOLS_ALLOWED" was used to determine if the libraries were required.

https://github.com/SimpleITK/SimpleITK/blob/master/CMake/sitkLanguageOptions.cmake#L23-L27

I don't see a replacement for this variable. Any suggestions on how to update the behavior?

blowekamp:
@jcfr Any feedback or suggestions on this issue?

jcfr:
When using manylinux, the following is done:
 
  https://github.com/SimpleITK/SimpleITKPythonPackage/blob/426b971c730ac5ceaf81f87fac0142192b9031f1/scripts/internal/manylinux-build-wheels.sh#L26-L40

This allows to keep `find_package(PythonLibs)` happy while ensuring no target is linking against the python library. Simply using `QUIET` would not play the same role.

blowekamp:
What is missing from the solution of marking a find language library as QUIET, and then not linking against it?

The prior solution does not include a check to ensure linking does not occur, but it does not require custom configuration when a system or language is encountered without the libraries. The problem ( or purposeful intent ) of language development packages not including libraries has occurred several time in the real world ( outside of building on manylinux). It is necessary to continue supporting these systems output the box.

An improved solution to the existing one is welcomed, if it meets the requirements.

jcfr:
> solution of marking a find language library as QUIET, and then not linking against it?

Looking at your [previous link](https://github.com/SimpleITK/SimpleITK/blob/master/CMake/sitkLanguageOptions.cmake#L23-L27), I missed the part where it doesn't link  against it. It looks like this is done in function `sitk_target_link_libraries_with_dynamic_lookup` implemented in  [sitkTargetLinkLibrariesWithDynamicLookup.cmake#L81-L89](https://github.com/SimpleITK/SimpleITK/blob/79aecd16284586a9ac3649bf58a7e3e9e209e77f/CMake/sitkTargetLinkLibrariesWithDynamicLookup.cmake#L81-L89)

It turns out that the implementation of that [same function in scikit-build](https://github.com/scikit-build/scikit-build/blob/master/skbuild/resources/cmake/targetLinkLibrariesWithDynamicLookup.cmake#L428-L476) also conditionally links against the python library. We are good on that front.

The problem is in two locations:

* (1) [SimpleITK/Wrapping/Python/CMakeLists.txt](https://github.com/jcfr/SimpleITK/blob/8c3bfc53da7adee24e86e1140cd964dacd08f21f/Wrapping/Python/CMakeLists.txt#L17-L22) where `SITK_UNDEFINED_SYMBOLS_ALLOWED` is not set anymore.

and 

* (2) [ SimpleITKPythonPackage/CMakeLists.txt
](https://github.com/SimpleITK/SimpleITKPythonPackage/blob/426b971c730ac5ceaf81f87fac0142192b9031f1/CMakeLists.txt#L193) 
 
### Problem (1)

For the problem (1), changing that code from 

```
if ( SITK_UNDEFINED_SYMBOLS_ALLOWED )
  set( _QUIET_LIBRARY "QUIET" )
else()
  set( _QUIET_LIBRARY "REQUIRED" )
endif()
```

to 

```
check_dynamic_lookup(MODULE
  SHARED
  SITK_UNDEFINED_SYMBOLS_ALLOWED
  unused
  )
if ( SITK_UNDEFINED_SYMBOLS_ALLOWED )
  set( _QUIET_LIBRARY "QUIET" )
else()
  set( _QUIET_LIBRARY "REQUIRED" )
endif()
```
should address the issue.

Similarly to the [other CMake modules](http://scikit-build.readthedocs.io/en/latest/cmake-modules.html), I will make sure it is documented on read the docs. 

In the mean time, associated documentation is here: https://github.com/scikit-build/scikit-build/blob/d592c8084a7e0502d9da9e7aacf0e465be40cc67/skbuild/resources/cmake/targetLinkLibrariesWithDynamicLookup.cmake#L61-L92

### Problem (2)

For the problem (2) in `SimpleITKPythonPackage/CMakeLists.txt`, the code:

```

  if(NOT DEFINED PYTHON_INCLUDE_DIR
     OR NOT DEFINED PYTHON_LIBRARY
     OR NOT DEFINED PYTHON_EXECUTABLE)

    find_package ( PythonLibs REQUIRED )
    find_package ( PythonInterp REQUIRED )

  endif()
```

should be changed into 

```
  # TODO: Download  https://raw.githubusercontent.com/scikit-build/scikit-build/v0.4.0/skbuild/resources/cmake/targetLinkLibrariesWithDynamicLookup.cmake
  include(targetLinkLibrariesWithDynamicLookup)
  check_dynamic_lookup(MODULE
    SHARED
    SITK_UNDEFINED_SYMBOLS_ALLOWED
    unused
    )

  if(NOT DEFINED PYTHON_INCLUDE_DIR
       OR NOT DEFINED PYTHON_LIBRARY
       OR NOT DEFINED PYTHON_EXECUTABLE)
    
    if ( SITK_UNDEFINED_SYMBOLS_ALLOWED )
      set( _QUIET_LIBRARY "QUIET" )
    else()
      set( _QUIET_LIBRARY "REQUIRED" )
    endif()
    find_package ( PythonLibs ${_QUIET_LIBRARY})
    find_package ( PythonInterp REQUIRED )

  endif()
```

Todo:

* [ ] Generate documentation for `targetLinkLibrariesWithDynamicLookup.cmake`
* [ ] Improve `check_dynamic_lookup` so that the fourth argument `link_flags_var` is optional
* [ ] Even better, support signature `check_dynamic_lookup(SITK_UNDEFINED_SYMBOLS_ALLOWED)` that would be equivalent of calling `check_dynamic_lookup(MODULE SHARED SITK_UNDEFINED_SYMBOLS_ALLOWED unused)`

blowekamp:
Thanks!

jcfr:
https://github.com/scikit-build/scikit-build/pull/230 is adding documentation along with support for the short signature `check_dynamic_lookup(<ResultVar>)`

blowekamp:
CircleCI continuous build's configuration looks OK.

Merging into SimpleITK for nightly testing across platforms.

PR #81, opened 2016-12-08
Add CMake compiler launcher variable

blowekamp:


blowekamp:
@fbudin69500 Thanks for the documentation for the CMAKE_<LANG>_COMPILER_LAUNCHER variable.

I am trying to make similar changes to the SimpleITK SuperBuild... but running into a couple problems....

fbudin69500:
@blowekamp Nice. Just FYI, one detail I had forgotten in the patch for ITK was to verify that the CMake version was greater or equal to 3.4. CMAKE_<LANG>_COMPILER_LAUNCHER is only available starting at that version.

blowekamp:
Right, that is required for the CMake code here to work correct. I believe to maintain compatibility and support the ARG1 along with LAUNCHER is difficult, and may result in using distcc/ccache twice if both are used.

So, what do you think that if CMake<3.4 and ARG1 is set, then a CMake warning message should be displayed?

fbudin69500:
I agree that maintaining support of ARG1 would be complex (lots of "if", prone to bugs). Since using ARG1 is not the normal thing to do, I think a warning message could be shown regardless the version of CMake. The message would state that CMAKE_<LANG>_COMPILER_LAUNCHER and CMake >= 3.4 should be used.
If there is feedback concerning this warning message, a different solution could be implemented, but is most likely not necessary.

PR #82, opened 2016-12-21
Adding a test for R Translation Transform GetOffset method.

zivy:
The GetOffset method doesn't return the expected vector.

Change-Id: I7d152ce69955de049d94c1b6473cbc05102a0d36

### DIFF
--- a/Testing/Unit/RTransformTests.R
+++ b/Testing/Unit/RTransformTests.R
@@ -1,0 +1,22 @@
+## test that the transformation interface works in R.
+
+library(SimpleITK)
+
+translationTest <- function()
+{
+  fname <- deparse(sys.call(0)[[1]])
+
+  dimension <- 3
+  tx <- c(1,2,3)
+  translation <- TranslationTransform(dimension, tx)
+  offset <- translation$GetOffset()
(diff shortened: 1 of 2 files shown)


blowekamp:
Thank you for adding a test to demonstrate the issue with SWIG. This should be fixed in swig by swig/swig#839. Merging this patch is pending updating swig.

Testing/Unit/RTransformTests.R
```diff
@@ -0,0 +1,22 @@
...
+  tx <- c(1,2,3)
+  translation <- TranslationTransform(dimension, tx)
+  offset <- translation$GetOffset()
+  if (! (tx[1] == offset[1] && tx[2] == offset[2] && tx[3] == offset[3]))
+  {
+    cat("failure in", fname, "\n")
+    quit(save="no", status=1)
+   }
+}
+
```
blowekamp: Wasn't there another test in the gerrit patch set?

zivy: Yes, I just cherry picked it off of gerrit and pushed to this branch. Branch name "swigOffsetBug" is not correct for both commits but the commit messages describe the problems the tests address.

blowekamp:
topic merged into next

PR #83, opened 2016-12-21
Bug R get string vector

zivy:


blowekamp:
Great merging into next for testing on the dashboard.

PR #88, opened 2017-01-13
Initial checkin of GenerateDoc.py

dave3d:
GenerateDoc.py is a re-write, in Python, of the
GenerateDocumentation.groovy script.  I wrote this so that
we would no longer need groovy to update SimpleITK's documentation.

The script works the same as the groovy version:
  GenerateDoc.py SimpleITKClass.json ITK_XML_PATH

The script will update, in place, a SimpleITK class's JSON file.
It finds the corresponding ITK class's XML file, traverse the XML
structure and pulls any method descriptions that it finds from the XML
to the JSON.

Change-Id: I4129d7e6c59302f501bde2c329089238645bf794


blowekamp:
You can have the honors of deleting GenerateDocumentation.groovy.

blowekamp:
merging into next.

The documentation of usage can be improved. The if the command line arguments are missing the program does not immediately exit. A command line library like optparse or argparse should be used.

blowekamp:
merging whole topic into master...

blowekamp:
Topic was rebased and merged into next then master.

PR #91, opened 2017-01-24
Add seed parameter for registration point set generation

blowekamp:
Newly added to ITK, is a seed to parameter to generate the point-set
used for sampling the metric during registration. This adds the option
of having deterministic behavior or using a pseudo-random.

Change-Id: Id4b8e6c08837a56c1320d35af6df7deb25b894e6

### DIFF
--- a/Testing/Unit/sitkImageRegistrationMethodTests.cxx
+++ b/Testing/Unit/sitkImageRegistrationMethodTests.cxx
@@ -949,3 +949,37 @@ TEST_F(sitkRegistrationMethodTest, Optimizer_ScalesEstimator)
 
 
 }
+
+
+TEST_F(sitkRegistrationMethodTest, Optimizer_Sampling)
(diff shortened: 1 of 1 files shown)


blowekamp:
@zivy What do you think of this interface?

Code/Registration/include/sitkImageRegistrationMethod.h
```diff
@@ -423,11 +423,16 @@ namespace simple
 
     /** \brief Set percentage of pixels sampled for metric evaluation.
      *
+     * The seed parameter is used to seed the pseudo-random number
+     * generator used in generating the sampling set of points. If the
+     * seed parameter is 0, then the wall clock is used, otherwise the
+     * fixed seed is used for reproducible behavior.
```
zivy: In multi-level registration the seed is set only once.

PR #96, opened 2017-01-31
Update google test

blowekamp:
This is a WorkInProgress to address the issue in #92 to enable SimpleITK to work with Debian's source only installation of GoogleTest.


blowekamp:
The update to FindGTest.cmake is pushed to CMake for a pull request:
https://gitlab.kitware.com/cmake/cmake/merge_requests/454

PR #99, opened 2017-02-09
Enforce SITK_FORBID_DOWNLOADS

dave3d:
Previously, when building Documentation, if the USE_ITK_DOXYGEN_TAGS
flag was set in cmake, it would download the InsightDoxygen.tag file
from the ITK web site.  This would happen even if the
SITK_FORBID_DOWNLOADS flag was set.

This patch ensure the download will not happen if SITK_FORBID_DOWNLOADS
is set.  Also ITK_DOXYGEN_TAGFILE and ITK_DOXYGEN_ROOT_URL have been
added as cmake cache variables.  ITK_DOXYGEN_TAGFILE allows the user
to specify a local InsightDoxygen.tag file.  ITK_DOXYGEN_ROOT_URL
allows the user to specify a file or web location for the ITK Doxygen
documentation, where links to ITK from SimpleITK's documentation will
point.

### DIFF
--- a/Utilities/Doxygen/Doxygen.cmake
+++ b/Utilities/Doxygen/Doxygen.cmake
@@ -15,15 +15,30 @@ if (BUILD_DOXYGEN)
 
   option(USE_ITK_DOXYGEN_TAGS "Download ITK's Doxygen tags" ON)
 
+  set(DOXYGEN_TAGFILES_PARAMETER "")
+
   if (USE_ITK_DOXYGEN_TAGS)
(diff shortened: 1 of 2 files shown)


blowekamp:
LGTM!

dave3d:
Whew.  Finally.

blowekamp:
topic has been pushed to the stage repo and merged into next:
https://github.com/SimpleITK/SimpleITK/commit/c7caa776ab1d80bab20c3e83b6448535ddedec87

PR #102, opened 2017-02-10
Update to remove doxygen warnings

dave3d:
Doxygen was run with the "-u" flag to remove all the obsolete stuff
from doxygen.config.in.  Also DOT_FONTNAME was set to null because
FreeSans.ttf no longer comes with doxygen.


blowekamp:
merging into next...

blowekamp:
fixed in #108

PR #108, opened 2017-02-23
Removed font from Doxygen

dave3d:
The font FreeSans.ttf is no longer included in doxygen.  DOT_FONTNAME
is set to null to remove the warning.


blowekamp:
Great! Merging into next and master.

PR #110, opened 2017-02-24
Removed double backticks for Doxygen

dave3d:
Successive pairs of double backticks in the documentation were causing
Doxygen to think there was inline code.  These pairs of left quotes (and the
corresponding right quotes) where change to double quotes to
fix the problem.

Change-Id: Ibc5f9f9560bed84029a2d0155ec34ebb2a9fa103


blowekamp:
Looks great! Merging into next and master. 

Please update upstream ITK with the funny quotes to standard double quotes.

blowekamp:
Had to rebase.

PR #113, opened 2017-03-02
BUG: removed the loadPrivateTags parameter from GetGDCMSeriesFileNames.

zivy:
This parameter was not relevant for users, we always want to skip
reading the private tags when getting the series file names. Also got
rid of the separate functions in which one received a series and the
other did not (added a default value of empty string as the series).

Change-Id: Ie9f7d6119b26b8f1a1681199c9c035306eb48b57

### DIFF
--- a/Code/IO/include/sitkImageSeriesReader.h
+++ b/Code/IO/include/sitkImageSeriesReader.h
@@ -56,7 +56,7 @@ namespace itk {
        *
        * \param directory         Set the directory that contains the DICOM data set.
        * \param recursive         Recursively parse the input directory.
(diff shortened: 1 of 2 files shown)


Code/IO/include/sitkImageSeriesReader.h
```diff
@@ -84,16 +59,14 @@ namespace itk {
        * \param seriesID          Set the name that identifies a particular series.
```
blowekamp: What does the "" seriesID mean?

zivy:
Documentation updated.

PR #115, opened 2017-03-13
BUG: Comparison of image type tag (0008|0008) was incorrect.

zivy:
The string tag is required to be of even length. If it is not even
then it is padded with spaces. The specific location of the spaces is
not specified (prefix, suffix, somewhere in the middle).

Change-Id: I29533ca43628e92c68b9928a20c4b3c9fa4bdbeb


blowekamp:
Great! Thank you for digging into DICOM to find the correct solution!

PR #116, opened 2017-03-16
BUG: Check for undefined varName variable

dave3d:
On Windows we need to check that varName is defined before trying
to do a string replace operation on it.

Change-Id: I787db3b5266fcb47365e7ef7480d9c0d3ec38938


blowekamp:
Thank you for the improvement.

PR #122, opened 2017-03-21
Don't use ITK_MODULES_ENABLED for complete list of install ITK modules

blowekamp:
The ITK_MODULES_ENABLED variable may only include ITK modules enabled
during the initial configuration, and may not include install remote
or external modules.

It is preferred to use find_packages to "query" if ITK has a module
available, as a complete list of install ITK module may not exist.

Suggested-by: Matt McCormick <matt.mccormick@kitware.com>


blowekamp:
@thewtex Your suggestion seems to work.

thewtex:
:+1: Great! Thanks @blowekamp for taking the extra effort to make the changes.

blowekamp:
This patch was accidentally merged. It has been reverted.

PR #123, opened 2017-03-22
Support for downloading archive of R documentation from midas

richardbeare:
Change-Id: Ib713a817839e5122d0f8ccdf3cbcd808c2ec4d8d

This change fetches a tar.gz of automatically generated R documentation files and extracts them into the package build hierarchy.

### DIFF
--- a/Wrapping/R/CMakeLists.txt
+++ b/Wrapping/R/CMakeLists.txt
@@ -63,20 +63,33 @@ configure_file(
     "${CMAKE_CURRENT_BINARY_DIR}/Packaging/SimpleITK/DESCRIPTION"
     @ONLY )
 
-# download sample images, if allowed
+# download sample images and tar of Rd files, if allowed
 if(NOT SimpleITK_FORBID_DOWNLOADS)
   include(sitkExternalData)
 
-  file( GLOB_RECURSE content_links RELATIVE "${CMAKE_CURRENT_SOURCE_DIR}" "Packaging/SimpleITK/data/*.md5" )
+  file( GLOB_RECURSE image_links RELATIVE "${CMAKE_CURRENT_SOURCE_DIR}" "${CMAKE_CURRENT_SOURCE_DIR}/Packaging/SimpleITK/data/*.md5" )
 
-  foreach(link ${content_links})
+  file( GLOB_RECURSE doc_links RELATIVE "${CMAKE_CURRENT_SOURCE_DIR}" "${CMAKE_CURRENT_SOURCE_DIR}/Packaging/SimpleITK/man/*.md5" )
+
+  foreach(link ${image_links})
     string( REGEX REPLACE "\\.md5$" "" link ${link} )
     ExternalData_Expand_Arguments(  SimpleITKRpackageData
-      link_location
+      image_location
       DATA{${link}}
       )
-    set( COPY_DATA_COMMAND ${COPY_DATA_COMMAND} COMMAND ${CMAKE_COMMAND} -E copy ${link_location} ${CMAKE_CURRENT_BINARY_DIR}/Packaging/SimpleITK/data/ )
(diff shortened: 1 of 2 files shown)


Wrapping/R/CMakeLists.txt
```diff
@@ -67,7 +67,7 @@ configure_file(
 if(NOT SimpleITK_FORBID_DOWNLOADS)
   include(sitkExternalData)
 
-  file( GLOB_RECURSE content_links RELATIVE "${CMAKE_CURRENT_SOURCE_DIR}" "*.md5" )
+  file( GLOB_RECURSE content_links RELATIVE "${CMAKE_CURRENT_SOURCE_DIR}" "Packaging/SimpleITK/data/*.md5" )
```
blowekamp: Please Specify the last argument as "${CMAKE_CURRENT_SOURCE_DIR}/Packaging/SimpleITK/data/*.md5"

Wrapping/R/CMakeLists.txt
```diff
@@ -91,6 +91,33 @@ if(NOT SimpleITK_FORBID_DOWNLOADS)
     )
 
 endif()
+# Download tar of Rd files
+if(NOT SimpleITK_FORBID_DOWNLOADS)
+  include(sitkExternalData)
+
+  file( GLOB_RECURSE content_links RELATIVE "${CMAKE_CURRENT_SOURCE_DIR}" "Packaging/SimpleITK/man/*.md5" )
```
blowekamp: Same explicitly for the last argument here too.

Wrapping/R/CMakeLists.txt
```diff
@@ -91,6 +91,33 @@ if(NOT SimpleITK_FORBID_DOWNLOADS)
     )
 
 endif()
+# Download tar of Rd files
+if(NOT SimpleITK_FORBID_DOWNLOADS)
+  include(sitkExternalData)
```
blowekamp: This is duplication from above, I think the code block below and above should be consolidated, with just one SimpleITKRData target.

Wrapping/R/Packaging/SimpleITK/man/sitkRdfiles.tar.gz.md5
```diff
@@ -0,0 +1 @@
+a3a0acaeba6877ebf40f866d9e1c54ac
```
blowekamp: I have also uploaded the file to the ITK/SimpleITK folder.

Wrapping/R/CMakeLists.txt
richardbeare: I separated as the actions after download are different - copy vs tar extraction. Is there a good way to separate the two if the download code is common? Using filenames seems just as complicated, but feasible

richardbeare: Just realised you probably mean to use a single if statement

blowekamp: yes, I meant one if statement. You can add multiple commands to the PRE_BUILD step of the target to handle the two different types.

Wrapping/R/CMakeLists.txt
```diff
@@ -89,10 +102,13 @@ if(NOT SimpleITK_FORBID_DOWNLOADS)
     PRE_BUILD
     ${COPY_DATA_COMMAND}
```
blowekamp: Can you add ${EXTRACT_RMAN_COMMAND} here. Then remove the second add_custom_command.

I think that the PRE_BUILD command will be overwritten with the second statement without the APPEND option[1]. Simplifying, to a single add_custom_command should clear that caveat up.
[1] https://cmake.org/cmake/help/v3.0/command/add_custom_command.html

blowekamp:
Thank you I have rebased the commit onto the current master and made another pull request here #130.

This patch appears to have been based on next. Please base all pull request on master.

PR #125, opened 2017-03-22
BUG: Wrong metric value returned from Exhaustive optimizer.

zivy:
Metric value returned from the Exhaustive optimizer after registration
was the last evaluation and not the best metric value. Fixed to return
the best value.

Change-Id: I429b01da158b9c9d84f4ea5fc2dbc93b0c97d9c3


blowekamp:
Great! Thank you for the fix with a test.

PR #126, opened 2017-03-22
Fix setting ITK Doxygen tag in configuration file

blowekamp:
In the SimpleITK Doxygen configured configuration file the TAGSFILE
option was empty because the DOXYGEN_TAGFILES_PARAMETER was not being
set when it was empty.


blowekamp:
@dave3d I am not sure why this if statement was needed?

PR #128, opened 2017-03-24
Fix the indentation of See nodes in the docs

dave3d:
Previously leading spaces were chopped from the XML documentation
in all cases except for text within SimpleSect nodes of kind "See".

Now all leading spaces are chopped, and then a leading space is
inserted for SimpleSect See nodes.  This fix is needed because
Doxygen changed the format of the See nodes in the XML.


blowekamp:
Nice job.

PR #134, opened 2017-03-31
Adding support for writing DICOM series.

zivy:
The series was previously overwritten by the GDCMIO generating a
unique series/study/frame of reference ID for each image. We now allow
the user to tell GDCMIO to use the information found in the meta-data
dictionary.

Change-Id: Ifbf559d016396406d95e3e63804c8704d163dd28


Code/IO/include/sitkImageFileWriter.h
```diff
@@ -27,6 +27,10 @@
 #include <memory>
 
 namespace itk {
+
+// Forward decalaration for pointer
+class ImageIOBase;
+
```
blowekamp: Following sitk::ImageReaderBase, please add the following forward declaration here:


template<class T>
class SmartPointer;

zivy: Done

blowekamp:
This branch has been updates in #139

PR #135, opened 2017-04-01
Improved support for converting equations for R

richardbeare:
Minor change to the documentation generation that does a better job of dealing with maths. Greek
symbols now display correctly in Rstudio. More complex equations don't render correctly, but they
don't produce warnings as they used to.

An update to the midas-hosted documentation archive is coming via a second pull request.


blowekamp:
Great, thank you for fixing the generation part of the code!

@dave3d Any comments on this change?

dave3d:
Looks fine to me.

Issue #131, opened 2017-03-25
Serious memory leak in R package


richardbeare:
While images created by calls to the Image() constructors are garbage collected correctly, those returned from functions and filters are not:

```r
library(SimpleITK)

p <- replicate(5, Image(512, 512, 512, 'sitkFloat64'))
## check memory usage with ps
p <- 0
gc()
## check again - OK

## What about if we use a filter
InputIm <- Image(512, 512, 512, 'sitkFloat64')

ss <- ShiftScaleImageFilter()
ss$SetShift(1.0)
ss$SetScale(3)

pp <- replicate(5, ss$Execute(InputIm))
## check RAM
pp <- 0
gc()
## check again - memory leak

## Same problem with operations via functions?

ProcessedIms <- replicate(5, ShiftScale(InputIm, 1, 2))
## check RAM with PS
ProcessedIms <- 0
gc()
## check again - problem - doesn't change
```
Tracking this down now.

blowekamp:
@zivy Have you encountered this problem? Is it new or has is existed for a while?

zivy:
@blowekamp, @richardbeare 

I was able to reproduce the problem with R, OS X and linux (works correctly in Python). I am guessing this is a SWIG issue? The image destructor is not invoked when the image is the output of a filter. 

To see this I added a print statment to the image destructor:
std::cout<<"Deleting image"<<std::endl;

Add this at line 38 of sitkImage.cxx 

Below is the code I used:

```
library(SimpleITK)

img <-  Image(512, 512, 512, 'sitkFloat64')

# Using the object oriented filter interface
ss <- ShiftScaleImageFilter()
ss$SetShift(1.0)
ss$SetScale(3)

# the following line should show two images deleted (internal to the filter)
ss_image <- ss$Execute(img)

rm(img)
# the following line should show one deletion
gc()

rm(ss_image)
# the following line should show one deletion BUT DOESN'T
gc()

# Using the procedural filter interface
# the following line should show two deletions (internal filter images),
g_image <- GaussianSource()

rm(g_image)
# the following line should show one deletion BUT DOESN'T
gc()

```

[...]

PR #137, opened 2017-04-02
Typemap "ret" to attach C finalizer to Image and Transform objects.

richardbeare:
This deals with leaking when these objects are returned
by the procedural interface or Execute method call. Otherwise swig
is only attaching finalizers to the constructor outputs.

Change-Id: I1712bfe9efcbf55bce6f3bcdd8041a487dc14f85

### DIFF
--- a/Wrapping/R/R.i
+++ b/Wrapping/R/R.i
@@ -40,6 +40,21 @@
 }
 %}
 
+// itk::Simple::Image should always have a finalizer attached in R
+// simpleitk deals with reference counting.
+%typemap (ret) itk::simple::Image, itk::simple::Image &, itk::simple::Image &&, itk::simple::Image *, itk::simple::Image *const
+%{
+  R_RegisterCFinalizerEx(r_ans, delete_Image_void, TRUE);
+%}
+
+
+// same thing for transforms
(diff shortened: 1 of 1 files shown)


blowekamp:
This patch will close #131

richardbeare:
I think this should be a bugfix - without these additions and R session will eventually use loads of RAM because simpleitk images returned by filters, as opposed to created directly via constructors, will never be disposed.

blowekamp:
The big concern with this approach is that it specifically handles only these two types. If there are other _new_ object that are returned the the reference counting will be wrong for them too. This really seems like it the underlying problem should be address in SWIG.

However, for SimpleITK I have been unable to find other types that return _new_ object, so this patch does address our immediate needs.

blowekamp:
This patch has been rebased onto the _release_ branch #182

richardbeare:
I think that the typemaps I've used are swig's way of dealing with the problem. I suspect that there must be times when marking objects for deletion by the interpreter is the wrong thing to do, and thus swig forces the user to name the ones to be handled this way.

blowekamp:
I agree that it is SWIG's way. However, the default behavior of SWIG for all the other languages is to assume that SWIG owns the return object, and delete it when the SWIG object goes away. Case in point, when the methods were returning `Self &`,  SWIG would take owner ship and delete the object by default for all the other languages. So we have removed the `return self` when wrapping with SWIG.

I believe that R SWIG  needs to be updated to by default take owner ship of object return

PR #138, opened 2017-04-03
Add exception for trying to write DICOM with ImageSeriesWriter

blowekamp:
The default behavior of the ITK ImageSeriesWriter with DICOM does not
produce a correct DICOM series. Each file written has a different
series ID, and overrides the physical origin to (0,0,0).

Due to these problems we are disabling writing a DICOM series with
this class.


zivy:
LGTM - code + documentation and ran it - does what it should.

PR #139, opened 2017-04-03
Adding flag and example to write DICOM series

blowekamp:


zivy:
LGTM, thanks for fixing the image read.

Issue #146, opened 2017-04-05
Error with Superbuild


iMichka:
Hi. I just wanted to test last simpleITK, because the new release is coming soon. Had no time to look into this earlier, so sorry if I am a little bit late.

When building master with homebrew, I get the following error
> -- Configuring done
CMake Error in Code/Explicit/src/CMakeLists.txt:
  Target "SimpleITKExplicit" INTERFACE_INCLUDE_DIRECTORIES property contains
  path:
    "/tmp/simpleitk-20170405-63774-kll36r/sitk-build/ITK-prefix/include/ITK-4.11"
  which is prefixed in the source directory.
-- Generating done

This seems to be OK outside of homebrew. Maybe the error will give a hint of the problem.

blowekamp:
Thank you for reporting this issue. What version of CMake are you using?

iMichka:
I'm using 3.7.2. I will try to debug this today.

blowekamp:
Looking at this closer this morning I am confused by the error message.

```
"/tmp/simpleitk-20170405-63774-kll36r/sitk-build/ITK-prefix/include/ITK-4.11"
which is prefixed in the source directory.
```

The "site-build/ITK-prefix" is the ITK Superbuild **install** directory not a **source** directory**.

I could see this error message occurring if the ITK_DIR was set to the ITK **build** directory, and not the install directory.

Can you share your configuration file you used?

iMichka:
I'm annoyed now because I can't reproduce this anymore. I played a little bit around, and it went away when I built from a tag instead of master. I may have had a stale build dir somewhere.

This may explain it; the error happened during the build with homebrew, but not when I was building manually outside of the homebrew context. It is still not really clear what happened but it seems to be building.

As soon as the new release is on sourceforge, I'll update the homebrew formula.
Sorry for the wrong error report.

blowekamp:
Thank you for trying out SimpleITK, and maintaining Homebrew recipes, and showing me a CMake error I have never seen before too.

iMichka:
Spoke too fast ... the error pops up when simpleitk is being built. I thought it happened earlier in the process, so I did not see it at first.

Here is a full build log: https://gist.github.com/iMichka/4ba263d73e0c03c99f98e0f43623bd8e

And two CMakeCache files:
[CMakeCache-Main.txt](https://github.com/SimpleITK/SimpleITK/files/903999/CMakeCache-Main.txt)
[...]

PR #154, opened 2017-04-18
WIP: Address "SimpleITKExplicit" INTERFACE_INCLUDE_DIRECTORIES source

blowekamp:
This patches addresses the following CMake error:
CMake Error in Code/Explicit/src/CMakeLists.txt:
Target "SimpleITKExplicit" INTERFACE_INCLUDE_DIRECTORIES property
contains
path:
"/tmp/simpleitk-20170405-63774-kll36r/sitk-build/ITK-prefix/include/ITK-4.11"
which is prefixed in the source directory.
-- Generating done

This occurs when the build directory is a sub-directory of the source
tree and `CMAKE_INSTALL_PREFIX` is specified.

The SimpleITKExplicit library is intended to only be directly used by
other SimpleITK libraries.

### DIFF
--- a/CMake/sitkTargetUseITK.cmake
+++ b/CMake/sitkTargetUseITK.cmake
@@ -3,16 +3,26 @@
 # This function has the following form:
 #   sitk_target_use_itk(<target>
 #                       <PRIVATE|PUBLIC|INTERFACE>
+#                       [BUILD_INTERFACE]
 #                       [itk_module [itk_module [...]]])
 #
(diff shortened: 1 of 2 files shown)


blowekamp:
Possible solution to #146.

The include of ITK with the SimpleITK interface needs to be closely examined to determine correctness of this patch.

blowekamp:
This issues does not need to be addressed. As per #146, when this error is produced by CMake the installed SimpleITK libraries will not be useable.

PR #155, opened 2017-04-19
Fix ITK Targets file path in SimpleITKConfig

blowekamp:


thewtex:
This LGTM.  Eventually all ITK target file loading should be done by ITKConfig.cmake when `find_package(ITK` is called.

blowekamp:
Thanks for the look. I will assume that the existing conditional `if(NOT ITK_TARGETS_IMPORTED)` shall help so ensure this future change will still be compatible with the usage.

Issue #156, opened 2017-04-25
Improve tolerance of DiscreteGaussian filter


gut:
I'm testing SimpleITK on a POWER8 system and I noticed that the `BasicFilters_DiscreteGaussianImageFilter_short_Test` test is not passing. The situation is the following:

1) If I turn off all compiler optimizations  with `-O0`, then the test outputs a difference of `0.0` and it doesn't fail
2) If I let the default compiler optimizations with `-O3`, then the test outputs a difference of around `0.48` and the test fails.

I am taking a closer look to understand why such a discrepancy exists but then I wanted to ask why is 0.4 the threshold as I noticed that this has changed already in the past, but no reason is given in the commit:
https://github.com/SimpleITK/SimpleITK/commit/89e45ae2d26de68b4c9ed6101f94c7134dc2f69c

Would it be the solution to simply readjust this value? (Then please allow me to run several times more in order to come up with a safer value)

gut:
@blowekamp : Hi, it's been a while, but you could expose the rationale behind this change:
https://github.com/SimpleITK/SimpleITK/commit/89e45ae2d26de68b4c9ed6101f94c7134dc2f69c

blowekamp:
Hi!

I have not compiled SimpleITK for a POWER8 system. Judging by only reporting one test failing it sounds like it was quite successful! :tada:

I believe that this field was just adjusted based on the reports from the dashboard for different architectures and configurations.

I think the algorithm is operating correctly, and the tolerance could just be adjusted. But I would also verify that the fpu model for your compiler is proper to check that the computation is still precise and not taking short cuts which degrade accuracy. For example on some Intel compilers I have had to add this flag: `-fp-model precise`

Unfortunately, this RMS output is omitted on the our testing dashboard. So I manually ran this test a couple ways:

 1. x86_64 architecture in Release -> RMS: 0
 1. i686 w/ sse2 architecture in Release -> RMS: 0.3.7
 1. i686 w/ sse2 architecture in Debug -> RMS: 0.455 *failure*

What is important to know about this test is that the filter's input and output are signed short integers. While the intermediate processing is done with floats. So there is some "rounding" that need to occour, with RMS approaching 0.5 it is seeming like the rounding is done in a non-portable way.

@thewtex 
It may be the the filter should be improve to have more portable results, perhaps adding in a proper rounding in the conversion to the output is required.

thewtex:
Yes, this is likely related to floating point precision errors.

[...]

PR #157, opened 2017-04-27
Relax tolerance for DiscreteGaussianImageFilter_short test

blowekamp:
Thanks to Gustavo Serra Scalet for reporting the test failure on a
POWER8 system with `-O3` optimization on as 0.48. This was also
reproduced on intel architectures with `-m32` in Debug with a failing
RMS of 0.455.

We are relaxing to 0.5, with consideration being made for numerical
precision and this algorithm does a C++ cast for int->float resulting
in truncation during conversion.


blowekamp:
closes  #156

PR #159, opened 2017-05-01
Add Python script used to generate download plots

blowekamp:
This script was originally written as a Jupyter notebook. I creates
some useful charts from SourceForge and PyPI download statistics. The
SourceForge statistics are automatically obtained. To get the PyPI
download statistics currently a Google Big Query needs to be made, and
exported as a csv file.


zivy:
LGTM.

I have two questions with regard to the BigQuery SQL code as I'm not sure I understand it correctly:
1.  Is the date range set from current time back one year? DATE_ADD(CURRENT_TIMESTAMP(), -12, "month") . I though you queried from an arbitrary date, 01/01/2013, to current time.

2. The LIMIT 100 limits the number of rows returned, is this correct? Why this specific limit?

blowekamp:
@zivy 
 1. You are correct that the PyPI statistics reported only go back one year. There are only 390 days in the psf pypi big query data set so there are only 12 complete months there to report. SourceForge goes back to 1/1/2013.
 1. The limit was the default and sufficient, given that the entry are `group-by` the month and the SimpleITK versions the number or generated row is less than the limit.

PR #166, opened 2017-05-09
Fixed doxy2swig.py to handle multiple See issue

dave3d:
XML documentation generated by doxygen version 1.8.7 and newer
would produce seperate "See" nodes for multiple "See" commands.
Earlier versions of doxygen would group all those "See" commands
into one "See" node.

These multiple "See" nodes were showing up in the ".i" files that
swig produces for documentation of the Python, R, and Java wrappings.

This patch will only output one "See" message in the documentation
for multiple, sibling "See" nodes, making the documentation more
readable.

Also there's code for a command line flag to enable debugging
messages.

### DIFF
--- a/Utilities/GenerateDocs/doxy2swig.py
+++ b/Utilities/GenerateDocs/doxy2swig.py
@@ -35,7 +35,10 @@ import textwrap
 import sys
 import types
 import os.path
+import getopt
 
+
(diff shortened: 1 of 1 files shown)


blowekamp:
Nice! Great job on getting this detail right. 👏

PR #170, opened 2017-05-09
Print the Show command line when an error occurs.

dave3d:
The Show function trys to launch an image viewing app when called.
By default that app is ImageJ.  If there is an error, it will now
print out the command line that was used to try and invoke the
image viewer.

Change-Id: Ieefcf3c23738e4b9dca3b94d629da7169c951643


blowekamp:
nice job.

PR #172, opened 2017-05-10
Support building debug SimpleITK against release Python

blowekamp:
http://www.swig.org/Doc3.0/SWIGDocumentation.html#Python_nn12

### DIFF
--- a/Wrapping/Python/CMakeLists.txt
+++ b/Wrapping/Python/CMakeLists.txt
@@ -49,6 +49,13 @@ SWIG_add_module ( SimpleITK python
   sitkPyCommand.cxx )
 set(SWIG_MODULE_SimpleITKPython_TARGET_NAME "${SWIG_MODULE_SimpleITK_TARGET_NAME}")
 target_link_libraries( ${SWIG_MODULE_SimpleITKPython_TARGET_NAME} ${SimpleITK_LIBRARIES} )
+if(NOT PYTHON_DEBUG_LIBRARIES AND MSVC)
+  # If there is not a specified debug library use the release library,
+  # via a special sitkPython.h header.
+  target_compile_definitions(${SWIG_MODULE_SimpleITKPython_TARGET_NAME}
+    PRIVATE
+    $<$<CONFIG:Debug>:SWIG_PYTHON_INTERPRETER_NO_DEBUG>)
+endif()
 sitk_target_link_libraries_with_dynamic_lookup( ${SWIG_MODULE_SimpleITKPython_TARGET_NAME} ${PYTHON_LIBRARIES} )
 
 target_include_directories( ${SWIG_MODULE_SimpleITKPython_TARGET_NAME}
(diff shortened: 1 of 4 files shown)


blowekamp:
@lassoan This is the patch I am working on and testing (slowly) right now

lassoan:
Is this ready for testing? How can I try it?

blowekamp:
It's not working... It is still trying to link against `python35_d.lib` as opposed to the release library.

I am looking in how to specify the linkage to the release library in debug mode. I believe the following is needed:

```
target_link_libraries( SimpleTarget optimizer python35.dll debug python35.dll)
```

But that ran into some problems: scikit-build/scikit-build#255....

blowekamp:
I am trying this patch against Python35 from python.org.

I changed the problematic link line to:
```
target_link_libraries( ${SWIG_MODULE_SimpleITKPython_TARGET_NAME} optimized ${PYTHON_LIBRARY} debug ${PYTHON_LIBRARY} )
```

with:
```
PYTHON_LIBRARY: C:/Python35/libs/python35.lib
```

But I am still getting the link error:

```
  LINK : fatal error LNK1104: cannot open file 'python35_d.lib' [C:\b\SimpleITK-build\Wrapping\Python\SimpleITK_PYTHON.vcxproj]
```

With Slicer compiled in debug, what is the python `lib` file? does is have the `_d` suffix?

@jcfr @thewtex 
I am not sure where to go from here. I know some users wanted to link as I have tried with the downloaded python.org release version. It's unclear how VTK behaves in this situation.

jcfr:
@blowekamp Simply using the regular library in place of the debug on is not enough. You have to provide a modified `Python.h` (e.g `SimpleITKPython.h`) that should be included everywhere you were including Python.h.

Here are examples:

* https://github.com/commontk/PythonQt/blob/patched-6/src/PythonQtPythonInclude.h#L48

* https://github.com/Kitware/VTK/blob/master/Utilities/Python/vtkPython.h

blowekamp:
Thank you for you  assistance. That is done in this patch.

blowekamp:
OK, I see the problem in the `sitkPython.h` file.

blowekamp:
@lassoan 
The patch now works, and the SimpleITK tests are passing. You are welcome to test it. Please keep in mind that in Debug mode SimpleITK will have to be compiled with shared libraries, and Slicer has not been upgraded to sitk 1.0 yet, so there might be some other changes needed in the Slicer CMake.

jcfr:
> Slicer has not been upgraded to sitk 1.0

If sitk works with ITK4.12rc1, we should be able to update.

Do we need to always force SimpleITK to be built in Release or keeping the current options will work ? 
See https://github.com/Slicer/Slicer/blob/ccf66718aa15fc3a95df098cd629464d62ecfbd9/SuperBuild/External_SimpleITK.cmake#L93-L95

PR #175, opened 2017-05-12
Fix use of setuptools

vallsv:
Hi,

I was not able to create wheel using your `setup.py`
```
pip setup.py bdist_wheel
```

I spend few time to understand that the way used to autodetect `setuptools` is not the right one.

Here is a fix to use `setuptools` if it is available on your environment. That's a really common way in the Python world.

Now i can create wheels without any problems.

### DIFF
--- a/Wrapping/Python/Packaging/setup.py.in
+++ b/Wrapping/Python/Packaging/setup.py.in
@@ -1,8 +1,8 @@
 import sys
 
-if "setuptools" in sys.modules.keys():
+try:
     from setuptools import setup, Distribution
-else:
+except ImportError:
     from distutils.core import setup
     from distutils.dist import Distribution
 


blowekamp:
Thank you for you contributing your solution and letting us know of your problem.

We have a separate repository to make the build process and installation more Pythonic:
https://github.com/SimpleITK/SimpleITKPythonPackage

These lines were designed to only use setuptools when the setupegg.py file was used:
https://github.com/vallsv/SimpleITK/blob/c37ae89d985ee4509165b15e5d016e0931f71834/Wrapping/Python/Packaging/setupegg.py

I believe it was inspired from behavior of numpy at that time. If I recall correctly we found that many system setup_tools are old and buggy. Trying to import it when not needed resulted in buggy behavior and problematic installation.

vallsv:
`sys.modules` is anyway only containing already loaded modules. Then if this `setup.py` is an entry-point, `sys.modules` will be mostly empty.

vallsv:
Well, i see, that's not only an entry point... then good luck :-)

BTW, I follow that page https://itk.org/Wiki/SimpleITK/GettingStarted#Python_installation explaining `dist` target is available, but it's not. Maybe you can update it with your other branch.

Regards

blowekamp:
I have updated the instruction on the Wiki you linked too.

I believe you were not in the "SimpleITK-build" subdirectory of the Superbuild for the "dist" target. This would be the recommended way to create a wheel as opposed to directly using the setup.py script.

vallsv:
Thanks for the fix. I thought the `dist` target was a `setup.py` target. Now it is clear.

Issue #176, opened 2017-05-15
SITK_4D_IMAGES is broken


blowekamp:
@kaspermarstal reported
https://github.com/SimpleITK/SimpleITK/commit/adc03d972935f348d16ae16139f2fec611cd44aa#commitcomment-22107008

The recent changes to rename user exposed CMake variable to constantly have the `SimpleITK` prefix over `SITK` has broken the 4D image option.

blowekamp:
closed by #177

PR #177, opened 2017-05-15
BUG: SITK_4D_IMAGES is a preprocessor directive, not a CMake option, and should not be renamed to SimpleITK_4D_IMAGES

kaspermarstal:
SimpleITK/SimpleITK#176


blowekamp:
Great! Thank you for contributing!

 I thought there was something else that was needed too! This is not based on the a recent release branch, but a "arbitrary" patch before the release. It should still be fine to merge into the release branch after testing on next and master.

kaspermarstal:
I based it on the v1.0.0 tag. Glad it worked out anyway :)

PR #182, opened 2017-05-18
Typemap "ret" to attach C finalizer to Image and Transform objects.

blowekamp:
This deals with leaking when these objects are returned
by the procedural interface or Execute method call. Otherwise swig
is only attaching finalizers to the contructor outputs.

Change-Id: I1712bfe9efcbf55bce6f3bcdd8041a487dc14f85

### DIFF
--- a/Wrapping/R/R.i
+++ b/Wrapping/R/R.i
@@ -40,6 +40,21 @@
 }
 %}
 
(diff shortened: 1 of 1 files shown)


blowekamp:
@zivy  I have merged this topic into the next branch. Can you please verify it works with the Notebooks. Thanks!

PR #183, opened 2017-05-18
Fix show linux issues

dave3d:


blowekamp:
Great! This should work on my clean linux install with imagej 👍

PR #185, opened 2017-05-22
Fix linux show args for release

dave3d:


Code/IO/src/sitkShow.cxx
```diff
@@ -528,6 +590,8 @@ namespace itk
   std::string Macro = "";
   std::vector<std::string> CommandLine;
 
+  // is the imagej we're running a script or a binary?
+  bool ImageJScriptFlag = false;
```
blowekamp: I believe this will be marked as an unused variable on apple and windows. Think you can declare it when it is initialized below.

Code/IO/src/sitkShow.cxx
```diff
@@ -539,56 +603,10 @@ namespace itk
...
-  if (!ExecutableName.length())
-    {
-    ExecutableName = FindApplication( "", "ImageJ", debugOn );
-    }
-  if (!ExecutableName.length())
-    {
-    ExecutableName = FindApplication( "", "imagej", debugOn );
-    }
+#if not defined(__APPLE__) and not defined(_WIN32)
+  ImageJScriptFlag = itksys::SystemTools::FileHasSignature( ExecutableName.c_str(), "#!" );
```
blowekamp: Please document what you are checking for here any why.

blowekamp:
Looks great 👍

PR #192, opened 2017-05-25
Fixed C Preprocessor errors.

dave3d:
As it turns out, CPP isn't Python.

Change-Id: Ib0814f31e60975f6223a0429fbbac7069d743362


blowekamp:
Great! I'll wait to verify the build completes OK before merging.

PR #193, opened 2017-05-26
Enable MINC ITK Support

gdevenyi:
This PR enables MINC IO support in the Superbuild


blowekamp:
Thank you for initiating this pull request.

The default configuration of SimpleITK, including which modules are enabled, generally follows that from ITK. There are a variety of reasons and constraints for this. However, I believe that the MINIC module has matured nicely in ITK. I ran valgrind on the modules [1], and it appears to be free from memory defects now.  Can you begin the discussion in the ITK community to enable these modules by default?

Any ITK modules can easily be enable in the SimpleITK superbuild simply by adding a CMake variable on the command like. i.e. `cmake -DModule_ITKIOMINC:BOOL=ON .` would enable that ImageIO module for a customized version of SimpleITK.

[1] https://open.cdash.org/viewDynamicAnalysis.php?buildid=4914374

gdevenyi:
Hi @blowekamp, where do I go to begin such a discussion?

blowekamp:
The ITK mailing list:
https://itk.org/ITK/help/mailing.html

Or you can just submit a patch with itks Gerrrit.

PR #195, opened 2017-05-31
Documentation and minor additions to windows configuration

richardbeare:
to support building the R package under windows.
Quoting and path stuff added to documentation. Commits squashed

### DIFF
--- a/CMake/sitkUseGTest.cmake
+++ b/CMake/sitkUseGTest.cmake
@@ -33,7 +33,9 @@ function(_sitk_gtest_use_gtest_source)
 
   set(BUILD_GTEST                 ON )
   set(BUILD_GMOCK                 OFF)
-
+  if ("${WIN32}")
+      set(gtest_disable_pthreads      ON CACHE INTERNAL "Disable use of pthreads in gtest")
+  endif()
 
   # Must build GTest as static since EXCLUDE_FROM_ALL, would exclude
   # needed GTest shared libraries from being installed.
--- a/Wrapping/R/CMakeLists.txt
+++ b/Wrapping/R/CMakeLists.txt
@@ -30,6 +30,10 @@ set_source_files_properties(${swig_generated_file_fullname} PROPERTIES COMPILE_F
 # set the output directory for the R library to the binary packaging location
 set_target_properties( ${SWIG_MODULE_SimpleITK_TARGET_NAME} PROPERTIES LIBRARY_OUTPUT_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/Packaging/SimpleITK/src/)
 
+if (WIN32)
+	set_target_properties( ${SWIG_MODULE_SimpleITK_TARGET_NAME} PROPERTIES LINK_FLAGS "-s -shared -static-libgcc -static-libstdc++ -static -lpthread" )
+endif()
+
 sitk_strip_target( ${SWIG_MODULE_SimpleITK_TARGET_NAME} )
 set(SWIG_MODULE_SimpleITKR_TARGET_NAME "${SWIG_MODULE_SimpleITK_TARGET_NAME}")
 
@@ -98,6 +102,6 @@ add_custom_command( TARGET ${SWIG_MODULE_SimpleITK_TARGET_NAME}
   COMMAND ${CMAKE_COMMAND} -E make_directory ${CMAKE_CURRENT_BINARY_DIR}/R_libs
   COMMAND ${CMAKE_COMMAND} -E copy ${CMAKE_CURRENT_BINARY_DIR}/SimpleITK.R ${CMAKE_CURRENT_BINARY_DIR}/Packaging/SimpleITK/R/
   # install for running tests and create binary package
-  COMMAND ${R_COMMAND} CMD INSTALL --build ${CMAKE_CURRENT_BINARY_DIR}/Packaging/SimpleITK --library=${CMAKE_CURRENT_BINARY_DIR}/R_libs
+  COMMAND ${R_COMMAND} CMD INSTALL --build ${CMAKE_CURRENT_BINARY_DIR}/Packaging/SimpleITK --library=${CMAKE_CURRENT_BINARY_DIR}/R_libs --no-multiarch
   COMMENT "Installing R package for testing and building binary version for distribution"
   )
--- a/Wrapping/R/CMakeLists.txt
+++ b/Wrapping/R/CMakeLists.txt
@@ -31,7 +31,7 @@ set_source_files_properties(${swig_generated_file_fullname} PROPERTIES COMPILE_F
 set_target_properties( ${SWIG_MODULE_SimpleITK_TARGET_NAME} PROPERTIES LIBRARY_OUTPUT_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/Packaging/SimpleITK/src/)
 
 if (WIN32)
-	set_target_properties( ${SWIG_MODULE_SimpleITK_TARGET_NAME} PROPERTIES LINK_FLAGS "-s -shared -static-libgcc -static-libstdc++ -static -lpthread" )
+  set_target_properties( ${SWIG_MODULE_SimpleITK_TARGET_NAME} PROPERTIES LINK_FLAGS ${R_WIN_LINK_FLAGS} )
 endif()
 
 sitk_strip_target( ${SWIG_MODULE_SimpleITK_TARGET_NAME} )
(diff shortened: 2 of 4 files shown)


CMake/sitkUseGTest.cmake
```diff
@@ -33,7 +33,9 @@ function(_sitk_gtest_use_gtest_source)
 
   set(BUILD_GTEST                 ON )
   set(BUILD_GMOCK                 OFF)
-
+  if ("${WIN32}")
+      set(gtest_disable_pthreads      ON CACHE INTERNAL "Disable use of pthreads in gtest")
+  endif()
```
blowekamp: This change has already been integrated into the release branch. Please rebase your topic onto the current master, and do a force push.

```
git checkout RWindowsBattlesSquash
git fetch origin
git rebase origin/master
git push -f github_richardbeare HEAD
```

Wrapping/R/CMakeLists.txt
```diff
@@ -98,6 +102,6 @@ add_custom_command( TARGET ${SWIG_MODULE_SimpleITK_TARGET_NAME}
   COMMAND ${CMAKE_COMMAND} -E make_directory ${CMAKE_CURRENT_BINARY_DIR}/R_libs
   COMMAND ${CMAKE_COMMAND} -E copy ${CMAKE_CURRENT_BINARY_DIR}/SimpleITK.R ${CMAKE_CURRENT_BINARY_DIR}/Packaging/SimpleITK/R/
   # install for running tests and create binary package
-  COMMAND ${R_COMMAND} CMD INSTALL --build ${CMAKE_CURRENT_BINARY_DIR}/Packaging/SimpleITK --library=${CMAKE_CURRENT_BINARY_DIR}/R_libs
+  COMMAND ${R_COMMAND} CMD INSTALL --build ${CMAKE_CURRENT_BINARY_DIR}/Packaging/SimpleITK --library=${CMAKE_CURRENT_BINARY_DIR}/R_libs --no-multiarch
```
blowekamp: The way SimpleITK is build this should always be true.

Wrapping/R/CMakeLists.txt
```diff
@@ -30,6 +30,10 @@ set_source_files_properties(${swig_generated_file_fullname} PROPERTIES COMPILE_F
 # set the output directory for the R library to the binary packaging location
 set_target_properties( ${SWIG_MODULE_SimpleITK_TARGET_NAME} PROPERTIES LIBRARY_OUTPUT_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/Packaging/SimpleITK/src/)
 
+if (WIN32)
+	set_target_properties( ${SWIG_MODULE_SimpleITK_TARGET_NAME} PROPERTIES LINK_FLAGS "-s -shared -static-libgcc -static-libstdc++ -static -lpthread" )
+endif()
```
blowekamp: The problem with adding these flags like this is that they will aways be set when building SimpleITK R, and the user can not see them nor override them. I think these flags would better go into the SimpleITKRInstaller.

Wrapping/R/CMakeLists.txt
richardbeare: This is for the windows R package build process. By default R is looking for multi-architecture packages and will fail without this flag (Just confirmed).

blowekamp: OK, I am going to pull this change out and place it on the release branch.

blowekamp: see #202

Wrapping/R/CMakeLists.txt
richardbeare: I've had no luck setting these flags from the Superbuild for simpleitk. Any suggestions?

richardbeare: I can't see a way of setting these flags from cmake without adding an additional variable, like a "WIN_R_LINK_FLAGS". There doesn't appear to be an easy -Dsomething option for properties. How would you like me to do it?

blowekamp: I'll look at these closer. Why are these flags needed? Is it because the recommended mingw, version is going to have incompatible versions of the libraries with the R run-time. This looks like some pretty sketchy stuff. I can see having different versions of stdc++, but mixing different version of pthread, and libgcc seems like it will cause problems sooner or later.

Does setting the CMAKE_*_LINKER_FLAGS to these values work?

blowekamp: It not clear to me if this line is expected to link to libpthread statically or not. Linking to pthreads statically is likely a very bad idea. Does setting  CMAKE_*_LINKER_FLAGS to `-s -static-libstdc++` suffice?

richardbeare: I went around in circles getting this to work. Here's what I remember.
Setting CMAKE_*_LINKER_FLAGS didn't seem to do the job, but I'm not sure why. That was what I tried first. I think they may have broken other tests.

the -static flag gets set by the normal package build process (which we aren't using), and I think the normal version of mingw in the RTools package must have different defaults set up causing the appropriate libraries to be linked. If I don't have the std libraries linked statically I get missing symbols.

I wasn't happy with the pthread situation either. There doesn't seem to be any standard way of dealing with this under R/windows that I can see.

blowekamp:
What about setting `CMAKE_THREAD_LIBS_INIT` to the full path to the static pthread library?

I would expect setting all the `CMAKE_*_LINKER_FLAGS` with `-static-libgcc -static-libstdc++` should work.

richardbeare:
Will give it a try

richardbeare:
Is CMAKE_THREAD_LIBS_INIT from FindThreads? I couldn't see it being used anywhere yet. If I set it via command line I get the warning "Manually specified variables are not used by the project:"

blowekamp:
I think you need to add it here so that it will be propagated in the superbuild:
https://github.com/SimpleITK/SimpleITK/blob/next/SuperBuild/SuperBuild.cmake#L229

CMake/sitkUseGTest.cmake
richardbeare: Something odd is going on. When I test the release branch (which has an if (MINGW) branch rather than a WIN32), I get the compilation error with GTest. However the problem goes away if I use the WIN32 check.

This seems strange as the compiler appears to be detected as MINGW in the simpleitk cmake files, but perhaps the detection isn't done properly before gtest is built?

richardbeare:
Good news - apart from the gtest problem it appears that variants of what we've discussed above allow a windows build given the appropriate cmake command line!

blowekamp:
This is great news! 

Want to do a pull request for adding the variable to the superbuild?

I'll do more research and experimentation on WIN32 vs MINGW

blowekamp:
I have made the pull request for `CMAKE_THREAD_LIBS_INIT` here #210.

CMake/sitkUseGTest.cmake
blowekamp: I am noting that CMake 3.1, does not define `MINGW` but [CMake 3.2 does](https://cmake.org/cmake/help/v3.2/variable/MINGW.html).

richardbeare: cmake on the windows machine I'm using is version 3.7.2. It is the version installed via msys2

blowekamp: I tested with mingw. The problem does not appear to be with the `if(MINGW)` but with not setting the variable as `CACHE`.
https://github.com/SimpleITK/SimpleITK/blob/08a81298823ecc50ab952a4c63a01a8d692d4643/CMake/sitkUseGTest.cmake#L40

This seems familiar, I may have had these changes local and forgot to commit them.

richardbeare:
I'll test these latest changes under windows.

richardbeare:
I spoke too soon - must have been accidentally installing an old version or not restarting correctly.
Neither cmake_thread_libs_init nor LINKER_FLAGS are being passed to the dll link phase.

Testing further now.

richardbeare:
It is looking to me as though cmake_thread_libs_init doesn't behave normally under windows. 

Is the simplest way forward to add a pair of R variables:

R_SHARED_LINK_FLAGS
R_SHARED_LINK_LIBS

so that they can be specified by hand? (other names as you see fit)

blowekamp:
This patch has been labeled with "backlog". It contains some good workaround to get R bindings working on windows. But it will need to be updated. My hope that the R tool chain gets fixes I don't think is realistic anything soon anymore.

richardbeare:
Note - my windows resources were limited. It is vaguely possible that a larger machine, e.g Azure cloud bases, may get the R tool chain to work. We should test this if possible on azure.

PR #198, opened 2017-06-01
Catch error of null ImageJ application

dave3d:
In this patch we are checking for a null ImageJ executable string.
We do this at the point where that string is replacing the "%a"
macro in the command string.  This allows the user to replace
the command string, possibly not using "%a".  This sitation
would occur if the user was using some other application.  In
that case, a null ImageJ app string would not cause an error.

Change-Id: I87a0c4b7072adc0e00744de28659b8a7366010a1

### DIFF
--- a/Code/IO/src/sitkShow.cxx
+++ b/Code/IO/src/sitkShow.cxx
@@ -111,6 +111,10 @@ namespace itk
               break;
             case 'a':
               // %a for application
+              if (!app.length())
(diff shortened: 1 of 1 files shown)


zivy:
LGTM. 
Tested using Python, with/without ImageJ, and with junk/valid setting for SITK_SHOW_COMMAND. Error messages were informative in all cases.

PR #200, opened 2017-06-01
Add meta data dictionaries to image series reader

blowekamp:


zivy:
From a dirty build I got an error:
```
sitkImageSeriesReader.cxx:235:35: error: 
      use of undeclared identifier 'SITK_NULL_PTR'
      this->m_pfGetMetaDataKeys = SITK_NULL_PTR;
```
I started a clean build and got the same error. Is there a missing include?

blowekamp:
The second underscore needs to be remove.

blowekamp:
I push the fix I had locally. CircleCI should come back green now.

zivy:
Compiles correctly, but the GetMetaDataKeys method throws an exception, no informative information: 
```
RuntimeError: Exception thrown in SimpleITK ImageSeriesReader_GetMetaDataKeys: std::exception
```

blowekamp:
This happens when the `MetaDataDictionaryArrayUpdate` parameter has not been enable. Please enable it to get the dictionaries to be updated.

The underlying cause is the  C++ `std::(tr1)::function` for the callback methods such as `GetMetaDataKeys` is NULL when the dictionaries are not updated, either before execution or when the array update is not enabled. This throws a `bad_function_call` exception who's `what` is just `std::exception`, and not something useful like "bad function call".

This same paradigm is used for all of the similar "active measurement", that the `function` pointer is set to NULL when it is not valid to make the call. What should be done to improve it?

zivy:
As far as I can see we have two options:
1. Leave the code as is and add documentation that warns the user (they won't read it so I guess this should also be added to the FAQ).
2. Change the SimpleITK code, adding a layer in between so that SimpleITK functions don't map directly to ITK ones (try-ITKFunction-catch).

Not sure if option 2 is viable?

blowekamp:
As this exception is the same as with other uses of ITK call backs, can you please create an issue to report the un-helpfull error message. And we'll commit this patch as is.

zivy:
Done

PR #211, opened 2017-06-09
More idiomatic setuptools import

ghisvail:


blowekamp:
See #175 for prior discussion. Clearly the current implementation is not the expected behavior.

ghisvail:
I am not quite sure I follow. The Python community has been standardizing on `setuptools` for a long time, so why the `sys.modules` hack instead of a true import? Anyway, feel free to close if the PR is not relevant.

blowekamp:
Here the PyPa currently recommends using setuptools:
 https://packaging.python.org/current/?highlight=distutils#installation-tool-recommendations

As I recall there are some systems ( Ubuntu 12 or 14? ) where the system setuptools was actually distribute, and directly importing setuptools was not reliable. As we only need setuptools to generate eggs and now wheels, we have a separate script setupegg.py which add the dependency of setuptools. It have been beneficial to not require a current version of setuptools, to enable compilation and testing of Python on older systems. Additionally, setuptools is outside the Python standard library.

I think a better solution may be to only try to import setuptools if it's required for the command. And then check that the version is new enough.

BTW, did you encounter a situation where not using setuptools caused a problem?

ghisvail:
> and directly importing setuptools was not reliable

Hence the `try catch` block?

> setuptools is outside the Python standard library

Not much of an argument considering how ubiquitous `setuptools` is.

> BTW, did you encounter a situation where not using setuptools caused a problem?

Probably not in your case, though the `sys.modules` query is very unpythonic.

blowekamp:
Thank you for this, I have rebased and updated your patch here #213.

PR #212, opened 2017-06-12
Fixing spelling mistakes in documentation and comments

blowekamp:
Debian's spellintian from lintian was used from a docker container to
check the code.

### DIFF
--- a/Code/Common/include/sitkPixelIDValues.h
+++ b/Code/Common/include/sitkPixelIDValues.h
@@ -45,7 +45,7 @@ struct ImageTypeToPixelIDValue
 
 /** \brief Enumerated values of pixelIDs
  *
- * Each PixelID's value correspondes to the index of the PixelID type,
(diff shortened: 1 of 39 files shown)


blowekamp:
@mrguilfoyle There were not as many spelling issues in the SimpleITK code as I feared.

zivy:
LGTM.

When looking at the mistakes, in some cases I barely noticed the difference between the correct spelling and the original, makes a difference for a spell checker, not so much for a human being.

Code/Common/include/sitkPixelIDValues.h
```diff
@@ -45,16 +45,16 @@ struct ImageTypeToPixelIDValue
 
 /** \brief Enumerated values of pixelIDs
  *
- * Each PixelID's value correspondes to the index of the PixelID type,
+ * Each PixelID's value corresponded to the index of the PixelID type,
```
dave3d: I think it should be **corresponds**.

PR #216, opened 2017-06-15
Create compare driver

blowekamp:


zivy:
LGTM.
I do prefer the naming convention in CMake  IMAGE_COMPARE which "hides" the "--compare" which is less meaningful (yes backwards compatibility and compatibility with the ITK driver).

PR #232, opened 2017-06-29
Hello world example

dave3d:


Examples/HelloWorld/CMakeLists.txt
```diff
@@ -0,0 +1,8 @@
+
```
zivy: Add all of the examples as tests, similar to the SimpleGaussian example [CMakeLists file](https://github.com/SimpleITK/SimpleITK/blob/next/Examples/SimpleGaussian/CMakeLists.txt).

Examples/CMakeLists.txt
```diff
@@ -15,6 +15,11 @@ if(NOT CMAKE_PROJECT_NAME STREQUAL "SimpleITK")
 
 endif()
 
+option(SimpleITK_SHOW_TESTING_ENABLED "Enable the Show function in SimpleITK testing." OFF)
+if (NOT SimpleITK_SHOW_TESTING_ENABLED)
+  set(ENV{SITK_NOSHOW} TRUE)
```
blowekamp: I am not sure what this code block does. The tests are not effected by changes to when the CMake process is running.

Examples/HelloWorld/CMakeLists.txt
```diff
@@ -6,3 +6,45 @@ target_link_libraries ( HelloWorld ${SimpleITK_LIBRARIES} )
...
+  return ()
+endif ()
+
+if ( SimpleITK_SHOW_TESTING_ENABLED )
+
+  sitk_add_test( NAME CXX.Example.HelloWorld
+    COMMAND $<TARGET_FILE:HelloWorld>
+    )
+
+  if ( WRAP_CSHARP )
```
blowekamp: This `WRAP_CSHARP` should not be needed the `sitk_add_csharp_test` macro should take care of it.

Examples/HelloWorld/CMakeLists.txt
```diff
@@ -6,3 +6,45 @@ target_link_libraries ( HelloWorld ${SimpleITK_LIBRARIES} )
 if ( WRAP_CSHARP )
   csharp_add_executable( HelloWorldCS SimpleITKCSharpManaged.dll ${CMAKE_CURRENT_SOURCE_DIR}/HelloWorld.cs )
```
blowekamp: Using the `sitk_add_csharp_test macro below should take care of building the csharp code.

dave3d: I'm not using the sitk_add_charp_test.  It's actually using sitk_add_test.  I wanted the Csharp to compile even without testing.

Examples/CMakeLists.txt
dave3d: I wanted to force the environment variable SITK_NOSHOW to be true by default when running the tests.  Not sure it actually accomplishes that though.

blowekamp: The `ENVIRONMENT` property needs to be modified for each test. Just about every test goes through the sitk_add_test macro. The code block from the python and lua add_tests can be consolidated in to sitk_add_test. The `[APPEND`](https://cmake.org/cmake/help/v3.0/command/set_property.html) option should be added to the `set_property` command as well.

Examples/HelloWorld/CMakeLists.txt
blowekamp: Using `sitk_add_csharp` would make it consistent with the other two csharp examples and java. the C++ does appears to always get compiled though.

I am not sure the behavior of `BUILD_TESTING` is consistent. The behavior could be make consistent inside the `sitk_add_java_test` and `sitk_add_csharp_test`. Please remove the separate sharp compilation, and use the macro. The logic for building the tests can be modified in another patch inside these  testing macros.

blowekamp:
Github seems to be a little confused on this patch now after the rebase. I am going to close it so you can open a new pull request, that we can review. 

Thanks Brad

PR #234, opened 2017-07-03
BUG: Use of floor division instead of regular division.

zivy:
The compose filter expects all inputs to have the same pixel
type. This was not the case when using regular division. Changed it to
floor division.

Change-Id: I9e7dacd9d04f5d4d680c97fa3fcdc14aabfc2da8

### DIFF
--- a/Examples/DemonsRegistration1/DemonsRegistration1.py
+++ b/Examples/DemonsRegistration1/DemonsRegistration1.py
@@ -72,5 +72,8 @@ if ( not "SITK_NOSHOW" in os.environ ):
     out = resampler.Execute(moving)
     simg1 = sitk.Cast(sitk.RescaleIntensity(fixed), sitk.sitkUInt8)
     simg2 = sitk.Cast(sitk.RescaleIntensity(out), sitk.sitkUInt8)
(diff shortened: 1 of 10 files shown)


blowekamp:
This usage occurs in a number of files:
```
$ git grep -l sitk.Compose
DemonsRegistration1/DemonsRegistration1.py
DemonsRegistration2/DemonsRegistration2.py
ImageRegistrationMethod1/ImageRegistrationMethod1.py
ImageRegistrationMethod2/ImageRegistrationMethod2.py
ImageRegistrationMethod3/ImageRegistrationMethod3.py
ImageRegistrationMethod4/ImageRegistrationMethod4.py
ImageRegistrationMethodBSpline1/ImageRegistrationMethodBSpline1.py
ImageRegistrationMethodBSpline2/ImageRegistrationMethodBSpline2.py
ImageRegistrationMethodDisplacement1/ImageRegistrationMethodDisplacement1.py
ImageRegistrationMethodExhaustive/ImageRegistrationMethodExhaustive.py
```

Do you think these others are effected but this Python 3 compatibility too?

zivy:
Good catch. Fixed this in all the files.

blowekamp:
Thanks for the additional changes! 👍

PR #235, opened 2017-07-03
Adding text to the Demons registration documentation files.

zivy:
Adding the overview for the algorithms. Explanations of what the code
is doing.

Change-Id: I822e8359f7179c5988399444e815fdeac3e858bb

### DIFF
--- a/Examples/DemonsRegistration1/Documentation.rst
+++ b/Examples/DemonsRegistration1/Documentation.rst
@@ -7,9 +7,20 @@ DemonsRegistration1
 Overview
 --------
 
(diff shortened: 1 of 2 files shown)


blowekamp:
The additional prove look pretty good. There are two points to discuss:
* Your paragraphs are a single line of text. I think it would be better to wrap the lines as is done in the C++ doxygen comments and [here](https://raw.githubusercontent.com/SimpleITK/SimpleITK/next/Examples/SimpleGaussian/Documentation.rst).
* The paragraph regarding using histogram matching is mostly repeated. As with code, repeating prose should be avoided. Perhaps a link from the second example to the prose in the first would suffice?

zivy:
I changed the paragraphs to use hard line breaks. 

With regard to the content, for now I would like each of the examples to be self contained so let's keep it as is. We should revisit this once we figure out the overall layout for SimpleITK documentation (user guide, examples, doxygen, notebooks) and what content goes where.

PR #256, opened 2017-08-02
Wrapping IterativeInverseDisplacementFieldImageFilter

blowekamp:


hjmjohnson:
@blowekamp This looks like a good addition with a test.  I'd recommend merging.

blowekamp:
@hjmjohnson This filter got stalled when I tried to look at the difference between the filters in ITK to invert displacement fields. It's not clear to me what filters in ITK are "good" and useful. Nor what the current recommended filter to invert displacement fields is. Perhaps you can provide some information, or update the ITK Doxygen to clarify these filters.

Are you familiar them: [InvertDisplacementFieldImageFilter](https://itk.org/Doxygen/html/classitk_1_1InvertDisplacementFieldImageFilter.html),  [InverseDisplacementFieldImageFilter](https://itk.org/Doxygen/html/classitk_1_1InverseDisplacementFieldImageFilter.html), [IterativeInverseDisplacementFieldImageFilter](https://itk.org/Doxygen/html/classitk_1_1IterativeInverseDisplacementFieldImageFilter.html), or the remote [FixedPointInverseDisplacementFieldImageFilter](https://itk.org/Doxygen/html/classitk_1_1FixedPointInverseDisplacementFieldImageFilter.html).

Currently we have the first two, but there have been reports that these are providing poor results and other algorithms are needed. I am not sure if it is worth the binary bulk providing the additional algorithms. In trying to understand the differences between the implementation the Doxygen is not of much help.

Issue #255, opened 2017-08-02
Python wrapping for FixedPointInverseDisplacementFieldImageFilter


rcarthigesan:

blowekamp:
The filter you are requesting is already wrapped.

Currently there are two filters to invert displacement fields in SimpleITK:
 * [InvertDisplacementFieldImageFilter](https://itk.org/SimpleITKDoxygen/html/classitk_1_1simple_1_1InvertDisplacementFieldImageFilter.html)
  * [InverseDisplacementFieldImageFilter](https://itk.org/SimpleITKDoxygen/html/classitk_1_1simple_1_1InverseDisplacementFieldImageFilter.html)

There are two additional filters in ITK which can invert displacement fields:
 * [IterativeInverseDisplacementFieldImageFilter](https://itk.org/Doxygen/html/classitk_1_1IterativeInverseDisplacementFieldImageFilter.html)
 * Remote module: [FixedPointInverseDisplacementFieldImageFilter](https://itk.org/Doxygen/html/classitk_1_1FixedPointInverseDisplacementFieldImageFilter.html)

ramanacarthigesan:
Sorry, I meant to write "FixedPointInverseDisplacementFieldImageFilter", which is not yet wrapped.

Apologies for the confusion. Will adjust the title accordingly.

blowekamp:
Please see pull request #257. To use the filter you will need to compile SimpleITK yourself from the Superbuild with the added CMake option `-DModule_FixedPointInverseDisplacementField=ON`.

rcarthigesan:
Hi,

I just tried doing this. After building was complete, the FixedPointInverseDisplacementFieldImageFilter module wasn't included in the Python wrapping.

However, it _was_ included in the ITK build - when I look at the SimpleITK-build/ITK-build/Modules/Remote/ directory, there is a folder named FixedPointInverseDisplacementFieldImageFilter.

Not sure why this is happening?

For reference, this is the code I used for building:

[...]

PR #257, opened 2017-08-02
Wrapping FixedPointInverseDisplacementFieldImageFilter

blowekamp:
addresses #255


blowekamp:
This filter is in a remote ITK module, for SimpleITK to use it ITK must be configured with `Module_FixedPointInverseDisplacementField` enabled. This can be done when building SimpleITK via the Superbuild by adding `-DModule_FixedPointInverseDisplacementField=ON` to the cmake command line when configuring from the SimpleITK Superbuild.

Currently, the tests segfault because the filter is not handling a zero sized regions properly.

blowekamp:
The segfault has been addressed in the remote module InsightSoftwareConsortium/ITKFixedPointInverseDisplacementField#2 and ITK has been updated
InsightSoftwareConsortium/ITK@a01a3ab50f51fdc8a5db4b7daff3bb7d91944b6b. This wrapping will require ITK newer that 4.12.1

Issue #260, opened 2017-08-08
Explicitly enable C++11 support


msmolens:
On Linux, I configure ITK 4.12 to build with C++11 by setting the following CMake options:
```
-DCMAKE_CXX_STANDARD:STRING=11
-DCMAKE_CXX_STANDARD_REQUIRED:BOOL=ON
-DCMAKE_CXX_EXTENSIONS:BOOL=OFF
```
I configure the SimpleITK SuperBuild with those same options and  set `ITK_DIR` to the ITK build. However, this doesn't enable C++11, and the SimpleITK build fails:

Configure step:
```
-- Performing Test CXX_HAS_stdcxx11
-- Performing Test CXX_HAS_stdcxx11 - Success
-- Checking if c++11 is required...
-- Checking if c++11 is required... NO
```
Build step error, for example:
```
.../SimpleITK/Code/Explicit/src/sitkExplicitITKDefaultVectorPixelAccessor.cxx:19:
.../ITK-build/Modules/Core/Common/itkConfigure.h:59:6: warning: #warning "WARNING:  The current project is configured to use a C++ standard version older than the C++ standard used for this build of ITK" [-Wcpp]
     #warning "WARNING:  The current project is configured to use a C++ standard version older than the C++ standard used for this build of ITK"
      ^
```

How can I explicitly enable C++11 for the SimpleITK SuperBuild?

blowekamp:
I have always just added the proper flag to CMAKE_CXX_FLAGS, i.e. `-DCMAKE_CXX_FLAGS:STRING=-std=c++11`. The CMake variable you set were not propagate from the super build to the sub-projects.

I have a proposed patch #261, that I have not tested but should work for this case. I need to verify some behavior and compatibilities with CMake versions and interactions with CXX_FLAGS.

jcfr:
@blowekamp  Thanks for fixing the issue and integrating into master. I am doing a test build with Slicer and will keep you posted.

blowekamp:
It is recommended to still use `-DCMAKE_CXX_FLAGS:STRING=-std=c++11` due to [CMP0067](CMP0067), unless CMake>3.8.2. 

Please take note of the following:
https://github.com/SimpleITK/SimpleITK/blob/next/CMake/sitkCheckCXX11.cmake#L21-L29

[...]

PR #261, opened 2017-08-08
Propagate CMake variables related to compiler standard  and extension

blowekamp:
If the SimpleITK Superbuild was configured with CMake variables such
as:

  -DCMAKE_CXX_STANDARD:STRING=11
  -DCMAKE_CXX_STANDARD_REQUIRED:BOOL=ON
  -DCMAKE_CXX_EXTENSIONS:BOOL=OFF

Then these options were not propagated to the external projects in the
superbuild.


blowekamp:
Addresses #260

hjmjohnson (approved):

blowekamp:
CMAKE_CXX_STANDARD was introduced with CMAKE 3.1. Currently, SimpleITK require CMake>=3.0. Passing this variable if it is defined is not a compatibility issue.

However, the file `sitkCheckRequiredFlags.cmake` may add `std=c++11` to the cxx_flags. Compatibility and interaction needs further investigation, as using two methods to set the CXX standard may pose problems.

blowekamp:
When `std=c++11` is automatically added to CXX_FLAGS or added by the user it will be added to the "compiler_options" for the exported SimpleITK targets. When CXX_STANDARD is requested or required  
, this is currently will not be propagated to the interface for use of the SimpleITK libraries.

https://cmake.org/cmake/help/v3.1/prop_tgt/CXX_STANDARD.html#prop_tgt:CXX_STANDARD

It is not clear how to specify this as part of the interface to a library...

blowekamp:
On linux with cmake 3.7.2, the `SITK_HAS_CXX11` are failing even thought  CMAKE_CXX_STANDARD is set to 11. Manually propagating the CMAKE options to the try_compile make let is success as expected.

PR #271, opened 2017-08-30
Work around CMP0054 warning

blowekamp:
Add 'x' prefix and suffix to ensure extra de-referencing of variable
name does not occur.


hjmjohnson:
@blowekamp Looks good to merge.

Issue #275, opened 2017-09-05
 Wrapping/Python/CMakeLists.txt: Fix CMake warning related to policy CMP0026


jcfr:
```
CMake Deprecation Warning at Wrapping/Python/CMakeLists.txt:12 (cmake_policy):
  The OLD behavior for policy CMP0026 will be removed from a future version
  of CMake.

  The cmake-policies(7) manual explains that the OLD behaviors of all
  policies are deprecated and that a policy should be set to OLD only under
  specific short-term circumstances.  Projects should be ported to the NEW
  behavior and not rely on setting a policy to OLD.
```

blowekamp:
close by #279

PR #279, opened 2017-09-06
Replace configuration_file_build_time

blowekamp:
The old build time configuration file was fundamentally flawed. The
configuration specific path to a target is can be only conveyed with
CMake via generator expressions. But this function did not use them,
but relied on setting and the LOCATION property of the target to the
same location for each target. So the build time configuration wasn't
even needed. Also the LOCATION manipulation produced a CMP0026
warning.

This new approach does an initial basic configuration, then does a
second pass which actually does a build time configuration with CMake
generator expressions manually passed.

The old CMake file to provide the function has been removed.


blowekamp:
This should address issue #275

Wrapping/Python/LegacyPackaging.cmake
```diff
@@ -2,35 +2,40 @@ if ( SimpleITK_DOC_FILES )
...
 endif()
 
-include( sitkConfigureFileBuildtime )
-configure_file_buildtime( "${CMAKE_CURRENT_SOURCE_DIR}/Packaging/setup.py.in"
-  "${CMAKE_CURRENT_BINARY_DIR}/Packaging/setup.py" )
+# do initial configuration of CMake variables
+set(SimpleITK_BINARY_MODULE "@SimpleITK_BINARY_MODULE@")
+configure_file(
+  "${CMAKE_CURRENT_SOURCE_DIR}/Packaging/setup.py.in"
+  "${CMAKE_CURRENT_BINARY_DIR}/Packaging/setup.py.in" )
```
zivy: should this be 'setup.py'  or are you just copying it over to the build directory?

PR #281, opened 2017-09-06
Initial checkin of HelloWorld example

dave3d:
A simple "Hello World" example for SimpleITK in all the supported languages.
This example creates an image and then calls the Show function, which
launches ImageJ with the image.  To run the corresponding tests,
set the CMake variable SimpleITK_SHOW_TESTING to ON.

Change-Id: I19a6de7f669eacd2517d7840ddcdfd445b00e55e

### DIFF
--- a/Examples/CMakeLists.txt
+++ b/Examples/CMakeLists.txt
@@ -15,6 +15,11 @@ if(NOT CMAKE_PROJECT_NAME STREQUAL "SimpleITK")
 
 endif()
 
+option(SimpleITK_SHOW_TESTING_ENABLED "Enable the Show function in SimpleITK testing." OFF)
+if (NOT SimpleITK_SHOW_TESTING_ENABLED)
+  set(ENV{SITK_NOSHOW} TRUE)
+endif ()
+
 # Add individual cxx executables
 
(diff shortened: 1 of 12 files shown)


Examples/CMakeLists.txt
```diff
@@ -15,6 +15,11 @@ if(NOT CMAKE_PROJECT_NAME STREQUAL "SimpleITK")
 
 endif()
 
+option(SimpleITK_SHOW_TESTING_ENABLED "Enable the Show function in SimpleITK testing." OFF)
```
blowekamp: please add  `mark_as_advanced( SimpleITK_SHOW_TESTING_ENABLED)`

Examples/CMakeLists.txt
```diff
@@ -15,6 +15,11 @@ if(NOT CMAKE_PROJECT_NAME STREQUAL "SimpleITK")
...
+if (NOT SimpleITK_SHOW_TESTING_ENABLED)
+  set(ENV{SITK_NOSHOW} TRUE)
```
blowekamp: This `if` block has no effect and should be removed.

blowekamp:
This looks good. Merging into next.

PR #300, opened 2017-09-22
Do not create /.git folder when the project is configured

lassoan:
A .git folder kept popping up on in the root folder of my and many of my coworkers' computers.
It was really annoying, because it confused git clients to think that all folder is under revision control
and we had to delete the folder manually. After that the .git folder appeared again, and again...
It remained a mystery for several years ... until now.

We've finally found the culprit: it was a copy-paste error in sitkCheckSourceTree.cmake.
This commit fixes the issue.


blowekamp:
Nice find! Glad you were able to track this down.

Would you be able create a branch off origin/release, and cherry pick this commit? 

Thanks for contributing!

lassoan:
I guess this is what you meant: https://github.com/SimpleITK/SimpleITK/pull/301

blowekamp:
Merging from #301

PR #301, opened 2017-09-22
Do not create /.git folder when the project is configured

lassoan:
A .git folder kept popping up on in the root folder of my and many of my coworkers' computers.
It was really annoying, because it confused git clients to think that all folder is under revision control
and we had to delete the folder manually. After that the .git folder appeared again, and again...
It remained a mystery for several years ... until now.

We've finally found the culprit: it was a copy-paste error in sitkCheckSourceTree.cmake.
This commit fixes the issue.


blowekamp:
Great! Thank you

blowekamp:
Merged topic into master branch:
92db92245ae762cdb6e9bfb44a41dedc254af9cb
and release branch:
cabfb21dd189aea822c688e863a38fdc40a9f1fe

PR #306, opened 2017-10-02
Adding instructions on error reporting.

zivy:
Added detailed instructions on how to report an issue.

Change-Id: I152bbfffd31e1e10b6830b5662dce730a2c190c6


blowekamp:
Looking over that top level readme, it is in desperate need or re-organization, and an update.

This looks pretty good, thanks for tackling this.

But for this particular section you are missing specification for the binary. Did the user compile it themselves? Or download and install it from some place? Also, we could include a little encouragement for the community to participate in the bug reporting process.

zivy:
Addressed all of the comments in the updated pull request.

PR #313, opened 2017-10-14
Remove setup scripts for Gerrit/Git itk.org server

addisonElliott:
From my experience, there is usually no setup scripts required to
configure your Git environment when using Github to host your project.
However, when Gerrit and an external git repository is used not hosted
on Github's servers, then there are necessary steps to get up and
running.

This is usually automated in the form of scripts. I removed these
scripts since they were preventing me from submitting new changes due to
the git hooks.

When I go to run:
`git commit`

I get the following:
Your work tree has not been configured for SimpleITK development.
```
Paste the following commands into a shell:

 ./Utilities/SetupForDevelopment.sh

See https://www.itk.org/Wiki/ITK/Git/Develop for more details.
```

When running the SetupForDevelopment.sh script, I got the following message:
`We advise setting git://itk.org/SimpleITK.git as your origin.`


blowekamp:
These scripts should stick around the setup some useful local hooks. The itk.org is the official upstream, Github is just a mirror of it. Additionally there are some hooks to help manage data too. 

These script could use updating, removing gerrit, providing a standard remote name for the user’s github repo and this one. 

Must of this we had followed ITK on. But now we are ahead of the with github usage, so we are waiting for them to establish conventions for the ITK community for us to follow.

addisonElliott:
Most of the projects I've worked on that have been hosted on Github, do not have any Git hooks. As an example, here are the guidelines for contributing via OpenCV: [here](https://github.com/opencv/opencv/wiki/How_to_contribute)

I did a basic look through the scripts and I just didn't see any relevance for them:
* **GitTips.sh** - Doesn't **do** anything but it provides some tips to the user such as how to enable color coding when running git commands. (I think this is on by default because I have it). I think this is fairly obvious to most people and if not, the information can be found all over the web.
* **SetupGerrit.sh** - This can be removed since SimpleITK can use Github instead of Gerrit now, right?
* **SetupGitAliases.sh** - Contains aliases that I personally don't think I would use. And if I wanted too, couldn't I add this via commandline at a different time?
* **SetupTopicStage.sh** - Creates public/private key for communicating via itk.org from what I understand. In Github, I had to setup an SSH key once when I first installed it but not afterwards. So, if this isn't your first time contributing via Github, it won't be necessary.
* **SetupUser.sh** - Configures user name and email. This only needs to be done once and then it should be saved globally. I think when you install Git it has you set it up upon installation if I remember correctly.
* **SetupHooks.sh** - I think the Python JSON validation and KWstyle hooks may be necessary. I overlooked this at first... not sure it's necessary to pull/fetch these from itk.org if they are just stored locally.
* **SetupForDevelopment.sh** - Runs all of the above files. Didn't see anything else necessary in it.
* **git-gerrit-push** - Looking through this, nothing seems to be essential. 
* **pre-commit** - Code for checking *.txx file extension being deprecated = useful, refuse addition of submodules, style hooks, validate json = useful
* **prepare-commit** - Looks useful
* **.hooks-config.bash** - Necessary if we want to keep some of the hooks.

Let me try revising this and keeping some of the stuff I think is useful.

addisonElliott:
@blowekamp Can you take a look at this?

I got it down to what I think are necessary or helpful to the user. Below is basically a list of all of the fundamental things that the hooks do. I'm going to **bold** the ones that I **don't** think are necessary. Let me know what you think.

* **I removed the hook chain script because there is no chain anymore. All of the hook script is contained in the Utilities/Hooks folder until ITK gets a repository updated for Github and removes their Gerrit code. In the future, do you think it will be beneficial to keep synced with ITK's hooks? The hooks are not complicated enough to matter imo**
* **Check for \*.txx extension and say it's deprecated**
* **Restrict addition of submodules**
* **Check valid username and email (must have full name with space and email must have @)**
* Disallow non-ASCII filenames
* **Require newline at end of file**
* Check filemodes to make sure that executable files have +x
* **Check max size (2MB), throw error if exceeded**
* **Add BUG,COMP,DOC,ENH prefix to commit message. I don't think this is necessary since you can submit a pull request to Github when you are ready to have code reviewed and doesn't github have labels?**
* In first line of commit message, must be > 8 chars and < 78
* Second line of commit message must be blank

blowekamp:
A lot of these hooks a processes come from the work that Kitware has done, and the processes established with ITK or CMake. Much of the prior documentation from the software processes was just we are following projects like ITK or CMake for the software processes. For you to get up to speed on all the processes and history of them, there would be a bit of digging through the wiki documentation required. Also some that was reference from SimpleITK appears to have moved/changed.

Here is one page that about how the "topic stage" is used to manage topic in the next branch:
https://public.kitware.com/Wiki/Git/Workflow/Stage

The branches managed in the topic stage, and merging them into the next and then the master branches is managed by the core developers, and not integrated into the Github workflow well.

blowekamp:
To move this forward a good place to start is to just remove the Gerrit stuff.

The other issue to address is Github vs itk.org for the upstream. I have created an issue #316, to describe the problem and discuss the best solution to document and setup a contributors repos.

PR #314, opened 2017-10-14
Include make.bat for building Sphinx docs on Windows

addisonElliott:
Include make.bat file to allow for building documentation via Sphinx on
Windows. 

This file is directly modelled from the make.bat generated by
`sphinx-quickstart`.


blowekamp:
This looks like a good idea since we have a ‘Makefile’ too. Ideally I’d like to move to a CMake build tool, but haven’t found the right CMske module yet.

addisonElliott:
Yeah, when you type `sphinx-quickstart` it generates the Makefile and make.bat.

Using a CMake module is interesting and would make things much easier. I don't have a **lot** of experience with writing CMakeLists, but is a module even necessary?

It sounds like it'd be a pretty simple script:

1. Add a BUILD_DOCS option. If BUILD_DOCS is set, then do this stuff.
2. Make sure we have Python. Throw an error if not found.
3. Install sphinx? Just do `pip install sphinx`
4. Build the docs by running the makefile.

I may be oversimplifying things...

blowekamp:
Your topic is based off of the next branch it must be based off of the master branch. Aside from that is it looks OK to merge.

PR #321, opened 2017-10-16
Include make.bat for building Sphinx docs on Windows

addisonElliott:
Include make.bat file to allow for building documentation via Sphinx on
Windows. This file is directly modelled from the make.bat generated by
`sphinx-quickstart`.

### DIFF
--- a/Documentation/docs/make.bat
+++ b/Documentation/docs/make.bat
@@ -1,0 +1,36 @@
+@ECHO OFF
+
+pushd %~dp0
+
+REM Command file for Sphinx documentation
+
+if "%SPHINXBUILD%" == "" (
+	set SPHINXBUILD=sphinx-build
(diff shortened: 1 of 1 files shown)


addisonElliott:
Okay, sorry if you got a bunch of annoying notifications on this. I was experimenting on the best way to switch the branch to derive from master instead of next.

A rebase wasn't so simple because there was conflicts and I would have to manually merge them. Instead, I just created a patch, created a new branch and applied the patch to the branch.

blowekamp:
The pull requests are merged with the stage repo, which does some additional validation, and management of a topic so that is can be merged into next, then master as needed. So it does not matter too much where the pull request is targeted to.

In this case I am merging the topic into next now, and Github will close the issue when its merged into master.

PR #322, opened 2017-10-16
Remove gerrit setup scripts2

addisonElliott:
From my experience, there is usually no setup scripts required to
configure your Git environment when using Github to host your project.
However, when Gerrit and an external git repository is used not hosted
on Github's servers, then there are necessary steps to get up and
running.

This is usually automated in the form of scripts. I removed these
scripts since they were preventing me from submitting new changes due to
the git hooks.

When I go to run:
`git commit`

I get the following:
Your work tree has not been configured for SimpleITK development.
```
Paste the following commands into a shell:

 ./Utilities/SetupForDevelopment.sh

See https://www.itk.org/Wiki/ITK/Git/Develop for more details.
```

When running the SetupForDevelopment.sh script, I got the following message:
`We advise setting git://itk.org/SimpleITK.git as your origin.`

Recreated pull request so that topic branch comes from master branch instead of next
Reference this pull request: https://github.com/SimpleITK/SimpleITK/pull/313


addisonElliott:
I did not make any changes to this because I basically removed the Gerrit stuff. This should be a good first step for the scripts that can be updated afterwards.

Here is a summary of what it does:
* Only the hooks are setup now. The other scripts were specific to Gerrit I believe.
* Instead of pulling the hooks from the ITK repository, they are now sourced locally. This is because the repository scripts are specific to Gerrit. 
* Local ones are contained in Utilities/Hooks/*. The KWstyle check and Python JSON validation are still contained. The pre-commit, commit and commit-msg **functionality** is still the same but without any Gerrit specific information.

I have tested this on a fresh clone and it worked for me.

blowekamp:
Please see #323 for a more focused patch.

PR #323, opened 2017-10-17
Remove legacy Gerrit setup, hooks and configured aliases

blowekamp:


addisonElliott:
This looks good. My question is what about SetupHooks.sh where it pulls from the ITK repository here: https://public.kitware.com/gitweb?p=ITK.git;a=tree;h=refs/heads/hooks;hb=refs/heads/hooks

There are particular checks such as requiring the ChangeId line that could be removed since that is a Gerrit specific thing, right?

blowekamp:
The upstream hooks branch is easily configurable with git config variables. The GerritId hook can be controlled with the `hooks.GerritId` being false or unset.

addisonElliott:
I did not read the code closely enough, I didn't notice the `hooks.GerritId` variable. I definitely agree that your pull request is much cleaner than mine. I will go ahead and close mine.

blowekamp:
Would you be able to test the updated setup scripts on your current repository and a new repository to determine if it behaves correctly?

addisonElliott:
I'm happy to be the guinea pig.

When running the SetupforDevelopment.sh script, I receive the following:
```
$ ./Utilities/SetupForDevelopment.sh
Checking Git version...
Git version 2.11.0.windows.1 is OK.

We advise setting git://itk.org/SimpleITK.git as your origin.

If you choose not to do that, then other instructions will not work as expected.
Do you wish to continue with this non-standard layout? [y/N]:
Please fix your origin remote, and re-run this script.

Please run the following to correct the origin url:

git remote set-url origin git://itk.org/SimpleITK.git
``` 

I think this is discussed in the other pull request that Gerrit used origin as the actual repository while in Github it is typically the user's fork.

blowekamp:
This issue was addressed in the following patch:
https://github.com/SimpleITK/SimpleITK/pull/324

PR #324, opened 2017-10-18
Update setup for development with Kitware git setup

blowekamp:


blowekamp:
@bradking @thewtex 

I am starting to update the SetupForDevelopment script to configure for GitHub usage. I have started by updating to what appears to be the latest set of development scripts. This seems fairly straight forward.

I see a couple things missing:
 - Where should git alias be set up? Use to be DevelopmentSetupScripts/SetupGitAliases.sh, these have some useful ones for core developers using the stage repo.
 - Where should custom configuration for hooks go? [this](https://github.com/SimpleITK/SimpleITK/blob/0a162003a73e8252eec1208ff89b1df632771b65/Utilities/DevelopmentSetupScripts/SetupHooks.sh#L59-L72)

The next step will be to create a setup-github script to configure the GitHub upstream, the users repo, and add an additional fetch for pull request.

thewtex:
> Where should git alias be set up? 

I think we should consider `git review-push`.  `git github-push` would follow the current convention, but `review` describes better what the command is intended to do; queue the branch for a code review. And, we will not have change the alias again if we ever had two migrate (knock on wood :deciduous_tree:).

bradking:
Aliases are per-project choices so the script can stay outside of the `GitSetup` directory.  The current `SetupGitAliases.sh` would be fine.

blowekamp:
> I think we should consider git review-push. git github-push would follow the current convention, but review describes better what the command is intended to do; queue the branch for a code review. And, we will not have change the alias again if we ever had two migrate (knock on wood 🌳).

I am not sure that a replacement is needed for `git gerrit-push`. With Github you need to:

1. Commit locally to a branch
1. Push to your GitHub fork 
1. Make a pull request to the `upstream` GitHub repo

As long as we can establish a convention for the different repo name, the workflow should be straight forward for contributors to use git directly, and it should be a very familiar workflow to them ( we just may set up different repo names than the standard [Github workflow ](https://gist.github.com/Chaser324/ce0505fbed06b947d962)).

I am thinking of the following repo names:
 - `origin` - still https://itk.org/SimpleITK.git
 - `upstream` - https://github.com/SimpleITK/SimpeITK.git
 - `github` - https://github.com/yourname/SimpleITK.git

Optional
- `stage` - still https://itk.org/Stage/SimpleITK.git

thewtex:
I think just `upstream` and `origin` remotes will keep us closer to the standard GitHub workflow remotes and keep it simpler.

`git review-push` or `git github-push` would push the branch to the fork, open or update the PR, and spit out the PR URL in the console.

bradking:
If we are fully on github then the `itk.org` repos should not be needed at all.  They will remain only as mirrors.  We shouldn't give them names or encourage their use during development.

Lots of GitHub tutorials suggest forking and *cloning one's own fork*.  That's why the fork gets the name `origin` and a separate `upstream` is needed for the original repo.  I've never liked that approach when the goal is to develop in the upstream repository with the personal forks used only for submitting PRs.  Imagine people start using SimpleITK by cloning the upstream repository.  Then when a bug is found we expect them to fork and clone again to do development?  Instead they should be able to just run `SetupForDevelopment.sh` and get to a state ready for development.

In VTK we have everyone clone from the official repository so `origin` means the same thing for all users and developers.  Then the personal fork gets some other name (`gitlab`) set up by the `SetupForDevelopment.sh` script.

In CMake we don't try to enforce any conventions and just ask people to figure out fork+MR themselves.  There are lots of tutorials out there for people to learn and choose their own preferred combination.

blowekamp:
SimpleITK is still planning on using the `next` branch to fully test the staged topics on the nightly dashboard because comprehensive testing across platforms and languages is too cumbersome for
CI. The additional stability of the master branch if very nice with the system. This necessitates the continues use of the `stage` repository for at least the core developers who merge and integrate topic branches.

> I think just upstream and origin remotes will keep us closer to the standard GitHub workflow remotes and keep it simpler.

This is the Github workflow where:
 - `origin` - is your forked repo i.e. https://github.com/yourname/SimpleITK.git
 - `upstream` - https://github.com/SimpleITK/SimpleITK.git

This is different than what has been traditionally done with ITK development and what I have become a custom to using.... I have always thought of `origin` as the **authoritative** repository. And as bk points out, I usually end up with the `origin` being the authoritative repository I first checkout before I fork it for a contribution.

Many people have figured out how to do pull request with SimpleITK despite not updating the documentation, and just letting contributors figure out how to manage the remote, forks, and pull requests them selves.

Perhaps adding a `review-push` script may hinder the ability for some people to "figure it out". And the lest scripts we have that rely on the "expected" repo layout the better for people to figure it out.

Clearly there are many possibility and more thought and discussion would be useful.

bradking:
The only technical reason to ask developers to use a script to push is for ExternalData support.  That's one reason VTK uses it and CMake doesn't.  ITK uses ExternalData.  Does SimpleITK?

blowekamp:
We do use ExternalData, but SimpleITK never had it integrated with the gerri-push script. A separate script had to be manually run to upload the data. It's has gotten a bit clunky at the data repo have changed, and is in need of an update too.

thewtex:
Yes, `origin` + `github` or `yourname` may be easier to keep consistent with initial clones and check this in `SetupForDevelopment.sh`.

addisonElliott:
Being a newcomer to SimpleITK but accustomed to GitHub rules, my take is that the `upstream` pointing to the SimpleITK repo and `origin` point to the forked repo.

The origin remote is automatically setup by GitHub when you clone your fork and then a script to setup the `upstream` is all that is needed.

I like the idea of having a script to automatically fetch and update your repository with the upstream repo because you have to keep both the `next` and `master` branches in sync preferably. Something like `git upstream-sync` or something similar.

I don't think a command to automatically submit a pull request is necessary. Once you understand the workflow on GitHub, then it's all pretty self explanatory and creating a pull request from the website is easier in my opinion.

Just my two cents as a newcomer

thewtex:
@addisonElliott Thanks for the thoughts. Yes, `origin` and `upstream` are inline with the conventional GitHub workflow.

There are many scientific research software developers that are not familiar with GitHub, so we should make the barrier to entry as low as possible for them and the process well documented.

I think a `review-push` alias that opens a PR is also nice for experienced because it cuts the steps in half (you don't have to manually open a PR).

Thinking through the steps required for a new contributor to submit a patch, they would:

1. Sign up on GitHub
2. Fork the repository
3. Clone the fork
4. `cd SimpleITK`
5. `./Utilities/SetupForDevelopment.sh`
6. `git checkout -b my-topic`
7. `git review-push`

`SetupForDevelopment.sh` could do a sanity check that the `origin` remote is not *github.com/SimpleITK/SimpleITK*.

addisonElliott:
@thewtex I see your point. The thing about git aliases is you can add them and choose not to use them. So, if the person wants to use the alias, then it can be available but if I prefer to submit a PR via Github website, then I can do that as well.

In addition, I was testing a pull request and noticed that it can be a little difficult to checkout the pull request. This is how I did it according to this [article](https://github.com/TeamPorcupine/ProjectPorcupine/wiki/How-to-Test-a-Pull-Request)
```
git fetch upstream pull/<id>/head:<branch>
git checkout <branch>
```

To make things easier, this could be done with a Git alias such that you can say `git pr <id>` and it will get and check it out. The article I posted provides code for making that alias as well. Thoughts?

thewtex:
@addisonElliott Yes, the `git pr` alias looks very convenient, too!

PR #326, opened 2017-10-23
Adding Oriented Bounding Box to the LabelIntensity statistics filter

blowekamp:


blowekamp:
I am not certain the `LabelIntensityImageFilter` needs or should have all the `ShapeLabelStatistics` options and values. Currently the underlying ITK class always computes them, so we are exposing them. The OrientedBounding box attributes are optionally computed and are exposed in the `ShapeLabelStatistics`.

zivy:
From a design standpoint I think exposing all of the shape statistics doesn't make sense for this class, as its purpose is to provide statistics on the intensities of specific labels.  
As ITK already exposes this information through the class interface we should probably do the same, for consistencies sake.

PR #332, opened 2017-10-30
Remove jss.cls file from R vignette documentation

blowekamp:
This file is marked with a GPL license which is not compatible with
the Apache license. As this file is used to format and generate a
document, it is not clear what the implication for derived document
would be. To be conservative it is being removed.


zivy:
👍 
This may be a bit overboard, but as none of us are lawyers it ensures that there are no licensing issues. 

@richardbeare  
I understand you want to make the JSS paper the vignette for SimpleITK, following the journal's prompt to do so. Obviously we can do this using the Insight Journal style file, but it would be nice to keep the JSS style. 

Can you inquire with the journal with respect to the GPL license and whether including the style file in the SimpleITK code base for use in the vignette is considered derived work or if it has no effect on our license. 

@blowekamp 
If the journal says it has no effect on our license, possibly reconsider including it?

PR #334, opened 2017-10-31
Change to vector argument for sigma and variance of Gaussian

### LINKED ISSUE #163: Let sitk.DiscreteGaussianImageFilter support difference variances per direction
Referring to the discussion at http://public.kitware.com/pipermail/insight-users/2017-May/054511.html

The DiscreteGaussianImageFilter should allow setting variances for each direction. This can be useful when filter in-plane but not in-depth before resampling.

blowekamp:
Enable setting of the sigmas/variance on a per axis basis for the
SmoothingRecursiveGaussianImageFilter and the
DiscreteGaussianImageFilter. This is done may changing the parameter
to be a "dim_vec". The scalar method for the object oriented interface
remains.

Additionally, the signature of scalar procedural remains as part of
the Additional procedural interface.

closes #163

### DIFF
--- a/Code/BasicFilters/include/sitkAdditionalProcedures.h
+++ b/Code/BasicFilters/include/sitkAdditionalProcedures.h
@@ -100,6 +100,29 @@ SITKBasicFilters_EXPORT Image PatchBasedDenoising (const Image& image1,
 #endif
 
 
+
+
+    /**
+     * \brief Blurs an image by separable convolution with discrete gaussian kernels. This filter performs Gaussian blurring by separable convolution of an image and a discrete Gaussian operator (kernel).
+     *
+     * This function directly calls the execute method of DiscreteGaussianImageFilter
+     * in order to support a procedural API
+     *
+     * \sa itk::simple::DiscreteGaussianImageFilter for the object oriented interface
+     */
+     SITKBasicFilters_EXPORT Image DiscreteGaussian ( const Image& image1, double variance, unsigned int maximumKernelWidth = 32u, double maximumError =  0.01, bool useImageSpacing = true );
+
+
+    /**
+     * \brief Computes the smoothing of an image by convolution with the Gaussian kernels implemented as IIR filters.
+     *
+     * This function directly calls the execute method of SmoothingRecursiveGaussianImageFilter
+     * in order to support a procedural API
+     *
+     * \sa itk::simple::SmoothingRecursiveGaussianImageFilter for the object oriented interface
+     */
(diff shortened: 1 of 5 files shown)


blowekamp:
closes #163

blowekamp:
The object oriented interface remains compatible.

There is some incompatibility with the procedural interface in Python and keyword arguments. SWIG does not support the keyword arguments when the method is overloaded. With this patch we have overloaded the DiscreteGaussian, and the SmoothingRecursiveGaussian methods. If users used keyword arguments before the call will not longer work.

Code/BasicFilters/include/sitkAdditionalProcedures.h
```diff
@@ -100,6 +100,29 @@ SITKBasicFilters_EXPORT Image PatchBasedDenoising (const Image& image1,
...
+    /**
+     * \brief Blurs an image by separable convolution with discrete gaussian kernels. This filter performs Gaussian blurring by separable convolution of an image and a discrete Gaussian operator (kernel).
+     *
+     * This function directly calls the execute method of DiscreteGaussianImageFilter
+     * in order to support a procedural API
+     *
+     * \sa itk::simple::DiscreteGaussianImageFilter for the object oriented interface
+     */
+     SITKBasicFilters_EXPORT Image DiscreteGaussian ( const Image& image1, double variance, unsigned int maximumKernelWidth = 32u, double maximumError =  0.01, bool useImageSpacing = true );
+
```
zivy: I just noticed that for DiscreteGaussian we are using variance vs. sigma in the SmoothingRecursiveGaussian. I am sure this will cause confusion for people who move between the discrete and recursive implementations. Should we add a \sa to each pointing to the other and saying something about sigma vs. variance? Not sure this will help much as people will likely not read it. What do you think?

richardbeare:
There are warnings about this in the ITK docs:

[" Also, note that we have itk::RecursiveGaussianImageFilter::SetSigma(), but itk::DiscreteGaussianImageFilter::SetVariance()."](https://itk.org/Doxygen412/html/classitk_1_1RecursiveGaussianImageFilter.html)

zivy:
@richardbeare 

You have a point. Unfortunately, the documentation is in ITK recursive Gaussian, but not in the discrete Gaussian. For SimpleITK we have the challenge of documentation of the procedural interface (we have none as all documentation is in the OO interface), this is where this can be an issue.

zivy:
Please merge these changes so that they make it into the next release. I would like to use the vector specification of sigma/variance in the notebooks on data augmentation (upcoming SPIE tutorial).

PR #336, opened 2017-11-06
DOC: Improving documentation based on user input.

zivy:
Improving the build documentation based on user input from this
discourse discussion:
https://discourse.itk.org/t/using-cmake-to-build-simpleitk/373

Change-Id: Ifbcda487d58128e88c37ba21606448b26a6aa6c9

### DIFF
--- a/Documentation/docs/source/building.rst
+++ b/Documentation/docs/source/building.rst
@@ -19,7 +19,6 @@ might prefer to **build SimpleITK from source**:
     :local:
     :backlinks: none
 
-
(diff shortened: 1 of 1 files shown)


blowekamp:
This is a good start but I think we need command line instruction that will work in a copy and paste way for MS Windows. Specifically, the build command have a path that is too long, and we should instruct using `cmake --build <dir>` instead of the general comment "use your build tool". The trick with that is ensuring the build tool compiles with multiple cores.

PR #348, opened 2017-12-04
Use itk factory component

blowekamp:


blowekamp:
This can be compiled by using the latest ITK `cmake -DITK_GIT_TAG=master SimpleITK/Superbuild`

I am getting the following configuration error:
```
CMake Error at /scratch/blowekamp/build/SimpleITK/ITK-prefix/lib/cmake/ITK/ITKModuleAPI.cmake:66 (message):
  No such module: "ITKImageIO"
  /scratch/blowekamp/build/SimpleITK/ITK-prefix/lib/cmake/ITK/ITKModuleAPI.cmake:26 (itk_module_load)
  /scratch/blowekamp/build/SimpleITK/ITK-prefix/lib/cmake/ITK/ITKModuleAPI.cmake:101 (_itk_module_config_recurse)
  CMake/sitkTargetUseITK.cmake:17 (itk_module_config)
  Code/IO/src/CMakeLists.txt:19 (sitk_target_use_itk)


-- Configuring incomplete, errors occurred!
```

The error is occurring in this file:
https://github.com/SimpleITK/SimpleITK/blob/6affbbb0ca9c96947581cf8b1b8808a3c681835f/CMake/sitkTargetUseITK.cmake#L17

The cmake function `itk_module_config` does not know anything about the ITKImageIO module. I expect to get the libraries, includes, etc for this meta-module.

@fbudin69500
What advice do you have with using this new meta-factory module in this situation?

Thanks!

fbudin69500:
I hadn't thought about this usage. ITKImageIO is only known in [ITKConfig.cmake](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CMake/ITKConfig.cmake.in#L81-L92), where it is replaced by all the modules that refer to that factory.
This piece of code requires `itk_module_config(ITK ${ITK_MODULES_ENABLED})` to be run before that to be aware of all the factories.
Your usage of this macro makes me think of two things:
1) Maybe you can call `find_package(ITK COMPONENTS ....)`, and after I merge [this patch](http://review.source.kitware.com/#/c/22853/) submitted by @jcfr , that should work.
2) The next step in the redesign of the IO registration process was to add a CMake function that would allow to set your components on each target individually. This function has not been implemented yet, maybe you need to wait for this new function before SimpleITK is update to support the new IO registration.

blowekamp:
> The next step in the redesign of the IO registration process was to add a CMake function that would allow to set your components on each target individually. This function has not been implemented yet, maybe you need to wait for this new function before SimpleITK is update to support the new IO registration.

I am not sure what you mean by "allow to set your components on each target individually".  Could you please explain?

fbudin69500:
Looking back at my notes on the matter, here is an example of the design that was anticipated when using ITK:

```
cmake_minimum_required(VERSION 3.0)
project(MyApp)
find_package(ITK REQUIRED)

add_executable(MyApp main.cpp)
target_link_libraries(MyApp ${ITK_LIBRARIES})
target_include_directories(MyApp PUBLIC ${ITK_INCLUDE_DIRECTORIES})
itk_configure_factory(MyApp)
```
This would be what someone used to CMake 3.0 would expect with a target based design.

**So contrary to what I said above**, the components are not set (in this design), by the function that configures the factory, but they would still be set by the `find_package()` call.

So, forgetting the `2.` I wrote previously, could you use `find_package()` in your case as I suggested in `1.`?

blowekamp:
The [sitkTargetUseITK.cmake](https://github.com/SimpleITK/SimpleITK/blob/6affbbb0ca9c96947581cf8b1b8808a3c681835f/CMake/sitkTargetUseITK.cmake) is implemented following the CMake 3.0 style where the `find_package` exports a set of targets.  Then the user has the imported target with intrinsic properties for the include paths, library paths, and dependencies specified. So a call to `target_link_libraries` with a link interface specification can be used with the _enumerated_ targets the user needs.The function provided by sitkTargetUseITK is suppose to behave like `target_link_libraries` if all of ITK targets had the full CMake properties with their dependencies and usage paths.

For SimpleITK, I do a top level [`find_package(ITK)`](https://github.com/SimpleITK/SimpleITK/blob/6affbbb0ca9c96947581cf8b1b8808a3c681835f/CMakeLists.txt#L64), then for each SimpleITK library I specify the [ITK targets](https://github.com/SimpleITK/SimpleITK/blob/6affbbb0ca9c96947581cf8b1b8808a3c681835f/Code/Common/src/CMakeLists.txt#L31-L36) it needs to include. This reduces the number of include directories and libraries used when building each SimpleITK library.

Following you example:

```
cmake_minimum_required(VERSION 3.0)
project(MyApp)
find_package(ITK REQUIRED)

add_executable(MyApp1 main1.cpp)
target_link_libraries(MyApp ITKCommon)

add_executable(MyApp2 main2.cpp)
target_link_libraries(MyApp ITKCommon ITKImageIO)```

In this case you don't what both executable to know about all of ITK. Only during the compilation of  the second executable will the include path include the ImageIO.

fbudin69500:
This is a good point and for now this is not addressed. I'll look into it.

fbudin69500:
Hello @blowekamp : I don't have a specific idea. The general idea is something like what you have in SimpleITK, which also adds the include directory where the factory header files are generated, and that implements basically the registration code that is currently in USE_ITK.
One more requirement would be to be able to set the list of IOs that should be supported, maybe as COMPONENTS in the function call.

blowekamp:
The problem is that `itk_module_config` does not support the "meta" modules. Is there a way that it could? I need to find out what module it depends on, or what libraries and include paths are needed to use the meta-module.

fbudin69500:
I have an idea I will try to implement. I'll share that with you.

fbudin69500:
@blowekamp You can try this patch: http://review.source.kitware.com/#/c/22909/

PR #351, opened 2017-12-12
Use itk output window

blowekamp:


blowekamp:
pending this ITK patch:
http://review.source.kitware.com/#/c/22902/1

dave3d (approved):
Looks good to me.

PR #354, opened 2017-12-14
Prepare for it kv5

hjmjohnson:
Prepare for ITK v5.  Some deprecated interfaces are being used for SImpleITK.


hjmjohnson:
@blowekamp This is just the start of some fixes for SImpleITK to conform to future ITKv5.  There are deprecated interfaces being used.

blowekamp:
This is very much appreciated.  Can you please base the fixes on master?

On Dec 14, 2017, at 6:25 PM, Hans Johnson <notifications@github.com<mailto:notifications@github.com>> wrote:


@blowekamp<https://github.com/blowekamp> This is just the start of some fixes for SImpleITK to conform to future ITKv5. There are deprecated interfaces being used.

hjmjohnson:
Yes.  After master has a version of ITK that support the recent ITK_FUTURE_LEGACY_REMOVE bug.

 

Hans

blowekamp:
It is my understanding that these changes should still be compatible with the prior version. If you just need to test the against the current ITK, when you do you SimpleITK Superbuild configuration you could just add "-DITK_GIT_TAG=v4.13rc02 -DITK_FUTURE_LEGACY_REMOVE=ON" and the version of ITK build should be as you desire. There are a lot of options to configure ITK with and the should just get passed OK with the SimpleITK super build, just FYI about these.

There are a couple test failures that I have started to track down. Hopefully it won't be too long before we merge 4.13rc, so I guess there is no rush.

- https://github.com/SimpleITK/SimpleITK/blob/next/SuperBuild/External_ITK.cmake#L50

blowekamp:
rebased onto master here:
#355

PR #355, opened 2017-12-15
Remove itk depricated

blowekamp:
This branch contains patches needed when configuring SimpleITK with the following options:
"-DITK_GIT_TAG=v4.13rc02 -DITK_FUTURE_LEGACY_REMOVE=ON"


Code/Common/src/sitkPimpleImageBase.hxx
```diff
@@ -98,7 +98,7 @@ namespace itk
 
         dup->SetInputImage( this->m_Image );
         dup->Update();
-        ImagePointer output = dup->GetOutput();
+        ImagePointer output = dup->GetModifiableOutput();
```
blowekamp: @hjmjohnson 

The remaining interface to me seems odd. It the `ImageDuplicator` trying to be like a filter with `Update` and `GetOutput`?

This class seems stuck in the middle for its expected behavior, and I am not sure there is a clear reference for expected behavior for this "non-filter". There are the `Calculators` but the do not seem like a similar model.

hjmjohnson:
@blowekamp Should we add a non-const "GetOutput" for the image duplicator to mimic the behavior of a filter?

That seems fine to me.  Could you please add the desired interface so that this old code works?

PR #362, opened 2017-12-29
Add objectness measures wrapping

blowekamp:


zivy (commented):
LGTM, interface satisfies what I expect.

PR #376, opened 2018-01-16
Enable SimpleITK_4D_IMAGES by default

### LINKED ISSUE #375: 4D .mhd functionality not default behaviour
Though this issue has been discussed [in the past](https://github.com/SimpleITK/SimpleITK/issues?utf8=%E2%9C%93&q=is%3Aissue+4d), I would like to suggest that 4D .mhd input be a default compilation flag so that if one needs the functionality, one doesn't have to build from source and then install "custom" libraries but rather us the default software.  I don't understand the point in turning this feature off by default.

blowekamp:
Enable IO and ExtractImageFilter to operate with 4D images. Older
versions of VS only have a 32-bit linker, so this option is disabled
with those older compilers.

closes #375

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -118,7 +118,13 @@ option( SimpleITK_INT64_PIXELIDS "Instantiate 64-bit integer pixels, including u
 sitk_legacy_naming(SimpleITK_INT64_PIXELIDS)
 
 
-option( SimpleITK_4D_IMAGES "Add Image and I/O support for four spatial dimensions." OFF )
+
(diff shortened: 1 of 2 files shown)


waltsims (commented):
🤔  Isn’t this still off after [lines 122 -127](https://github.com/blowekamp/SimpleITK/blob/a7c8d44bf2b932f3248c46eb8fb5646aa0af030c/CMakeLists.txt#L122-L127)? Furthermore if 4D is now a default option should it remain an advanced option?

PR #379, opened 2018-01-19
Use "build-and-test" job name for CDash reporting

blowekamp:
Use the convention of "build-and-test" for the job name so that the
GitHub CDash status reporting can occur.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,6 +1,6 @@
 version: 2
 jobs:
-  build:
+  build-and-test:
     docker:
       - image: circleci/python:2.7
     working_directory: ~/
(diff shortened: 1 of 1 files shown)


blowekamp:
@thewtex I have added the "build-and-test" job name to a new workflow. I believe this is what you were referring to as the "context".

The build name for a job is define by this line:
https://github.com/SimpleITK/SimpleITK/blob/0529038ffea93da8873cc955f4773c514c8758ed/circleci.cmake#L39

Is this compatible with the CDash reporting?

thewtex:
The `ITKGitHubCDashStatus` has been enabled on the `SimpleITK` organization. @blowekamp when a new pull request comes in, please let me know, and I will update and test the service for the SimpleITK dashboard.

blowekamp:
@thewtex Here is a pull request: #381

Can the service please be enabled?

Thanks.

thewtex:
@blowekamp All set.

Click the green checkmark next to `7f1f0fa` in #381 to verify.

Changes required were: https://github.com/InsightSoftwareConsortium/ITKGitHubCDashStatus/commit/d102e1aa0ada7ea279a89249ba405e731cf1b6ca

PR #381, opened 2018-01-24
Adding information on how to cite SimpleITK.

zivy:


blowekamp:
Great start! This should also be added to the read the docs too. How about the home page?

zivy:
Will add it to read-the-docs.

PR #390, opened 2018-02-05
Updating landing page prose.

zivy:


blowekamp:
Should we remove the paragraph before this one which describes read the docs as a "prototype"?

zivy:
The reason I left it there is that I'm not sure that the documentation is ready for prime time. On the other hand, I should probably take my own advice, remove the WIP warning and update the documentation as we go along. If we wait for it to be perfect we will never make it official.

Slightly off topic question:
I locally generated the documentation using the html_theme 'classic' instead of 'alabaster' (conf.py). I think it looks better, but this probably due to reading too much Python documentation. Is there a particular reason for choosing 'alabaster'?

PR #402, opened 2018-02-11
Added Java DicomSeriesReader example in documentation

ijm7:
Adds an implementation of the DicomSeriesReader example in Java. I've also updated the Documentation.rst file to accomodate this.


blowekamp:
Outstanding!

Can you please add this to the CMakeLists.txt file in the directory so that it can be compiled and tested along with the other examples.

blowekamp:
After a closer look, I had to make a couple changes to get the code to build and test, which I noted.

Also all pull requests or branches must NOT be based on the next branch. This topic should be rebased onto master via `git rebase --onto master HEAD~4  example-java-dicomseriesreader` The next branch is for integrating and testing.

blowekamp:
Thank you for the pull request and contribution!

I'll address the last couple of nitpicks to merge.

ijm7:
Thanks! My Git repo got a bit mixed up with merges so I wasn't able to commit the final fix.

blowekamp:
The topic has been merged into next now.

It should have showed up here:
http://simpleitk.readthedocs.io/en/next/Examples/DicomSeriesReader/Documentation.html

but there was some problem:
http://readthedocs.org/projects/simpleitk/builds/6760467/

PR #411, opened 2018-02-26
Use MultiThreader typedef for future ITK compatibility.

blowekamp:
The patch enable compilation with ITK 4.13 and the current development
version of ITK 5.0. The new version is changing the base multi-theader
class.


blowekamp:
@thewtex  @dzenanz
Just an FYI of a tweak needed for compatibility with the multi-threader changes.

dzenanz (approved):

PR #413, opened 2018-02-27
Add image io list

blowekamp:


Code/IO/src/sitkImageIOUtilities.cxx
```diff
@@ -0,0 +1,89 @@
...
+
+#include "itkImageIOBase.h"
+#include <sstream>
+#include <list>
+
+namespace itk {
+namespace simple {
+namespace ioutils {
+
+std::string PrintRegisteredImageIOs()
```
zivy: Possibly change signature to:
void PrintRegisteredImageIOs( std::ostream & out )

zivy (commented):
We may want to draw the developer's attention to the fact that there are cases where multiple registered ImageIO components support the same file extension. Maybe add a message highlighting this in the PrintRegisteredImageIOs method?

PR #418, opened 2018-03-08
COMP: Associate Swig module and SimpleITKTargets with install components

jcfr:
This will ensure that project integrating SimpleITK (e.g Slicer) and
installing components can differentiate between runtime and archive libraries.
This is important to avoid distributing larger archive libraries.

Change-Id: I4541c2acb1acdbba8e26201dcd52588b25bb7038


blowekamp:
Thanks for the pull request!

Let me check if the other wrappings have a similar issue.

Have you tried a 1.1 RC with Slicer yet? I tagged 1.1rc2 yesterday. I expect that the "SimpleFilters" module may need an update.

jcfr:
Thanks for the review.

> Have you tried a 1.1 RC with Slicer yet? I tagged 1.1rc2 yesterday. 

Does it still work against ITK 4.13 ? 

> I expect that the "SimpleFilters" module may need an update

Roughly, what do you think would need to change ?

blowekamp:
Hmm.. You hit the green merge button? We are still using the legacy Kitware stage mechanism. I merged the topic, and I anticipate the robot will force push the new next.

> Does it still work against ITK 4.13 ?
Yes, it does, and it's compiling as well as expected against the ITK master too.

> Roughly, what do you think would need to change ?

This on me. I suspect just about every thing should still work with the old JSON in the module. But the JSON should be updated to the latest SimpleITK interfaces. The one stumbling block is the "newer" usage of named and optional inputs (images) for filters. I am not sure if that is supported at all by the Slicer module.

jcfr:
Ooops. 

You could may be restrict PR merge and/or add a CONTRIBUTING.md or pull request template in a `.github` directory 

>  I anticipate the robot will force push the new next.

Yeap. It should

> This on me.

Great. And thanks for the detailed explanation.

In the mean time, I now have a staging branch on https://github.com/Slicer/SimpleITK with the commit fixing the components.

jcfr:
Interestingly, I initially pushed to the stage .. but then, I checked if the project add PRs and then thought the project switched to GitHub ... but thinking more about it .. if that would be the case, the `gerrit-push` common would not be available. Arff

blowekamp:
We have been waiting for "ITK" to implement it's GitHub procedure to follow their lead. Currently, we are in-between trying to support contributions how users are familiar with. Only a few see the merge button on PR, so it is restricted if that is what you mean.

jcfr:
Makes sense.

PR #424, opened 2018-03-15
sitkLanguageOptions: Improve setting of default value for wrap options

jcfr:
This commit makes sure no calls to "find_package(<languageName>)" are
done if the option WRAP_<languageName> is explicitly set to OFF.

Indeed, systematically calling "find_package()" can have side-effect.
See https://discourse.slicer.org/t/nightly-mac-slicer-4-9-0-2018-03-13-macosx-amd64-does-not-open/2313/7

Change-Id: If58a333b29feef5055f03a4612443020e2a6afa7

### DIFF
--- a/CMake/sitkLanguageOptions.cmake
+++ b/CMake/sitkLanguageOptions.cmake
@@ -15,25 +15,15 @@ sitk_check_dynamic_lookup(MODULE
   SITK_UNDEFINED_SYMBOLS_ALLOWED
   )
 
+if ( SITK_UNDEFINED_SYMBOLS_ALLOWED )
+  set( _REQUIRED_LIBRARY "QUIET" )
+else ( )
+  set( _REQUIRED_LIBRARY "REQUIRED" )
+endif ( )
+
+
 option(WRAP_DEFAULT "The default initial value for wrapping a language when it is detected on the system." ON)
 mark_as_advanced(WRAP_DEFAULT)
-
-#
-# Macro to set "_QUIET" and "_QUIET_LIBRARY" based on the first
-# argument being defined and true, to either REQUIRED or QUIET.
-#
-macro(set_QUIET var)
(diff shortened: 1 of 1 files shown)


jcfr:
An alternative option would be to ensure option like `CMAKE_DISABLE_FIND_PACKAGE_<packageName>` are propagated down to the inner build.

jcfr:
The commit message in the corresponding Slicer commit should give more context:

```
COMP: update SimpleITK to effectively avoid unwanted find_package calls

This commit reverts the approach implemented in the previous commit. It
failed because SimpleITK has a superbuild structure and the CMAKE_DISABLE_FIND_PACKAGE_*
options are not propagated down to the inner build.

Instead, a patch improving the handling of SimpleITK wrapping options
has been proposed to the upstream project (and in the mean time integrated
in our Slicer fork).

The implemented solution consists in avoiding the call to "find_package()"
if the corresponding WRAP_<languageName> option has explicitly been set or
if WRAP_DEFAULT is set to OFF.
Indeed, there is no need to do these call to find an appropriate default value
for the option since the user (in this case Slicer build system) explicitly
set of value of for the options (eiter directly or by setting WRAP_DEFAULT).

See https://github.com/SimpleITK/SimpleITK/pull/424

List of changes:

$ git shortlog 67872f8..998bdbc --no-merges
Jean-Christophe Fillion-Robin (1):
      sitkLanguageOptions: Improve setting of default value for wrap options


```

blowekamp:
Great! Thanks for the pull request! The

This is some tricky logic and behavior, give me a bit to try it out for a few cases. It looks like this approach will results in multiple calls to the `find_packages` and there could be a macro to reduce some code duplications. I'll work on a refined patch.

blowekamp:
Please see #427  for the proposed improvement. This patch will only call find_package once, and there is some code reduction by keeping the macro.

PR #426, opened 2018-03-15
Propagate CMAKE_DISABLE_FIND_PACKAGE_* to SimpleITK superbuild

blowekamp:
Calling find_package for all languages can have side effects, such as
on OSX, displaying a prompt to install Java. There must be a way to
disable searching for wrapped languages. This patch enables a standard
CMake variable to disable finding languages.

Suggested-by: Jean-Christophe Fillion-Robin <jchris.fillionr@kitware.com>

See #424 for suggestion


jcfr (approved):
That will work too but it defeat the purpose of the `WRAP_<languageName>` option.

If explicitly set, no call to find_package are expected.

PR #427, opened 2018-03-15
sitkLanguageOptions: Improve setting of default value for wrap options

blowekamp:
This commit makes sure no calls to "find_package(<languageName>)" are
done if the options WRAP_DEFAULT or WRAP_<languageName> is explicitly
set to OFF.

Indeed, systematically calling "find_package()" can have side-effects.
See https://discourse.slicer.org/t/nightly-mac-slicer-4-9-0-2018-03-13-macosx-amd64-does-not-open/23

Coauthored-by: Jean-Christophe Fillion-Robin <jchris.fillionr@kitware.com>

### DIFF
--- a/CMake/sitkLanguageOptions.cmake
+++ b/CMake/sitkLanguageOptions.cmake
@@ -1,11 +1,12 @@
 #
-# - This module finds the languages supported by SimpleITK, and
-#present the option to enable support
-#
-
-#
-# Currently this will search for Python, Java, TCL, Ruby, C#, R
-# Additionally it give the option to wrap LUA.
+# - This module finds the languages supported by SimpleITK. By default
+#a find package will be performed for each language, and if it is
+#detected wrapping for that language will be enabled
+#(WRAP_<languageName>=ON). This behavior can be overridden by setting
+#"WRAP_DEFAULT" to FALSE. Individual languages can be disabled with
+#the WRAP_<languageName> option. If either WRAP_DEFAULT or WRAP_<languageName> is
+#false then no find_package for the language will be performed, and
+#the language will not be wrapped.
 #
 
 include(sitkTargetLinkLibrariesWithDynamicLookup)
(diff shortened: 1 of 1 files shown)


CMake/sitkLanguageOptions.cmake
```diff
@@ -15,23 +16,38 @@ sitk_check_dynamic_lookup(MODULE
...
+# WRAP_DEFAULT is "OFF"
+#
+macro(set_DEFAULT_QUIET var)
+  set( _QUIET "QUIET" )
+  if ( DEFINED  ${var} )
+    if ( ${var} )
+      set( _QUIET "REQUIRED" )
+    else()
+      # ${var} is OFF
+      set(${var}_DEFAULT ${var} )
```
jcfr: I suggest to set it to `OFF`  then. It facilitates reading of the code i think .

CMake/sitkLanguageOptions.cmake
```diff
@@ -15,23 +16,38 @@ sitk_check_dynamic_lookup(MODULE
...
+# if ${var} is defined and true then _QUIET will be set to "REQUIRED"
+# otherwise it'll be "QUIET".
+#
+# ${var}_DEFAULT is set to "OFF" only if ${var} is already OFF, or if
+# WRAP_DEFAULT is "OFF"
+#
+macro(set_DEFAULT_QUIET var)
+  set( _QUIET "QUIET" )
+  if ( DEFINED  ${var} )
+    if ( ${var} )
```
jcfr: I suggest to add this comment:
```
# Variable `WRAP_<languageName>` is defined and TRUE, we need
# to call  `find_package` with `REQUIRED`
```

CMake/sitkLanguageOptions.cmake
```diff
@@ -15,23 +16,38 @@ sitk_check_dynamic_lookup(MODULE
...
+      set( _QUIET "REQUIRED" )
+    else()
+      # ${var} is OFF
```
jcfr:
```
# Variable `WRAP_<languageName>` is defined and FALSE, we do  *NOT* need
# to call  `find_package` at all to initialize the `WRAP_<languageName>_DEFAULT`
# variable
```

CMake/sitkLanguageOptions.cmake
```diff
@@ -15,23 +16,38 @@ sitk_check_dynamic_lookup(MODULE
...
+      set(${var}_DEFAULT ${var} )
+    endif()
+  elseif( NOT ${WRAP_DEFAULT} )
+    # languages are globally disabled
```
jcfr:
```
# The variable `WRAP_<languageName>` is *NOT* defined and by default all
# languages are disabled, we do *NOT* need to call `find_package` to initialize
# the `WRAP_<languageName>_DEFAULT` variable.
```

CMake/sitkLanguageOptions.cmake
```diff
@@ -42,13 +58,12 @@ endmacro()
...
-  set( WRAP_LUA_DEFAULT ${WRAP_DEFAULT} )
-else()
-  set( WRAP_LUA_DEFAULT OFF )
+if( NOT DEFINED WRAP_LUA_DEFAULT OR WRAP_LUA )
+  find_package ( Lua ${_QUIET} )
+
+  set( WRAP_LUA_DEFAULT ${LUA_FOUND} )
 endif()
 
 set( LUA_ADDITIONAL_LIBRARIES "" CACHE STRING "Additional libraries which may be needed for lua such as readline.")
```
jcfr: Look like this call can be moved within the `if(WRAP_LUA)`, I don't think defining the variable `LUA_ADDITIONAL_LIBRARIES` makes sense otherwise

CMake/sitkLanguageOptions.cmake
```diff
@@ -42,13 +58,12 @@ endmacro()
...
 
-set_QUIET( WRAP_LUA )
-find_package ( Lua ${_QUIET} )
+set_DEFAULT_QUIET( WRAP_LUA )
 
-if ( LUA_FOUND )
-  set( WRAP_LUA_DEFAULT ${WRAP_DEFAULT} )
-else()
-  set( WRAP_LUA_DEFAULT OFF )
+if( NOT DEFINED WRAP_LUA_DEFAULT OR WRAP_LUA )
```
jcfr: Instead of checking variable `WRAP_<languageName>` and `WRAP_<languageName>`, I suggest to have a variable named `SITK_<languageName>_DO_FIND_PACKAGE` (or similar), then code  would simply be:
```
if( SITK_<languageName>_DO_FIND_PACKAGE )
```

the macro `set_DEFAULT_QUIET` could then be renamed too `sitkShouldDoFindPackageCall(<languageName>)` and set the two variables:
* ` SITK_<languageName>_DO_FIND_PACKAGE`
* ` SITK_<languageName>_FIND_PACKAGE_EXTRA_ARGS`

CMake/sitkLanguageOptions.cmake
```diff
@@ -72,19 +87,23 @@ endif()
...
+  else ()
+    find_package ( PythonLibs ${_QUIET_LIBRARY} )
+  endif()
+
+  if ( PYTHONLIBS_FOUND AND PYTHONINTERP_FOUND
+      AND (PYTHON_VERSION_STRING VERSION_EQUAL PYTHONLIBS_VERSION_STRING) )
+    set( WRAP_PYTHON_DEFAULT ON )
+  else()
+    set( WRAP_PYTHON_DEFAULT OFF )
+  endif()
```
jcfr: I think the check below:

```
if ( WRAP_PYTHON AND PYTHON_VERSION_STRING VERSION_LESS 2.7 )
...
endif()
```

can be moved inside the `if(WRAP_PYTHON)`

blowekamp:
Thank you for your review and great comments!

blowekamp:
Thanks again for the review, these improvements make this code much clearer and easier to read. I opted for using some `_local_varaiables` is stead of the initial suggested names.

CMake/sitkLanguageOptions.cmake
```diff
@@ -15,24 +16,53 @@ sitk_check_dynamic_lookup(MODULE
   SITK_UNDEFINED_SYMBOLS_ALLOWED
   )
 
-option(WRAP_DEFAULT "The default initial value for wrapping a language when it is detected on the system." ON)
+option(WRAP_DEFAULT "The default value for wrapping a language." ON)
```
blowekamp: "Enables automatic detection of all languages and enables language wrapping if available."

jcfr:
`Coauthored-by:` -> `Co-authored-by:`

I will test shortly

jcfr:
And well done, it looks much easier to read and understand now :+1:

PR #428, opened 2018-03-16
Adding wrapping for Watershed and Toboggan filters

blowekamp:


blowekamp:
The Toboggan filter has already been added to SimpleITK. As there is also the MorphologicalWatershedImage filter this PR is not needed.

PR #432, opened 2018-03-26
Adding support for vector images in as.image conversion.

zivy:
The as.image operation in R previously only supported scalar
images. The change in this commit adds a boolean isVector similar to
the Python implementation so that the given R array is interpreted as
a scalar or vector image accordingly.


blowekamp:
Can you please add a test?

blowekamp:
Great! Thanks you for adding tests!

PR #447, opened 2018-04-20
Updating the read the docs documentation.

zivy:
Adding a page on the fundamental concepts underlying the image and
transformation classes. Also modified the sphinx conf.py to use the
math extension (support latex formulas as png).


blowekamp:
Awesome! I am going to merge this and see how it renders. Then do any fixes in additional patched.

zivy:
Should I create a new branch with the content and get rid of this branch? I don't think we want the svg files in the repository's history and I already committed them, so removing them won't make a difference.

blowekamp:
Great! I'm going to quickly merge this into next, then master after checking the ReadTheDocs build.

blowekamp:
The RTD build seems successful, there are a couple things things need fixed such as the "How do you determine the physical location of a pixel:" code block. But this topic is being merging into master, and new fixes can be based on maser.

PR #451, opened 2018-04-24
Fix fundamental concepts code block

blowekamp:


blowekamp:
This patch is no longer needed.

PR #464, opened 2018-05-09
Added test of default ResampleImageFilter params

dave3d:
The new test creates a small image, and executes a ResampleImageFilter
on the image with the default filter parameters.  Right now this
test will always fail, because it seg faults when size == 0.


blowekamp:
SimpleITK master has updated it's ITK version with a fix so itk::ResampleImageFilter can produce images of size 0. Can this patch please be rebased, and updated.

blowekamp:
Going to rebase/squash the changes from this patch into a new one.

PR #467, opened 2018-05-10
Add ability to explicitly set ImageIO for readers

blowekamp:


Code/IO/include/sitkImageFileReader.h
```diff
@@ -42,6 +42,14 @@ namespace itk {
...
      * with the ReadImageInformation method.
      *
+     * Reading takes place by the ITK ImageIO factory mechanism. ITK
+     * contains many ImageIO classes which are responsible for reading
+     * separate file formats. By default, each ImageIO is asked if it
+     * "can read" the file, and the first one which can read the
+     * format is used. The list of available ImageIO can be obtained
+     * with the GetRegisteredImageIOs method. The ImageIO used can be
+     * overridden with the SetImageIO method.
+     *
```
zivy: Possibly change text starting at "The list..." to say:
The list of available ImageIOs can be obtained using the GetRegisteredImageIOs method. The user can request that a specific IO be used with the SetImageIO method. This is useful in cases when multiple ImageIOs "can read" the file and the user wants to select a specific IO (not the first).

Code/IO/include/sitkImageReaderBase.h
```diff
@@ -72,12 +72,28 @@ class SmartPointer;
...
+       *
+       * An option to override the automatically detected ImageIO used
+       * to read the image. The available ImageIOs are listed by the
+       * GetRegisteredImageIOs method. If the ImageIO can not be
+       * constructed an exception will be generated. If the ImageIO
+       * can not read the file an exception will be generated.
+       *
+       * By the value the value is and empty string "". This default
+       * value indicates that the proper ImageIO will be automatically determined.
+       */
```
zivy: "By the value..." , sentence doesn't make sense to me. Should this read:
The default value is the empty string (""). This indicates that the ImageIO will be determined automatically.

Code/IO/src/sitkImageReaderBase.cxx
```diff
@@ -78,8 +81,26 @@ itk::SmartPointer<ImageIOBase>
 ImageReaderBase
 ::GetImageIOBase(const std::string &fileName)
 {
-  itk::ImageIOBase::Pointer iobase =
-    itk::ImageIOFactory::CreateImageIO( fileName.c_str(), itk::ImageIOFactory::ReadMode);
+  itk::ImageIOBase::Pointer iobase;
+  if (this->m_ImageIOName == "")
+    {
+    iobase = itk::ImageIOFactory::CreateImageIO( fileName.c_str(), itk::ImageIOFactory::ReadMode);
+    }
```
zivy: Should this setting also have a side effect of updating the m_ImageIOName using the GetNameIOfClass method, or should it leave the IO name as the empty string so that reading another file with different extension will work too. I think the later (current behavior) is expected, but not sure.

zivy (commented):
LGTM. Some minor comments.

dave3d:
Zivy's comments make sense to me.  Otherwise looks good.

dave3d (approved):

PR #468, opened 2018-05-11
Add image io region extract stream

blowekamp:


blowekamp:
Still need to add a test reading from a 5-D file.

dave3d (approved):
Looks good.

Code/IO/include/sitkImageFileReader.h
```diff
@@ -138,6 +138,39 @@ namespace itk {
...
+       * By default the reader loads the entire image, this is
+       * specified when the size has zero length.
+       *
+       * If specified, then the image returned from `Execute` will be
+       * of this size. If the ImageIO and file support reading just a
+       * region, then the reader will perform streaming.
+       *
+       * The dimension of the image can be reduced by specifying a
+       * dimension's size as 0. For example a size of $[10,20,30,0,0]$
+       * results in a 2D image with size of $[10,20,30]$. This
```
zivy: 3D image, not 2D

Code/IO/include/sitkImageFileReader.h
```diff
@@ -138,6 +138,39 @@ namespace itk {
...
+       * enables reading a 5D image into a 3D image. If the size is
+       * larger or more dimensions than the image in the file, an
+       * exception will be generated. Missing dimensions are treated
```
zivy: Missing dimensions only occur with a shorter 'size' vector, so possibly say it explicitly? Something like:
If the size vector is shorter than the original image dimension then the missing values are assumed to be zero.

zivy (commented):
Good and useful added functionality and interface.
Minor comments on documentation.

PR #471, opened 2018-05-21
Set pointer to null after deleting

blowekamp:
ITKv5 multi-threading update cases multiple calls for update to 1.0
progress. The reveled a problem in the test which could try to delete
m_Cmd multiple time. Follow best practices and set pointers to deleted
memory to null.


blowekamp:
@dzenanz Random fix due to change in threading/progress behavior.

dzenanz (approved):

PR #474, opened 2018-05-25
Updated the link for the tutorial

seungkyoon:


blowekamp:
Great! Thanks for fixing the broken link!

blowekamp:
Your author information is not correct, and need to be updated before this patch can be merged.

PR #488, opened 2018-06-13
Cran preparations main

richardbeare:
Some changes resulting from my efforts to make the package pass the tests.

Vignette had some old indexing examples that are no longer legal.

A bug fix at about R 3.2 has changed how the "methods" package should be imported

I need to do more about explicitly exporting stuff from the package, as there is a lot that is used internally. Some of the issues may be dealt with by the optional namespace file swig can create, but it is somewhat broken and will require post processing.


blowekamp:
These changes look good. Could you please rebase the topic onto the v1.1.0 release, so that it can easily be merged into the release branch. This should be a able to be done with:
```
git fetch origin
git checkout CRANPreparationsMain
git rebase --onto v1.1.0 origin/master CRANPreparationsMain
git push -f mygithub HEAD
```

richardbeare:
Just rebased and pushed.

blowekamp:
Richard,

I thought I merged this topic last week 😕 Thanks for working on it.

The commit message with these namespace changes are rather brief. It is not clear to me what issue they are addressing, and why there are needed. Can you please document this information in the commit message so the community understand the motivation for the changes more. Thanks!

richardbeare:
Updated the commit message.

Aim is to avoid exporting everything. Among other things, this will help with getting the package
to pass the CRAN checks as there will be fewer undocumented objects (only exported functions require documentation).

blowekamp:
All but the most recent patch was merged into next.

Unfortunately, After I added the R testing to Circle CI (see #490 ), the R tests are segfaulting: https://open.cdash.org/viewTest.php?onlyfailed&buildid=5425967

Do you expect this latest patch to address that issue? Do think this topic is ready to go?

richardbeare:
All the tests pass on my local machine. I went through some iterations to get the combinations of exports correct. I will test those updates locally.

richardbeare:
I see the problem on circleCI. Not sure how to get access to the correct circleci project to begin  debugging. Still have the old installer projects, but can't trigger builds under the SimpleITK banner due to permission issues.

blowekamp:
Have you tried enable CircleCI for you fork?

richardbeare:
OK - enabled for my fork, but the build is timing out. Is there some way I can access SimpleITK resources, because at the moment I'm running under the hobby plan. Apologies - I don't know my way around the circleCI interface yet.

blowekamp:
Have you tried building locally on Linux with shared libraries?

richardbeare:
I haven't tried linux for a few days, but it was passing all R tests before I moved my testing to the laptop (while travelling) and it passes all on the laptop. I'm thus assuming that there is something funny about how the circleci testing is configured, and need to reproduce those failures.

blowekamp:
I am suggesting it is likely the shared library option. I have you tried that?

richardbeare:
Currently firing up a VM to try it out.

richardbeare:
VM build under ubuntu 17 appears to pass all tests when building with shared libraries on. I had to set LD_LIBRARY_PATH before the build, otherwise the package won't install under R, but it looks like you're already doing that in the circleci configuration. Am I testing the right thing?

``` bash
ctest -j4 -R "^R\\..+"

100% tests passed, 0 tests failed out of 519

Label Time Summary:
R    = 1740.70 sec*proc (519 tests)

```
``` bash
git branch
  master
* next

git log
commit 69e6b69e1c8569f958535539d989af26775323da (HEAD -> next, origin/next)
Merge: f4d34ee0 5d0811bd
Author: Bradley Lowekamp <blowekamp@mail.nih.gov>
Date:   Fri Jun 22 14:33:17 2018 -0400

    Merge topic 'UpdateITKv4.13release' into next
    
    5d0811bd Update ITK superbuild version along ITK v41.3 release

```

blowekamp:
Hmm... Have you used Docker?

The following "Dockerfile" can reproduce the problem:
```
FROM ubuntu:16.04

RUN apt-get update && \
    apt-get install -y lua5.1 kwstyle swig r-base r-base-dev python2.7-dev python-pip git && \
    pip install cmake
RUN git clone https://github.com/SimpleITK/SimpleITK.git &&\
     cd SimpleITK &&\
     git checkout next

RUN mkdir bld && cd bld &&\
    cmake \
    -DBUILD_DOCUMENTATION:BOOL=OFF \
    -DBUILD_EXAMPLES:BOOL=OFF \
    -DSHARED_LIBS:BOOL=ON \
    -DSimpleITK_USE_SYSTEM_LUA:BOOL=ON \
    -DSimpleITK_USE_SYSTEM_SWIG:BOOL=ON \
    -DSimpleITK_EXPLICIT_INSTANTIATION:BOOL=OFF \
    -DWRAP_DEFAULT:BOOL=OFF \
    -DWRAP_R:BOOL=ON \
    ../SimpleITK/SuperBuild

RUN cd bld &&\
    export LD_LIBRARY_PATH=/bld/ITK-prefix/lib:/bld/SimpleITK-bld/lib &&\
    make -j $(nproc) 

ENTRYPOINT cd bld/SimpleITK-build && ctest -j $(nproc)
```

You are use it like this:
```
docker build -t simpleitk_r -f Dockerfile . # builds the system, and compiles sitk
docker run simpleitk_r # runs the default entry point which runs test
docker run -it --entrypoint=/bin/bash simpleitk_r # give shell access to run tests/debug
```

Interestingly, when the upstream docker image was `ubuntu:17.10` all the tests passed. So that confirms your results, and raises questions what version difference is causing the problem?

richardbeare:
The default R under ubuntu 16 is only 3.2, which is exactly where that methods package bug was around. It seems to work OK if you enable updated versions, as below. The installer tests were doing something similar (for debian) to get the updated R

``` bash
RUN apt-get update && \
    apt-get -y install software-properties-common apt-transport-https && \
    add-apt-repository -y "deb https://cloud.r-project.org/bin/linux/ubuntu xenial-cran35/" && \
    apt-key adv --keyserver keyserver.ubuntu.com --recv-keys E084DAB9 && \
    apt-get update && \
    apt-get install -y lua5.1 kwstyle swig r-base r-base-dev python2.7-dev python-pip git && \
    pip install cmake
```

blowekamp:
So this new code should be conditional on the version of R?

Do want to make a pull request to update the circleCI?
https://github.com/SimpleITK/SimpleITK/blob/master/.circleci/config.yml#L175

richardbeare:
It gets very tricky. Should we support versions of R that are 3 years old?
We can set a requirement for R > 3.3. Also, I think the issue is a
combination of the description file and the namespace file, with the
mechanism of methods use differing. I dont think the new code that limits
the number of exported functions is the issue.

blowekamp:
What version of R is on cantos and Ubuntu long term support?

On Jun 28, 2018, at 4:12 PM, Richard Beare <notifications@github.com<mailto:notifications@github.com>> wrote:

It gets very tricky. Should we support versions of R that are 3 years old?
We can set a requirement for R > 3.3. Also, I think the issue is a
combination of the description file and the namespace file, with the
mechanism of methods use differing. I dont think the new code that limits
the number of exported functions is the issue.

richardbeare:
Cran currently has install mechanisms for 3.5 for the long term support.
They appear to be trailing official R releases by a few weeks.
Will investigar centos.

richardbeare:
[CRAN](https://cran.r-project.org/bin/linux/ubuntu/README.html) instructions for newer versions of R under ubuntu.

Default versions (tested with docker) 
Ubuntu version : R version 
14.04.5 - Trusty Tahr    : 3.0.2
16.04.4 - Xenial Xerus  : 3.2.3
18.04 - Bionic Beaver   :  3.4.4

My opinion is that we shouldn't worry too much about the base versions of R, especially as we require upgrades of the build tools anyway (i.e cmake). There are official sources of newer versions of R for ubuntu and my reading suggests that the situation is similar for redhat derived distributions. I'll confirm this next.

richardbeare:
Under centos it appears necessary to access R via the epel repositories, and under centos 6.9, which is about to be the one under long term release, R 3.5.0 is installed.

blowekamp:
OK, so we are only going to support the newer versions of R.
- This needs to be documented in the read the docs/build requirements
- A configuration check, for the proper version of R needs to be made.

Also, do you have some type of reference or link to the R importing change/bug/expected behavior?

richardbeare:
Reference for the change in behaviour:

http://r-pkgs.had.co.nz/description.html,  the "other dependencies" section.
http://r-pkgs.had.co.nz/namespace.html , the S4 section at the end of the
page mentions the pre/post 3.2 differences.

I'll update this branch further to deal with the configuration checks. Will
also update the installer to require the correct R.

blowekamp:
Cool! We should also check the R version during configuration. It may save some one hours for time trying to compile with the wrong version of R. You can find the version check done for Python [here](https://github.com/SimpleITK/SimpleITK/blob/master/CMake/sitkLanguageOptions.cmake#L138). R is at the end if the file.

richardbeare:
OK - I've made what is probably a dodgy attempt at implementing the R_VERSION_STRING components.

I've also bumped the version of R in the installer, also on the CRANPreparations branch.

CMake/FindR.cmake
```diff
@@ -37,7 +39,7 @@ set(R_LIBRARIES ${R_LIBRARY_BASE})
 mark_as_advanced(RSCRIPT_EXECUTABLE R_LIBRARIES R_INCLUDE_DIR R_COMMAND R_LIBRARY_BASE)
 
 
-set( _REQUIRED_R_VARIABLES R_INCLUDE_DIR R_COMMAND )
+set( _REQUIRED_R_VARIABLES R_INCLUDE_DIR R_COMMAND R_VERSION_STRING)
```
blowekamp: R_VERSION_STRING should not be listing in the require variables.

Line 50 needs to be updated. See https://cmake.org/cmake/help/v3.12/module/FindPackageHandleStandardArgs.html.

We need to switch to the second signature. Something like:

```
find_package( R
 REQUIRED_VARS ${_REQUIRED_R_VARIABLES} 
 VERSION_VAR R_VERSION_STRING )
```

CMake/sitkLanguageOptions.cmake
```diff
@@ -302,6 +302,10 @@ endif()
 option ( WRAP_R "Wrap R" ${WRAP_R_DEFAULT} )
 
 if ( WRAP_R )
+  message("R version: ${R_VERSION_STRING} ")
```
blowekamp: Please remove the extra message

richardbeare:
The latest changes appear to work on my test config. I have rebased the last two commits and force pushed.

Need to update the circleci config to include more recent R

blowekamp:
This pull request is running in your CircleCI account which is not using the full large system requests. So the CI is being terminated due to time out because it is using too many resources.

I recommend disabling CircleCI for you SimpleITK repo. You are welcome to update this pull request to use SimpleITK CI.

richardbeare:
I have disabled testing under my account. How do I start using SimpleITK CI?

blowekamp:
Just make a pull request to a SimpleITK repo and you will be using our account.

PR #502, opened 2018-06-27
Use standard find_package for ITK in SimpleITK Config

blowekamp:
To properly import ITK's targets and it's transitive dependencies, use
the find_package(ITK). This issues was detected when enabling IODCMTK,
and the examples failing to link properly.


blowekamp:
@thewtex @dzenanz This enables building against ITKv5 with TBB. The TBB libraries are not picked up properly.

dzenanz:
It is good that you found the issue!

PR #507, opened 2018-07-10
Convert pngs to web-optimized jpgs

seungkyoon:


blowekamp:
Great! Merging into next.

PR #515, opened 2018-07-17
R package rules don't allow non standard file formats in the data dir…

richardbeare:
…ectory.

This package relocates images to the extdata folder.

There may also be a need to change notebooks, if they use the sample
images.


blowekamp:
The [SimpleITK-Notebooks](https://github.com/InsightSoftwareConsortium/SimpleITK-Notebooks) have their own data, so they should not be using this.

Looks good, merging into next for testing.

PR #524, opened 2018-07-24
formatting of license field so CRAN is happy

richardbeare:
```r 
R CMD check 
```
is strict about the formatting of the license text. Leaving out "License" leads to "Unknown license" errors.


blowekamp:
This looks like we need to add some type of test or check to out build system to validate these things...

richardbeare:
Updated the commit message, also in the installer

PR #525, opened 2018-07-24
Vignette index entry in latex

richardbeare:
Indexing meta data in the latex preamble


blowekamp:
Please provide a good commit message. Here is some suggested reading:
https://chris.beams.io/posts/git-commit/
https://git.kernel.org/pub/scm/git/git.git/tree/Documentation/SubmittingPatches?id=HEAD#n133

First line should use the imperative tense ( Add, Remove, Update etc) on the action done. Then a brief explanation on how and why.

Perhaps something like this:
```
Add R vignette index entry in later

The satisfies the CRAN requirement blah blah....
```

richardbeare:
Updated.

PR #530, opened 2018-07-25
Remove internal function in ImageRegistrationMethod1

seungkyoon:
Replace internal commandIteration function with in-line lambda-like
function. Also rename title to match styling of similar examples.


blowekamp:
topic merged to next

PR #534, opened 2018-07-27
circleci: Fix and simplify install of CMake

jcfr:
Since using the "install_cmake.py" scikit-ci-addon is not supported
on CircleCI 2.0, this commit directly install the CMake wheel.


blowekamp:
Thanks for the prompt fix. We are still using the Kitware topic stage repo to merging.

jcfr:
Ooops. I fall in the trap each time ...

PR #536, opened 2018-07-30
BUG: Directly use pip to install CMake - for release

blowekamp:
Using the "install_cmake.py" scikit-ci-addon is not supported on
CircleCI 2.0.

ref: scikit-build/scikit-ci-addons#69


blowekamp:
release and master have a fix for this issue

PR #547, opened 2018-08-07
Text output in process obj

dave3d:


blowekamp:
Thank you! merging into next...

blowekamp:
I think there is something back with the ccache for this PR number on CircleCI. I have merged the fix into the next branch.

PR #548, opened 2018-08-07
R packaging imrovements

blowekamp:


richardbeare:
This looks good. The R package checks pass, so we are close to a CRAN submission too.

Let me know about tarballs for JSS.

blowekamp:
topic merged into next...

PR #550, opened 2018-08-14
Fix: spurious bracket in script

richardbeare:
The SourceTarball script had a typo leading to errors when downloading test data


blowekamp:
Topic merged into next...

PR #557, opened 2018-08-22
ENH: Improving formatting of DESCRIPTION file 

richardbeare:
R CMD check --as-cran

complains about title formatting (case) and the use of "This package" at the start of description.


blowekamp:
topic merged into next

PR #562, opened 2018-09-04
Updatind publication citation list.

zivy:
Updating the JDI and JSS publications.


blowekamp:
topic merged into next

PR #566, opened 2018-09-07
Add mirror of SimpleITK External Data repository

blowekamp:
This mirror is utilized for CMake External Data


blowekamp:
merged into next

PR #573, opened 2018-09-18
Fix for older Visual Studio

dave3d:
Tried to initialize a vector with a list.  Can't do that in
the older VS.


blowekamp:
Thank you for fixing, but we we are going to go with PR#574.

Issue #572, opened 2018-09-18
Failure compiling sitkImageViewer in windows with Rtools\mingw_64\bin\c++


richardbeare:
Compilation problems with ImageViewer in windows using the R toolchain

```bash
FAILED: Code/IO/src/CMakeFiles/SimpleITKIO.dir/sitkImageViewer.cxx.obj
C:\Rtools\mingw_64\bin\c++.exe  -DITK_IO_FACTORY_REGISTER_MANAGER -DSITK_USE_EXPLICITITK -ICode/IO/src/ITKFactoryRegistration -IE:/S/SimpleITK/Code/IO/include -IE:/S/SimpleITK/Code/IO/src -IE:/S/SimpleITK/Code/Common/include -ICode/Common/include -IE:/S/SimpleITK/Code/Explicit/include -isystem E:/S/B/ITK-prefix/include/ITK-4.13 -Wa,-mbig-obj  -mthreads  -mthreads -O3 -DNDEBUG -fvisibility=hidden -fno-keep-inline-dllexport   -Wa,-mbig-obj -Wno-long-long -Wno-unused-local-typedefs -Wno-strict-overflow -Wextra -Wformat=2 -Wno-format-nonliteral -Wunused -Wpointer-arith -Winvalid-pch -Wcast-align -Wdisabled-optimization -Woverloaded-virtual -Wshadow -Wwrite-strings -Wstrict-null-sentinel -Wno-invalid-offsetof -MD -MT Code/IO/src/CMakeFiles/SimpleITKIO.dir/sitkImageViewer.cxx.obj -MF Code\IO\src\CMakeFiles\SimpleITKIO.dir\sitkImageViewer.cxx.obj.d -o Code/IO/src/CMakeFiles/SimpleITKIO.dir/sitkImageViewer.cxx.obj -c E:/S/SimpleITK/Code/IO/src/sitkImageViewer.cxx
E:/S/SimpleITK/Code/IO/src/sitkImageViewer.cxx: In static member function 'static void itk::simple::ImageViewer::initializeDefaults()':
E:/S/SimpleITK/Code/IO/src/sitkImageViewer.cxx:151:93: error: in C++98 'win_dirs' must be initialized by constructor, not by '{...}'
   std::vector<std::string> win_dirs = { "PROGRAMFILES", "PROGRAMFILES(x86)", "PROGRAMW6432" };
                                                                                             ^
[...]

PR #574, opened 2018-09-19
Fix: C++11 style init causing problems with RTools(windows) c++

richardbeare:
issue #572 

Looks like you've already done it. in #573


blowekamp:
Thanks for the fix! Merging to next.

PR #587, opened 2018-11-08
Add support for vector images to the clamp image filter

blowekamp:
Compatibility note: this changes the default output type from
defaulting to float32 to matching the input type.


blowekamp:
topic merged to next.

PR #589, opened 2018-11-08
Add image registration get metric sampling percentage

### LINKED ISSUE #584: Add wrapper for GetMetricSamplingPercentagePerLevel
The ImageRegistrationMethod class has no method for getting the 'metric sampling percentage per level' in SimpleITK.  This function exists in the base C++ ITK, but no wrapper was made to port it to SimpleITK.  

The corresponding Set function has already been implemented.

I found this out while trying to implement tests, as part of a workflow where the user performs dozens of registrations in a row.  I'd intended to allow the user to change the sampling metric, but found that I couldn't verify it actually changing to the correct value in a test, and that I couldn't print out the current value to the user.  

This seems like it should be a relatively simple wrapper to implement, as it's just returning a vector double list.

blowekamp:
closes #584

### DIFF
--- a/Code/Registration/include/sitkImageRegistrationMethod.h
+++ b/Code/Registration/include/sitkImageRegistrationMethod.h
@@ -452,10 +452,18 @@ namespace simple
 
     /** \brief Set percentage of pixels sampled for metric evaluation.
      *
+     * The percentage is of the number of pixels in the virtual domain
+     * per level after the shrink factor has been applied.
+     *
      * The seed parameter is used to seed the pseudo-random number
      * generator used in generating the sampling set of points. If the
      * seed parameter is 0, then the wall clock is used, otherwise the
      * fixed seed is used for reproducible behavior.
+     *
+     * When providing multiple values the number of entries must match
+     * the number of shrink factors and smoothing sigmas. If a single
+     * value is given, then the same percentage is used for all
+     * levels.
      *
      * \sa itk::ImageRegistrationMethodv4::SetMetricSamplingPercentage
      * @{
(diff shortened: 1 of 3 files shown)


Code/Registration/include/sitkImageRegistrationMethod.h
```diff
@@ -451,19 +451,30 @@ namespace simple
     SITK_RETURN_SELF_TYPE_HEADER SetMetricMovingMask( const Image &binaryMask );
 
     /** \brief Set percentage of pixels sampled for metric evaluation.
+     *
+     * The percentage is of the number of pixels at the virtual domain
```
zivy: replace at with in - "number of pixels in the virtual domain"

Code/Registration/include/sitkImageRegistrationMethod.h
```diff
@@ -451,19 +451,30 @@ namespace simple
...
      * The seed parameter is used to seed the pseudo-random number
      * generator used in generating the sampling set of points. If the
      * seed parameter is 0, then the wall clock is used, otherwise the
      * fixed seed is used for reproducible behavior.
      *
+     * The number of levels must either be one, or match the number of
+     * shrink factors and smoothing sigmas. If the number of
+     * percentages is one, then the same percentage is used for all
+     * levels.
+     *
```
zivy: Not sure about the documentation above, suggested modification:

When providing multiple values the number of entries must match the number of shrink factors and smoothing sigmas. If a single value is given, then the same percentage is used for all levels.

Code/Registration/include/sitkImageRegistrationMethod.h
```diff
@@ -451,19 +451,30 @@ namespace simple
...
      * \sa itk::ImageRegistrationMethodv4::SetMetricSamplingPercentage
      * @{
      */
     SITK_RETURN_SELF_TYPE_HEADER SetMetricSamplingPercentage(double percentage, unsigned int seed = sitkWallClock);
     SITK_RETURN_SELF_TYPE_HEADER SetMetricSamplingPercentagePerLevel(const std::vector<double> &percentage, unsigned int seed = sitkWallClock);
     /** @} */
 
+    /** \brief Get the percentage of pixels used for metric evaluation.  */
+    const std::vector<double> &SetMetricSamplingPercentagePerLevel() const;
+
```
zivy: Function signature says SetMetric... , should be GetMetric...

blowekamp:
topic merged to next

PR #598, opened 2018-11-19
Adding image viewing example.

zivy:
Adding example illustrating the use of the Show function and the new
ImageViewer class.


blowekamp:
Topic merged into next

PR #601, opened 2018-11-21
ENH: Adding new example illustrating new SimpleITK IO.

zivy:
Illustrating how to only read the meta-data information, set a
specific IO class from those supported and how to read sub-regions of
an image.

### DIFF
--- a/Examples/AdvancedImageReading/Documentation.rst
+++ b/Examples/AdvancedImageReading/Documentation.rst
@@ -1,0 +1,60 @@
+.. _lbl_advanced_image_reading:
+
+Advanced Image Reading
+==============================
+
+Overview
+--------
+
+This example illustrates advanced image reading options:
+  1. Explicitly selecting a specific IO for reading.
+  2. Querying an image for its meta-data without reading the bulk intensity data.
+  3. Reading a sub-region of an image.
+
+Selecting a Specific IO
+++++++++++++++++++++++++
+When using the default settings for the `ImageFileReader class <https://itk.org/SimpleITKDoxygen/html/classitk_1_1simple_1_1ImageFileReader.html>`_ or
+the `ReadImage <https://itk.org/SimpleITKDoxygen/html/namespaceitk_1_1simple.html#ae3b678b5b043c5a8c93aa616d5ee574c>`_ function
+you have minimal control over the reading. That is, the specific IO mechanism is
+determined automatically based on the file type and the whole file is read into
+memory.
+
+In some cases there are multiple IO classes that support reading the same image
+format and you do not know which one was used. For example the 'SCIFIOImageIO'
+and 'JPEGImageIO' both support reading JPEG images. The SCIFIOImageIO reading
+is much slower than the JPEGImageIO. With the advanced reading approach you can
+find out which IO classes are supported by the ImageFileReader and indicate which
+one you want to use.
+
+Querying an image for its meta-data
+++++++++++++++++++++++++++++++++++++
+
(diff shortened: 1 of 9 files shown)


Examples/AdvancedImageReading/Documentation.rst
```diff
@@ -0,0 +1,60 @@
...
+--------
+
+This example illustrates advanced image reading options:
+  1. Explicitly selecting a specific IO for reading.
+  2. Querying an image for its meta-data without reading the bulk intensity data.
+  3. Reading a sub-region of an image.
+
+Selecting a Specific IO
+++++++++++++++++++++++++
+When using the default settings for the `ImageFileReader class <https://itk.org/SimpleITKDoxygen/html/classitk_1_1simple_1_1ImageFileReader.html>`_ or
```
blowekamp: The seletecting an imageIO should be a separate example from the read meta-data and sub-region example.

Examples/AdvancedImageReading/Documentation.rst
```diff
@@ -0,0 +1,60 @@
...
+the `ReadImage <https://itk.org/SimpleITKDoxygen/html/namespaceitk_1_1simple.html#ae3b678b5b043c5a8c93aa616d5ee574c>`_ function
+you have minimal control over the reading. That is, the specific IO mechanism is
+determined automatically based on the file type and the whole file is read into
+memory.
+
+In some cases there are multiple IO classes that support reading the same image
+format and you do not know which one was used. For example the 'SCIFIOImageIO'
+and 'JPEGImageIO' both support reading JPEG images. The SCIFIOImageIO reading
```
blowekamp: The setting ImageIO example should be setup with prose indicating that SimpleITK can be configure with optional ITK ImageIO modules such as DCMTK and SCIFIO. Both provide support of image file types that are already supported by other ImageIOs.

Examples/AdvancedImageReading/Documentation.rst
```diff
@@ -0,0 +1,60 @@
...
+is much slower than the JPEGImageIO. With the advanced reading approach you can
```
blowekamp: Please be positive of the toolkits features.

Examples/AdvancedImageReading/Documentation.rst
```diff
@@ -0,0 +1,60 @@
...
+find out which IO classes are supported by the ImageFileReader and indicate which
+one you want to use.
+
+Querying an image for its meta-data
+++++++++++++++++++++++++++++++++++++
+
+In some cases you may only want to read an image's meta-data without reading the
+bulk intensity data into memory. For instance, if you want to display image
```
blowekamp: For this particular case we need to know the image's size before we do sub-region IO.

Please reference the DICOM meta-dictionary example which already does this... stay focused on the region IO.

Examples/AdvancedImageReading/Documentation.rst
```diff
@@ -0,0 +1,60 @@
...
+information (e.g. patient name) to a user in a GUI which then allows them to
+select a specific image.
+
+Reading an image sub-region
++++++++++++++++++++++++++++
+In some cases you are only interested in reading a sub-region of an image. This
+may be due to memory constraints or if it is known that the relevant content is
+always in a sub region of the image.
```
blowekamp: We should note that not every ImageIO of image file may support streaming, perhaps include a list of ImageIOs which do currently support stream reading.

blowekamp:
This is a great addition. I'm merging into next, then I'll add a couple tweaks myself.

Issue #602, opened 2018-11-23
Host without imagej cannot redefine viewers via SITK_SHOW_COMMAND


richardbeare:
The SITK_SHOW_COMMAND allows redefinition of the viewer used by _Show_. However, _Show_ searches for imagej before doing anything else  https://github.com/SimpleITK/SimpleITK/blob/a4eba04b0380c3f76386d635a4c7defe92b2bb61/Code/IO/src/sitkShow.cxx#L616

_FindImageJ_ issues an exception if it can't find imagej:  https://github.com/SimpleITK/SimpleITK/blob/a4eba04b0380c3f76386d635a4c7defe92b2bb61/Code/IO/src/sitkShow.cxx#L495

Thus, if there's no imagej on a host, supplying an environment variable doesn't help because the exception bypasses the environment variable code.

```r
library(SimpleITK)
m <- ReadImage("/usr/share/fsl/data/standard/MNI152_T1_1mm_brain.nii.gz")
Show(m)
Error in f(...) : Exception in SITK - check warnings
In addition: Warning message:
In f(...) :
  Exception thrown in SimpleITK Show: /home/richardb/Projects/ITK4/SimpleITK/Code/IO/src/sitkShow.cxx:495:
sitk::ERROR: No appropriate executable found.

```

```bash
mkdir /tmp/bin
# pretend imagej is installed
touch /tmp/bin/imagej
export PATH=${PATH}:/tmp/bin
export SITK_SHOW_COMMAND=$(which itksnap)
```
```r
library(SimpleITK)
m <- ReadImage("/usr/share/fsl/data/standard/MNI152_T1_1mm_brain.nii.gz")
Show(m)
# Success
```

dave3d:
I've submitted a patch that should fix the problem.

https://github.com/SimpleITK/SimpleITK/pull/604

blowekamp:
@Dave You can reference an issue in a pull request and indicates that it "fixed issues ###" and GitHub will automatically link and close the issue when merged. Here is the docs:
https://help.github.com/articles/closing-issues-using-keywords/

rojinsafavi:
I am having the same issue

I ran following code from your examples:

`python ImageRegistrationMethod1.py ~/Downloads/chrome_dl/k/C2-Composite0.tif ~/Downloads/chrome_dl/k/C2-Composite1.tif trial.hdf5`

and I got:

[...]

PR #604, opened 2018-11-28
Fix for Show w/o ImageJ or Fiji

dave3d:
Previously Show would look for ImageJ or Fiji first, before
checking the SITK_SHOW_COMMAND env var.  That could result in
an exception.

Now it checks SITK_SHOW_COMMAND first.  Thus if we don't need
to look for ImageJ or Fiji, we don't.


dave3d:
Fixes #602

blowekamp:
The changes look good, thanks! merging into next!

PR #607, opened 2018-12-04
Add multi-resolution BSpline registration support

blowekamp:


Code/Registration/include/sitkImageRegistrationMethod.h
```diff
@@ -144,6 +145,20 @@ namespace simple
     { return this->m_InitialTransformInPlace;}
     /** @} */
 
+    /** \brief Set an initial BSpline transform to optimize.
+      *
```
zivy: replace method with version? : "A specialized version of SetInitialTransform..."

Does the number of scaleFactors need to match the shrink factors? If for 3 levels I just give [1,2] will the third level be twice the number of control points as the first (retains the number from the second pyramid level)?

blowekamp: With the current implementation if there is now scalingFactor then no adaptor will be created, and the previous transform will be used unmodified.

zivy: Got it. This should be added to the documentation.

blowekamp:
I think this patch is ready to go. I'll add an example in another PR.

PR #608, opened 2018-12-04
Improve test coverage for ImageViewer class.

dave3d:
The ImageViewer test now invokes the class's Execute method.
The greatly improves the coverage for the class's code.

Prior to calling Execute the display command is set to use
'cmake -E md5sum' which computes the md5sum on input image.
This avoids trying to pop up an ImageJ or Fiji viewer.


blowekamp:
Looks good. Merging into next.

blowekamp:
There appears to be valgrind defect related to this patch:
https://open.cdash.org/viewDynamicAnalysisFile.php?id=5990621

PR #609, opened 2018-12-05
Update the ITK superbuild version along release-4.13

blowekamp:
This update to ITK include a critical fix to the LBFGS2 optimizer.


blowekamp:
topic merged into next

PR #615, opened 2018-12-12
In ITK Superbuild ignore CMake developer warnings

blowekamp:


blowekamp:
This does not suppress the CMake warnings.

PR #618, opened 2018-12-12
Fixed memory leak

dave3d:
When we've finished executing the display process, we weren't always
deleting the KWSys process object.


Code/IO/src/sitkImageViewer.cxx
```diff
@@ -530,14 +530,16 @@ void ExecuteCommand( const std::vector<std::string> & cmdLine, const unsigned in
       // This first case is what we expect if everything went OK.
 
       // We want the other process to continue going
-      itksysProcess_Delete( kp ); // implicitly disowns
+      //itksysProcess_Delete( kp ); // implicitly disowns
```
blowekamp: This line should be removed.

PR #622, opened 2018-12-18
Removed ImageJ stuff from Installation doc

dave3d:
Also tweaked some text.


blowekamp:
Great! Thank you, merging into next

PR #623, opened 2018-12-18
Adding option to procedure ReadImage to specify the ImageIO

blowekamp:
By default the "" string for the imageIO parameter allows the IO to
automatically choose the ImageIO, if it is specified than that one
will be tempted to be constructed and used for reading.


zivy (approved):
Very Nice! Code looks good. I like it, enables finer control over reading with minimal coding (OO approach is more verbose). Also, now the caveats in the documentation are gone.

PR #629, opened 2018-12-20
Update filter script and csv

dave3d:


blowekamp:
it looks like you did not have the ITK review module on when you build ITK.

blowekamp:
Nice update to the python script for Python 3. Merging into next and master.

PR #641, opened 2019-01-03
Add script to upload binary data to data.kitware.com

blowekamp:


zivy (commented):
Not sure what the script is doing with respect to md5 hashes, towards the end of the script. I thought the use of sha512 replaced the usage of md5 hashes. What am I missing here?

zivy (approved):

blowekamp:
Currently, we are maintaining both md5 and sha512 for the data files. I think just because of the SourceTarball.sh script and legacy Midas server. The check at the end of this script recommends removing an existing/conflicting md5 file which may not be consistent with keeping both. We should work towards just using sha512 though.

PR #649, opened 2019-01-11
Fix R documentation generation - corrected escaping.

richardbeare:
Percent signs were not being escaped. Rd is a latex
flavour and percent signs are comments.

Quotes were being escaped and shouldn't be.

This patch escapes percent signs are removes escapes from quotes.

I'll probably need to rebase this - let me know.


blowekamp:
Topic merged into next.

PR #650, opened 2019-01-15
Improve the N4 bias field correction example

blowekamp:


zivy (approved):
LGTM. Highlights the two things users need to be aware of (input image type required to have real pixel type and uint8 mask).

PR #662, opened 2019-01-31
Add user Applications to show search path

blowekamp:
When a user on Mac OSX does not have admin privileges, the Fiji.app
should be placed in ~/Applications. Search for this too.


dave3d (approved):
Looks good to me.

PR #672, opened 2019-02-08
Use VisualStudio CMake generator

blowekamp:


azure-pipelines.yml
```diff
@@ -105,16 +105,16 @@ jobs:
         workingDirectory: $(Agent.BuildDirectory)
       - script: python -m pip install ninja
         displayName: 'Install ninja dependency'
```
blowekamp: Ninja does not need to be installed anymore with a VS generator

blowekamp:
merged in ddce8a8feaccb3ffca798cb26eb123a22942aece

PR #681, opened 2019-02-25
Fix for hang in Show function on Windows

dave3d:
It was re-deleting a previously deleted process.


blowekamp:
This is great you tracked it down!

So the Delete method would get called twice?

Was there were we talked about the [resource acquisition is initialization](https://en.wikipedia.org/wiki/Resource_acquisition_is_initialization) (RAII) pattern? Here is a small example:
https://github.com/SimpleITK/SimpleITK/blob/master/Wrapping/Python/sitkPyCommand.cxx#L31-L46

dave3d:
> This is great you tracked it down!
> 
> So the Delete method would get called twice?
> 
> Was there were we talked about the [resource acquisition is initialization](https://en.wikipedia.org/wiki/Resource_acquisition_is_initialization) (RAII) pattern? Here is a small example:
> https://github.com/SimpleITK/SimpleITK/blob/master/Wrapping/Python/sitkPyCommand.cxx#L31-L46

Did we talk about that?  I don't recall, but I am getting old.

Interestingly, the bug didn't seem to affect OS X or Linux, even though it happened for those Ones too.

blowekamp:
The CI seemed to work well. Is this patch based on the release branch?

dave3d:
It's off master.  Should I have done release?

blowekamp:
Please rebase it onto release so that it can more easily be used to patch the release branch.

PR #687, opened 2019-03-03
Moved SPIE course from upcoming to past section.

zivy:


blowekamp:
Looks like a good update. But I want to figure out why the azure is failing...

PR #688, opened 2019-03-04
Test new VS2019 image

blowekamp:
[ci skip]


blowekamp:
See #695 for implementation

PR #693, opened 2019-03-06
Fix image reader direction

### LINKED ISSUE #691: GetDirection() behaves differently when called from ImageFileReader than from Image
When I use an ImageFileReader, call the ReadImageInformation() method then call GetDirection(), the tuple that is output is column-major whereas when I call GetDirection() from the Image itself (after reader.Execute()), the tuple is row-major.
So I have to transpose my 3x3 Direction matrix when I call direction from reader rather than the Image itself.
I don't need to load the images, I just need to get their information.
Is this normal behavior?

blowekamp:
Closes #691

### DIFF
--- a/Code/IO/include/sitkImageFileReader.h
+++ b/Code/IO/include/sitkImageFileReader.h
@@ -40,7 +40,10 @@ namespace itk {
      *
      * An interface is also provided to access the information from
      * the underlying itk::ImageIO. This information can be loaded
-     * with the ReadImageInformation method.
+     * with the ReadImageInformation method. The information is from
+     * the itk::ImageIO interface. In some degenerate cases reading
+     * the bulk data may produce different results. Please see
+     * itk::ImageFileReader for more details.
      *
      * Reading takes place by the ITK ImageIO factory mechanism. ITK
      * contains many ImageIO classes which are responsible for reading
--- a/Testing/Unit/sitkImageIOTests.cxx
+++ b/Testing/Unit/sitkImageIOTests.cxx
@@ -563,23 +563,32 @@ TEST(IO, ImageFileReader_ImageInformation )
   EXPECT_ANY_THROW( reader.GetMetaData("nothing") );
 
 
+  sitk::Image image = sitk::ReadImage(dicomFile1);
   reader.ReadImageInformation();
 
   EXPECT_EQ(reader.GetPixelID(), sitk::sitkInt16);
+  EXPECT_EQ(reader.GetPixelID(), image.GetPixelID());
   EXPECT_EQ(reader.GetPixelIDValue(), sitk::sitkInt16);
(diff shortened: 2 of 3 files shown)


Code/IO/include/sitkImageFileReader.h
```diff
@@ -40,7 +40,10 @@ namespace itk {
      *
      * An interface is also provided to access the information from
      * the underlying itk::ImageIO. This information can be loaded
-     * with the ReadImageInformation method.
```
zivy: "The information is from the from" -> "The information is from the itk::ImageIO..."

"In some degenerate cases the reading the bulk data may produce difference results." -> 
"In some degenerate cases reading the bulk data may produce different results.
"

Testing/Unit/sitkImageIOTests.cxx
```diff
@@ -563,23 +563,32 @@ TEST(IO, ImageFileReader_ImageInformation )
...
   EXPECT_EQ(reader.GetNumberOfComponents(), 1u);
+  EXPECT_EQ(reader.GetNumberOfComponents(), image.GetNumberOfComponentsPerPixel());
   EXPECT_VECTOR_DOUBLE_NEAR(reader.GetOrigin(), v3(-112, -21.687999, 126.894000 ), 1e-6);
+  EXPECT_VECTOR_DOUBLE_NEAR(reader.GetOrigin(), image.GetOrigin(), 1e-6);
   EXPECT_VECTOR_DOUBLE_NEAR(reader.GetSpacing(), v3(0.859375, 0.8593899, 1.600000 ), 1e-6);
-  EXPECT_VECTOR_DOUBLE_NEAR(reader.GetDirection(), v9(1, 0, 0,
-                                                      0, 0.46665081166793676, -0.88444164305490258,
-                                                      -0, 0.88444164305490258, 0.46665081166793676 ), 1e-8);
+  EXPECT_VECTOR_DOUBLE_NEAR(reader.GetSpacing(), image.GetSpacing(), 1e-6);
+  EXPECT_VECTOR_DOUBLE_NEAR(reader.GetDirection(), v9(1, 0, -0,
```
zivy: Out of curiosity, does the sign of the zero (-0 vs. 0) matter here? I don't expect it to. Is this related to how EXPECT_VECTOR_DOUBLE_NEAR is implemented?

zivy (commented):
Just some comment corrections and a minor question. The actual bug fix looks good to me.

Testing/Unit/sitkImageIOTests.cxx
blowekamp: Technically IEEE 754 floating point representation has two zeros for the sign. This baseline value is now how it was printed. It makes now numeric difference, but visually indicate the proper matrix orientation.

blowekamp:
Thanks for the review Ziv. I have address the issues. merging...

PR #698, opened 2019-03-11
WIP: Test VS2019 with CMake 3.14.0-rc4

blowekamp:


blowekamp:
The current VS2019 image and cmake 3.14.0 are not compatible.

PR #701, opened 2019-03-15
Fix Rpackaging failure to copy packaging directory

blowekamp:
The default build directory for R packages include "R.INSTALL", which
matched the regular expression "*.in", which case insensitive file
systems. Add end of input marked to ensure file extension is at the
end of the string. Add sha512 exclusion too.


blowekamp:
Addressed SimpleITK/SimpleITKRInstaller#32.

richardbeare (commented):
Can confirm a successful build on my mac laptop. However I was never able to reproduce the issue locally, so I guess by test isn't helpful.

blowekamp:
Great that it still works for you! I wonder if this problematic path is a new feature of a newer R or devtools.

PR #709, opened 2019-03-21
Azure ctest build line must be the last statement

blowekamp:
The return value of the last statement executed in a script block
determines if the step passes or fails. So the ctest script must be run
as the last statement so that the build results will be the results of
the script step.


bradking:
Do: check

blowekamp:
No longer relevant

PR #716, opened 2019-03-28
Added example Connected Threshold CSharp

pettaa123:


blowekamp:
Great! Thank you for contributing the example. Are you using git to commit or are you just uploading via the web page?

It would be wonderful of you could follow some of the other example which have a directory with files for the different language and a Documentation.rst file. Here is an example:
https://github.com/SimpleITK/SimpleITK/tree/master/Examples/ImageRegistrationMethod1

pettaa123:
I started [here](https://github.com/pettaa123/SimpleITK/tree/ThresholdSegExampleDirectory/Examples/ImageSegmentationMethodConnectedThreshold). Is that what you meant? Dont know how to handle the CMake file. Also I dont know what to do with segmentationExamplePrefix.rst

blowekamp:
Yes, that looks about right. I'll be adding CSharp CI for testing this afternoon. So we'll be able to test the PR with CSharp. You should also be building SimpleITK ITK and and uses CTest to run the tests/ examples too.

blowekamp:
Thank you very much for the PR. Can you please rebase onto the current master so that the CSharp examples will be tested by the CI?

pettaa123:
Correct?

blowekamp:
It looks like you merged origin master instead of re-basing. Thank OK.

You have not added this example to the CMake file with a `sitk_add_csharp_test`. Please do that so we can compile, and test the example.

pettaa123:
Thank you so much for your patience. Should we add the output of ITK's ConnectedThreshold Testing to test against with?

blowekamp:
No, Thank you for your effort to contribute code!

- Please remove the old file name.
- You can use an MD5 hash for image comparison with the [IMAGE_MD5_COMPARE](https://github.com/SimpleITK/SimpleITK/blob/master/CMake/sitkAddTest.cmake#L18) option. Just run it with a bad has the first time, then the check the output to set the excepted results.

pettaa123:
When I take the hast from [here](https://github.com/SimpleITK/SimpleITK/blob/master/Examples/test/CMakeLists.txt#L43) the test passes. I am wondering if there is a test log, which saves the values of the comparison? Or how do I get the correct hash?

blowekamp:
If you run the test from the command line with the ctest flag `-V` all the output will be displayed. The command should be something like:

```
$ crest -C Release -V -R  CSharp.Example.ConnectedThresholdSegmentation
```

When the CI is run, the output of the test or build warnings can be seen on [CDash](https://open.cdash.org/index.php?project=SimpleITK). Here is the output of your example successfully running: https://open.cdash.org/testDetails.php?test=753232271&build=5858421

Looks likely you contribution is ready to merge! 🎆

PR #738, opened 2019-04-27
added files

pettaa123:
I dont know how to specify or correctly add the second input image


blowekamp:
Awesome! 

ITK has the processed documented very well:
https://github.com/InsightSoftwareConsortium/ITK/blob/master/Documentation/UploadBinaryData.md

I think the easies way to upload the data to Girder is to use the Web interface.

For now, manually run a 'sha512' executable to get the sha512 hash or get it from Girder. The create the file with the name.

ITK's process is documented here, but you'll need to do in manually for SimpleITK: https://github.com/InsightSoftwareConsortium/ITK/blob/master/Documentation/Data.md

pettaa123:
thank you, I could follow the guide and add the correct hashes, but I dont get where this newline error comes from?

Testing/Data/Input/ThresholdSegmentationLevelSetVentricle.png.sha512
```diff
@@ -0,0 +1 @@
+4a75c62c8b9778683f39eafcd31fc4274fcb7c9096d4cbbbb1f798da5a37becc5657a24d24092f887c1919dfade1aa9964399cf982ace4a3b79865c501b8d746
```
blowekamp: If you open this file in a text editor, you should be able to hit enter and the end of the line add a new line there.

PR #741, opened 2019-05-02
Update git setup for Github as official upstream

blowekamp:


dave3d (approved):
Looks awesome!

Utilities/SetupForDevelopment.sh
```diff
@@ -80,29 +80,37 @@ echo -e "Git version $git_version is OK.\n"
 
 cd "${BASH_SOURCE%/*}/.." &&
```
zivy: For some reason cd "${BASH_SOURCE%/*}/.." didn't work on my mac (using bash). When I ran the script from inside the Utilities directory I get the error: "./SetupForDevelopment.sh: line 82: Utilities/GitSetup/setup-user: No such file or directory"

It does work when running from the directory above Utilities.

zivy (commented):
Had some issues with the setup. They do not appear to be critical as the origin and upstream repos were set correctly and when I do git config --list --show-origin and see what was set in the SimpleITK repo things look reasonable.

Utilities/SetupForDevelopment.sh
blowekamp: This second `cd` is not needed, and is broken with relative paths. Removed in next patch.

PR #743, opened 2019-05-03
added files for CSharp Canny Seg Example

pettaa123:


blowekamp:
Great! While there CI appears green there is a test failure:
https://open.cdash.org/testDetails.php?test=760896475&build=5892030
```
Unhandled Exception: System.ArgumentOutOfRangeException: Exception of type 'System.ArgumentOutOfRangeException' was thrown.
Parameter name: index
   at itk.simple.VectorUInt32.setitem(Int32 index, UInt32 val)
   at itk.simple.examples.FastMarchingSegmentation.Main(String[] args)
sitkCompareDriver: Process exited with return value: -532459699
```
This looks like the same error addressed here: #745

blowekamp:
So I improved the CI a little and rebased the on to master. This patch now looks ready.

PR #751, opened 2019-05-21
COMP: Update statistics filter without indexed inputs

blowekamp:
Updates to ITKv5 removed the generic image indexed inputs
methods. Updated the affected filters to use the names input
interface.


zivy (approved):
LGTM, though I'm sure someone will complain that we broke their code with this minor interface change. I still think this change makes sense.

blowekamp:
Not method names are changed with this, just argument names

PR #754, opened 2019-05-28
COMP: Update statistics filter without indexed inputs

blowekamp:
Updates to ITKv5 removed the generic image indexed inputs
methods. Updated the affected filters to use the names input
interface.


dave3d (approved):
Looks good to me.

PR #757, opened 2019-05-31
Allow for tolerance comparison of BSpline domain origin

blowekamp:
In ITKv5 with the commit 794c82d16ebc4b4035f37e0675176e7677f9bd68, the
consistency of the parameters has been fix but there is a little
variability with some of the transform domain values derived from the
parameters. This patch adds tolerance to certain test of derived
transform domain values as needed.


dave3d (approved):
Cool beans.

PR #764, opened 2019-06-04
Use named inputs for Chan and Vese level set filter

blowekamp:
Adding named inputs clarifies usage of the filter and what the
expected inputs are. Note the
itk::ScalarChanAndVeseDenseLevelSetImageFilter::SetFeatureImage does
not currently accept a pointer to a constant image only a constant
pointer.


zivy (approved):
LGTM.

PR #766, opened 2019-06-05
added setBuffer to getBuffer example

pettaa123:


blowekamp:
Very nice, thank you for the contribution.

Please remove the old `ImageGetBuffer.cs` file that this example replaces.

Also, looks like we have a configuration error here:
https://open.cdash.org/viewBuildError.php?buildid=5945378

blowekamp:
This looks like a great contribution. Are you able to fix the build errors?

Examples/CSharp/ImageGetSetBuffer.cs
```diff
@@ -0,0 +1,116 @@
...
+                return;
+            }
+
+            string inputFilename = args[0];
+            string outputFilename = args[1];
+
+            // Read input image
+            SitkImage input = SimpleITK.ReadImage(inputFilename);
+
+            // Cast to we know the the pixel type
```
zivy, line 46: "Cast to..." -> "Cast so..."

zivy (approved):
Probably should have been a "git mv" and then edited instead of "git rm" and creating a new file, but not critical.

PR #767, opened 2019-06-05
Add support for slicing operations on 4D images

### LINKED ISSUE #765: Pythonic slicing of 4D images are not supported.
It is possible to slice an image like so:
`img3D[:,:,5]` or `img3D[0:8,2:5,0:1]` which is VERY convenient! However, if i use JoinSeries to create a 4D image, the slicing does not work anymore:
```
img4D = sitk.JoinSeries(img3D,img3D)
img4D[:,:,:,0]

IndexError: invalid index
```
This would be great to have in 4D as well!

blowekamp:
Instatiate the SliceImageFilter in 4D. The Extract image filter is
alread instatiated in 4D.

Added support for detection of extract index to 4+ dimensions.

NOTE: Supported extract based indexing for 4D->2D is currently
implemented and the error is generated is rather generic.
closes #765

### DIFF
--- a/Code/BasicFilters/json/SliceImageFilter.json
+++ b/Code/BasicFilters/json/SliceImageFilter.json
@@ -4,6 +4,8 @@
   "template_test_filename" : "ImageFilter",
   "number_of_inputs" : 1,
   "pixel_types" : "NonLabelPixelIDTypeList",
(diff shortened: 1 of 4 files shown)


dave3d (approved):
Looks good.

PR #776, opened 2019-07-05
ENH: adding VS 2015 build

blowekamp:


blowekamp:
The AZP vs2015-win2012r2 image is old and not useful. The v140 toolset could be used, but that would be different PR.

PR #779, opened 2019-07-09
Need to update before installing apt-get package

blowekamp:


zivy (approved):
LGTM (not sure the forced yes is required).

PR #781, opened 2019-07-10
Need to update before installing apt-get package

blowekamp:


zivy (approved):
LGTM. In the notebooks repo CircleCI config I have `sudo apt-get update` without the forced yes to prompts. Is the forced yes needed?

PR #787, opened 2019-07-12
ENH: adding ProxTVImageFilter

dzenanz:
Is the all that is required to add a filter to SimpleITK's wrapping?


blowekamp:
Adding a json file is all that is usually needed for filter which file the standard ITK conventions. I usually end up spending a bit of time on the test case and some time find bugs.

blowekamp:
Thanks for the PR. To test it with the CI, you will need to temporarily enable this remote module in the [Superbuild](https://github.com/SimpleITK/SimpleITK/blob/master/SuperBuild/External_ITK.cmake#L98)

blowekamp:
@dzenanz Currently SimpleITK is compiling against ITKv4. The TotalVariation ITK Module is a recent addition only available in ITKv5. So this filter will require ITKv5 enabled to use and test, this is currently only done on [one build](https://github.com/SimpleITK/SimpleITK/blob/master/Testing/CI/Azure/azure-pipelines-batch.yml#L107), when topic is merged into master. To get the CI testing this filter you can hack it.

Also, the documentation in the json can be added from the ITK Doxygen with [this script](https://github.com/SimpleITK/SimpleITK/blob/master/Utilities/GenerateDocs/JSONDocUpdate.sh). Looking at the [filter](https://github.com/InsightSoftwareConsortium/ITKTotalVariation/blob/master/include/itkProxTVImageFilter.h), this class needs some documentation work.

dzenanz:
I created this PR in part to get feedback on content of that JSON file, because I am using it in https://github.com/BlueQuartzSoftware/ITKImageProcessing/pull/182. It is up to you whether you want to merge it, close it, keep it open until SimpleITK switches to ITK 5, or take over this PR.

blowekamp:
Thanks a gain for the contribution. I'll take of this PR to add the filter, so it'll be available when ITKv5 is enabled.

blowekamp:
This ITK remote module requires extra dependencies that are not working out of the box.

PR #789, opened 2019-07-17
Add URL_HASH to external projects

blowekamp:
Update the function used to download source code via CMake's
ExternalData to also produce a hash variable, then used it in the
external project calls. This addresses warning produce by CMake that
the URL_HASH is missing.


dave3d (approved):
Looks good.

PR #799, opened 2019-08-05
Azure build script improvements

blowekamp:


Testing/CI/Azure/azure.cmake
```diff
@@ -3,52 +3,68 @@
 # ------------
 #
 # Sets a CMake variable from an environment variable. If the
-# environment variable  is not  defined then either a fatal error is
-# generated or the CMake variable is not modified.
```
zivy: exist -> existing
use -> used

"environment variable is not defined then the existing CMake value will
persist. Next if a "DEFAULT" argument is provided then the "value"
parameter will be used to set the variable."

Issue #800, opened 2019-08-06
Feature request - BinaryPruningImageFilter


szhorvat:
This is a request to add `BinaryPruningImageFilter` to SimpleITK. Even if it is a 2D-only filter, this is extremely useful functionality. Adding it should not wait for a working 3D implementation.

blowekamp:
I will look in to wrapping this filter again in the next week or so. If I recall correctly, there were other issues with this filter. Would you be willing to test out the PR and and see if the filter will work for your use case?

szhorvat:
> Would you be willing to test out the PR and and see if the filter will work for your use case?

Yes, absolutely.

szhorvat:
I think I would be able to build SimpleITK from sources and test the function from C++ when it's ready. I have something to hook up SimpleITK to Mathematica—for personal use—using other tools than SWIG. This is how I would be using it. Mathematica has a pruning filter but it's not good enough for my use case (it is extremely slow).

blowekamp:
Closed by #803

PR #803, opened 2019-08-15
ENH: Wrapping BinaryPrunningImageFilter

blowekamp:
addresses #800


blowekamp:
@szhorvat This adds wrapping for the BinaryPruning filter. Can you please try the filter. I was not able to get meaning fully results for the couple images I tried. It would be good to come up with a little test ( preferable with the exiting data ) to test the filter. Thanks!

szhorvat:
@blowekamp I'll try, but it'll take me a little while to compile SimpleITK and get everything working. So I'm posting a simple example image and the expected output after running a pruning filter.

prunable

pruned

blowekamp:
Thanks for the images. The filter seemed to work very well on them. The input just needs to be 1's not 255.

blowekamp:
A test or two with the provided image is needed. Also it need to be determined if this filter works in 3D or if it should only be instantiated in 2D.

szhorvat:
I have not forgotten, it's just been an incredibly busy time for me. Please give me a few more days.

blowekamp:
Updated the testing with the sample image provided. Now we just need to consider if 3D should be supported. Since the filter only works on 0's' and 1's we should consider only supporting uint8_t types to save on compilation time and binary size to keep the SimpleITK package efficient.

zivy (commented):
I agree that the filter should only support uint8, to keep the library small. Also, this isn't much of a burden on users as it is trivial to move from other label image types to binary.

Is the UploadBinaryData.sh supposed to be part of this commit or is this an error?

blowekamp:
> Is the UploadBinaryData.sh supposed to be part of this commit or is this an error?

Nope, thanks for spotting it. It was needed from PR #641.

zivy (approved):
Minor spelling mistake in two places (binnary instead of binary).

PR #806, opened 2019-08-16
Add FAQ entry for missing windows dll dependency

blowekamp:
SimpleITK windows dll library is also dependent on the C++ library. For
VS2015 that is MSVCP140.dll, which is not packages with some Python
distributions.

### DIFF
--- a/Documentation/docs/source/faq.rst
+++ b/Documentation/docs/source/faq.rst
@@ -30,6 +30,20 @@ required for your project, then create a new environment:
 
   conda env create -f environment.yml
 
+Why am I getting "ModuleNotFoundError: No module named '_SimpleITK'" when importing SimpleITK on Windows?
+---------------------------------------------------------------------------------------------------------
+
+        Traceback (most recent call last):
+          File "C:\Python37\lib\site-packages\SimpleITK\SimpleITK.py", line 14, in swig_import_helper
+            return importlib.import_module(mname)
+          ...
+          File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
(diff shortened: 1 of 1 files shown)


Documentation/docs/source/faq.rst
```diff
@@ -30,6 +30,20 @@ required for your project, then create a new environment:
...
+        Traceback (most recent call last):
+          File "C:\Python37\lib\site-packages\SimpleITK\SimpleITK.py", line 14, in swig_import_helper
+            return importlib.import_module(mname)
+          ...
+          File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
+        ImportError: DLL load failed: The specified module could not be found.
+
+The above error occurs on Microsoft windows when system libraries SimpleITK depends on are not available on the system.
+This issue can be resolved by installing the appropriate
+`Redistributable for Visual Studio <https://www.microsoft.com/en-us/download/details.aspx?id=48145>`__.
```
zivy: I know the component name is "Redistributable for Visual Studio" but that is a bit strange, possibly we use "Redistributable Packages for Visual Studio" here?

zivy (approved):
Minor naming comment that is possibly ignored.

Documentation/docs/source/faq.rst
blowekamp: We could say  [Redistributable for Visual Studio](https://www.microsoft.com/en-us/download/details.aspx?id=48145) package. If you really thing the term package is needed.

zivy: That sounds more readable to me, I'm thinking of users that don't know what it is that we are telling them to install and just want to follow instructions.

PR #816, opened 2019-08-22
Use always available CMake executable for zipping archive

### LINKED ISSUE #647: CSharp packaging in separate build
The CSharp packaging is dependent on java jar to create the package. This needs to be exposed to better work when SimpleITK-CSharp and produce a warning of failure when not there.

blowekamp:
Available since CMake 3.3 the "format" option for the tar command
allows for generation of zip archive.

closes  #647


zivy (approved):
LGTM.

PR #818, opened 2019-08-23
Use ITK's enum typedef to qualify enum value

blowekamp:
The enables compatibility with ITKv5 C++11 call based enums.

### DIFF
--- a/Code/BasicFilters/src/sitkLandmarkBasedTransformInitializerFilter.cxx
+++ b/Code/BasicFilters/src/sitkLandmarkBasedTransformInitializerFilter.cxx
@@ -181,8 +181,9 @@ Transform LandmarkBasedTransformInitializerFilter::ExecuteInternal ( const Trans
 
   filter->SetLandmarkWeight ( this->m_LandmarkWeight );
 
+  typedef typename FilterType::TransformType::TransformCategoryType TransformCategoryType;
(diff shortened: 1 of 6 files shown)


dave3d (approved):
Coolness

hjmjohnson (commented):
ITKv5 instead of ITKv4 in the commit message.

blowekamp:
There are compilation errors on Linux along the lines:
```
/home/blowekamp/src/SimpleITK/Code/Common/src/sitkPimpleTransform.hxx: In member function ‘virtual bool itk::simple::PimpleTransformBase::IsLinear() const’:
/home/blowekamp/src/SimpleITK/Code/Common/src/sitkPimpleTransform.hxx:151:82: error: ‘itk::TransformBase<double>::TransformCategoryType’ is not a class or namespace
       return (this->GetTransformBase()->GetTransformCategory() == TransformBase::TransformCategoryType::Linear);
                                                     
```

This is occurring with gcc 5 only, as other versions seem OK. The error seems to occur with all the changes which add the qualifying typedef for the enum value. @hjmjohnson Have you encountered this? Do you know of a work around or an alternative to access the enum value that will work with both C++03, and C++11?

blowekamp:
This patch has been rebased onto the master which is now using ITKv5.

PR #824, opened 2019-09-12
Add testing for proper image size after stride indexing

blowekamp:
This add testing for the image slicing being smaller than that of
numpy. The issue was raised on ITK Discourse:
https://discourse.itk.org/t/image-slicing-in-python/2160

The test added include the example provided.


zivy (approved):
LGTM (I don't like the x,y vs. y,x indexing, but that is a feature).

PR #827, opened 2019-09-13
Add testing for proper image size after stride indexing

blowekamp:
This add testing for the image slicing being smaller than that of
numpy. The issue was raised on ITK Discourse:
https://discourse.itk.org/t/image-slicing-in-python/2160

The test added includes the example provided on Discourse.


zivy (commented):
Whats the difference with PR #824?  Looks the same.

blowekamp:
This PR targets the release branch while the other is the master.

zivy (approved):

PR #839, opened 2019-10-09
On AZP Packaging with VS2008 limit link jobs

blowekamp:
To address timeout presumably from swapping during multiple linkages,
limit the number of concurrent linkages.


blowekamp:
py27 is no longer supported

PR #840, opened 2019-10-09
Only remove ITK Legacy code with default known version of ITK

blowekamp:
This enables SimpleITK to build against ITKv5 by
default. Particularly, ITK 5.1 uses new enums which are not
backwards compatible by default.


dave3d (approved):
Looks good to me.

PR #843, opened 2019-10-10
Add methods to get per label measure from LabelOverlapMeasures.

### LINKED ISSUE #837: LabelOverlapMeasuresImageFilter, get measures for specific label values
Please add a feature, so that it is possible to get overlap measures for specific label values as it is in the original ITK code of the filter. Thank you!

blowekamp:
closes #837.


dave3d (approved):
Looks good.

PR #851, opened 2019-10-21
Update Superbuild to Lua 5.3.5

blowekamp:

PR #852, opened 2019-10-21
Update Superbuild to Lua 5.2.4

blowekamp:


blowekamp:
PR #851 has tested to be compatible with Lua 5.2 as well.

PR #853, opened 2019-10-22
Do not use sudo to install python dependencies on AZP

blowekamp:


blowekamp:
This change is not needed.

PR #855, opened 2019-10-23
AZP MacOS disable using virtualenv for testing.

blowekamp:
There is an unknown error being generated when constructing the
virtual environment for python.


blowekamp:
PR  #859 solved the Mac build error

PR #857, opened 2019-10-25
Add support to Python for "pickling" an Image object

blowekamp:


zivy (approved):
Very nice. Good to have pickling, we'll be able to do heavy tasks (registration) in multiple processes. I'm a happy camper.

PR #864, opened 2019-10-30
Add option for nearest neighbor extrapolator for ResampleImageFilter

### LINKED ISSUE #747: ResampleImageFilter, missing method  SetExtrapolator()
ITK (native) allows to specify interpolation values outside of the image boundaries using SetExtrapolator() method of  ResampleImageFilter, 
https://itk.org/Doxygen/html/classitk_1_1ResampleImageFilter.html

However in SimpleITK (python), there is no way to set the Extrapolator
https://itk.org/SimpleITKDoxygen/html/classitk_1_1simple_1_1ResampleImageFilter.html

In particular when resampling 3D images to a different resolution with linear interpolation, there is visible artifact on the very last row/column (because the default Extrapolator is zero padding)

please add SetExtrapolator() support. thanks

blowekamp:
Closes #747.


zivy (approved):
LGTM. I guess this is useful for some people, though I have always used a fixed intensity value (again maybe because I always have air surrounding anatomy).

PR #882, opened 2019-11-19
Upgrade to using ITK version 5.1b01

blowekamp:
ITKV4 compatibility is enabled, along with not removing deprecated
code.

The goal of migration to ITKv5 for SimpleITK is to maintain a
compatible SimpleITK interface and provide as little disruption as
possible to the algorithms.


blowekamp:
Lets merge this update and fix things in master.

PR #884, opened 2019-11-19
Transform test comparison multi baseline

blowekamp:


blowekamp:
This code is ugly, but it seems to work OK. I created another issue to plan to clean it up: #885

This is needed to use additional baseline image for the change in registration results with ITKv5.

dave3d (approved):
Look all right.

PR #888, opened 2019-11-25
Add option to AZP packaging builds to skip 32-bit builds

blowekamp:


blowekamp:
The packaging has been totally refactored so this is no longer relevant.

PR #890, opened 2019-11-25
Add support to specify compression level and compressor

blowekamp:


dave3d (approved):
Very cool!

zivy (approved):
LGTM. Nice to have for large images (microscopy) with the ability to use maximal compression setting vs. the default (I assume it wasn't max).

blowekamp:
> LGTM. Nice to have for large images (microscopy) with the ability to use maximal compression setting vs. the default (I assume it wasn't max).

The default was something like 5 (0-9), but was recently changes to 2 after a large about of performance comparison: https://discourse.itk.org/t/adding-parallel-compression-to-metaio/1835 There was very little difference in size but a significant performance boost to reducing the default level.

PR #891, opened 2019-11-27
Revert "Disable the explicit template instantiating by default"

blowekamp:
This reverts commit ddde6bf0d262877ed3d822d160463f47d952c088.

Some AZP builds are running out of memory during linking. Using the instantiated library should help. I was not able to reproduce the linking error with the explicit instantiated library enabled.


blowekamp:
This was not a problem

Issue #460, opened 2018-05-07
Add support for pickling


adizhol:
https://discourse.itk.org/t/how-to-pickle-simpleitk-transforms/901/4

Currently SimpleITK does not support pickling of transformations and images.

Please add support for pickling transforms.

blowekamp:
This would be a very good feature to add to better support Python MultiProcessing module.

blowekamp:
closed by #857

PR #893, opened 2019-11-29
Add transform pickling

blowekamp:
addresses #460

### DIFF
--- a/Testing/Unit/Python/ConcurrentImageRead.py
+++ b/Testing/Unit/Python/ConcurrentImageRead.py
@@ -19,6 +19,8 @@ from __future__ import print_function
 import sys
 import os
 import unittest
+import tempfile
+import shutil
 
 
 import SimpleITK as sitk
@@ -28,10 +30,16 @@ from multiprocessing.pool import ThreadPool
 class ConcurrentImageRead(unittest.TestCase):
     """Testing thread concurrent reading of Images"""
 
-    temp_directory = ""
-
     def setUp(self):
-        pass
+
+        # Create a temporary directory for output files
+        self.test_dir = tempfile.mkdtemp()
+
+    def tearDown(self):
+
+        # Delete the temporary directory and files contained within. Perhaps if tests fail then the output should  stick
+        # around to debug the problem
+        shutil.rmtree(self.test_dir)
 
     def _create_data(self, img_extension, img_type):
         """Method to create the many temporary image for reading"""
@@ -44,8 +52,7 @@ class ConcurrentImageRead(unittest.TestCase):
         for i in range(0,n):
             img = sitk.GaussianSource(t, [s,s], mean=[256*(i+1)]*2)
             fname = "g_{0}.{1}".format(i, img_extension)
-            if ConcurrentImageRead.temp_directory:
-                fname = os.path.join(ConcurrentImageRead.temp_directory, fname)
+            fname = os.path.join(self.test_dir, fname)
             files.append(fname)
             sitk.WriteImage(img, files[-1])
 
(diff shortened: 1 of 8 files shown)


Wrapping/Python/Python.i
```diff
@@ -653,6 +653,55 @@
 
 }
 
+
+%extend itk::simple::Transform {
+   %pythoncode %{
+
+        def __copy__(self):
+          """Create a SimpleITK shallow copy, where the internal image is shared with a copy on write implementation."""
```
zivy: I believe this should say "internal transform"

Wrapping/Python/Python.i
```diff
@@ -653,6 +653,55 @@
...
+          return self.__class__(self)
+
+        def __deepcopy__(self, memo):
+          """Create a new copy of the data and internal itk::Image object."""
```
zivy: itk::Transform

Wrapping/Python/Python.i
```diff
@@ -653,6 +653,55 @@
...
+          dc = self.__class__(self)
+          dc.MakeUnique()
+          return dc
+
+        def __setstate__(self, args):
+          if args[0] != 0:
+            raise ValueError("Unable to handle SimpleITK.Image pickle version {0}".args[0])
```
zivy: ...SimpleITK.Transform pickle version...

zivy (approved):
Just some minor issues where things are referred to as Image and not Transform as expected.

Wrapping/Python/Python.i
blowekamp: done

Wrapping/Python/Python.i
blowekamp: done

Wrapping/Python/Python.i
blowekamp: done

blowekamp:
Sharp is failing with linking since the ITKv5 upgrade... Need to figure out how to reproduce the problem. The Windows CI error is not related.

Testing/Unit/Python/ConcurrentImageRead.py
```diff
@@ -28,10 +30,16 @@
...
-
     def setUp(self):
-        pass
+
+        # Create a temporary directory for output files
+        self.test_dir = tempfile.mkdtemp()
+
+    def tearDown(self):
+
+        # Delete the create temporary directory and files with  in. Perhaps if tests fail then the output should  stick
```
zivy: Possibly change "Delete the create temporary directory and files with   in." to "Delete the temporary files and directory."

Testing/Unit/Python/ImageReadWriteInt64.py
```diff
@@ -27,7 +29,27 @@
...
 
-    temp_directory = ""
+    def setUp(self):
+
+        # Create a temporary directory for output files
+        self.test_dir = tempfile.mkdtemp()
+
+    def tearDown(self):
+
+        # Delete the create temporary directory and files with  in. Perhaps if tests fail then the output should  stick
```
zivy: Same comment as above (it's the same text).

Testing/Unit/Python/sitkTransformTests.py
```diff
@@ -0,0 +1,192 @@
...
+
+        for transform_type in transform_type_list:
+            self.addTypeEqualityFunc(transform_type, self.assertTransformEqual)
+
+        # Create a temporary directory for output files
+        self.test_dir = tempfile.mkdtemp()
+
+    def tearDown(self):
+
+        # Delete the create temporary directory and files with  in. Perhaps if tests fail then the output should  stick
```
zivy: Same comment as above.

zivy (approved):
There was one comment with respect to text, but it is in three places (same text).

blowekamp:
@zivy The improved interface for constructing a BSplineTransform from a tuple of images, is that something that needs to be updated in the Notebooks?

zivy:
Thanks for pointing this out, I've opened an issue for this: https://github.com/InsightSoftwareConsortium/SimpleITK-Notebooks/issues/246

PR #908, opened 2019-12-17
Changed the github links from directly pointing to the latest release.

zivy:
The previous link led to a page which only contained information with
respect to the latest version/release. It is better to point to the
page where all releases appear and the user can search that
page. Otherwise they won't know where to go if by latest release, they
actually wanted the release and not latest development version.

### DIFF
--- a/Documentation/docs/source/installation.rst
+++ b/Documentation/docs/source/installation.rst
@@ -89,7 +89,7 @@ C# binary files
 
 **Binaries for select C# platforms** can be found on SimpleITK's
 `GitHub releases
-<https://github.com/SimpleITK/SimpleITK/releases/tag/latest>`__
+<https://github.com/SimpleITK/SimpleITK/releases>`__
 under the appropriate version. Installing the library should only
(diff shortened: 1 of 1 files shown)


Documentation/docs/source/installation.rst
```diff
@@ -120,7 +120,7 @@ If the latest SimpleITK bug fixes or features are required then
 pre-release binaries are needed. The pre-release binaries are built
 from the actively developed master branch. The SimpleITK packages are
 available on `Github releases
-<https://github.com/SimpleITK/SimpleITK/releases/tag/latest>`__ under
```
blowekamp: This is the one link that should remain to the latest.

blowekamp:
> Thank you. That will make it a bit easier to find the Java Windows binaries. :)

There are not Java binaries in the `latest` tag.

Documentation/docs/source/installation.rst
zivy: Reverted this and force pushed, so resolved.

PR #910, opened 2019-12-17
Add baseline image for SLIC filter

blowekamp:
The baseline image allows for a not exact image matching with a
tolerance. This is needed for windows 32-bits as the results are
slightly different.


blowekamp:
> I thought we were no longer supporting 32 bit.

We are no longer providing 32-bit binaries.
 
The library should continue to work on other architectures. Keeping a 32-bit build is the easiest thing to do to ensure that we have some appropriate amount of tolerances in our test results.

PR #920, opened 2019-12-19
Update itk modules

blowekamp:


zivy (approved):
There are many whitespace changes, any reason for these other than alignment?

blowekamp:
There should be a pre-commit hook which run a python JSON beautifier. Apparently, that has not always happened. The second commit in this PR, notes beatification style changes that occur with this batch processing of the JSONs.

zivy:
Got it. Thanks for the clarification.

PR #922, opened 2019-12-19
Update to SWIG 4.0.1

blowekamp:


dave3d (approved):
Looks good

blowekamp:
The update to SWIG 4.0 made some significant changes to the mapping of types, and the wrapped `std::vector` class. The integer types have been a little poorly defined and not quite the same on system to system.

With SimpleITK 2.0 we should get this documented and clarified how the integer types are bing mapped.

zivy (approved):

PR #923, opened 2019-12-19
Update RTD configuration

blowekamp:


blowekamp:
@philipmac This should get the current configuration into the file, so it can be easily updated after you PR is merged.

philipmac (approved):

PR #930, opened 2019-12-20
Add a sphinx build to CircleCI

blowekamp:


.circleci/config.yml
```diff
@@ -277,6 +277,32 @@ jobs:
...
+      SPHINX_BUILD_DIR: /tmp/docs
+      SPHINX_OPTS: -aE
+    steps:
+      - checkout:
+          path : ~/SimpleITK
+      - run:
+          name: Create Python Environment
+          command: |
+              python -m pip install --user --upgrade pip
+              python -m pip install --user "sphinx<2" "sphinx-rtd-theme<0.5"
```
blowekamp: Should the above packaged be moved to the requirements.txt file?

philipmac: yeah, it should probably be in `docs/requirements.txt`

PR #939, opened 2020-01-08
For external project add required flags to preprocessor

blowekamp:
Address configure warnings in PCRE and SWIG related to headers being
rejected by the preprocessor by include "isysroot" flag to
the preprocessor.


dave3d (approved):
Looks good

PR #942, opened 2020-01-09
Removes breadcrumbs template

philipmac:
Note, if you would like this file back and make breadcrumbs work: 
add {{ super() }} in the block.


blowekamp:
@zivy Where the breadcrumbs intentionally removed from the top of the sphinx pages?

zivy:
@blowekamp, I don't remember. The last time I extensively edited the sphinx documentation was a long while ago.

Issue #943, opened 2020-01-13
SetLabelForUndecidedPixels method not working in MultiLabelSTAPLEImageFilter class


hubutui:
Here is my script to test:
```python
import SimpleITK as sitk
import numpy as np
import os
import os.path as osp


if __name__ == '__main__':
    dirname = 'testmasks'
    imgs = [sitk.ReadImage(osp.join(dirname, item)) for item in os.listdir(dirname)]
    for img in imgs:
        print(np.unique(sitk.GetArrayFromImage(img)))
    filter = sitk.MultiLabelSTAPLEImageFilter()
    filter.SetLabelForUndecidedPixels(255)
    result = filter.Execute(*imgs)
    print(np.unique(sitk.GetArrayFromImage(result)))
```
The label value range from 0 to 6, but I got 7 in the result. According to the [doc](https://itk.org/SimpleITKDoxygen/html/classitk_1_1simple_1_1MultiLabelSTAPLEImageFilter.html), this filter set label for undecided pixel to the maximum of the label value plus 1.
And the testfile: [testmasks.tar.gz](https://discourse.itk.org/uploads/short-url/2gksrGk9973KuHPEDfuC4iDx2hN.gz)
SimpleITK version: 1.2.4
related post: see [here](https://discourse.itk.org/t/multilabelstapleimagefilter-introduce-new-label-value/2593)

zivy:
Hello @hubutui,

Thanks for reporting. This is a bug in ITK. Working on a pull request for ITK which will then fix it in SimpleITK.

blowekamp:
The issue has been addressed in ITK, and SimpleITK has been updated. The patch made it into the latest packaging for SimpleITK. Please see here for the instruction to download:
https://simpleitk.readthedocs.io/en/master/installation.html#latest-binaries

PR #944, opened 2020-01-13
Update ITK along 5.1 development

blowekamp:
This include the following patch:
commit 3d5698b3103e08d73e5657d3e1af7999e02a7ef9
Author: Ziv Yaniv <zivyaniv@nih.gov>
Date:   Mon Jan 13 11:15:00 2020 -0500

    BUG: SetLabelForUndecidedPixels value ignored.

    The user requested label for undecided values was ignored and the
    maxLabel+1 was always used. The test also had a bug in its ground
    truth values.

closes issue #943


blowekamp:
With the exiting test data, I was not able to create an undecided label with the MultiLableSTAPLEImageFilter.

zivy:
Possibly create two images programmatically (img1 = [1,1], img2=[1,2], ground truth = [1,255] second pixel is undetermined).

PR #946, opened 2020-01-14
Update ITK superbuild postv5.1rc01

blowekamp:


zivy (approved):
Supporting all non label pixel types is always a good thing.

PR #950, opened 2020-01-17
Remove ITK legacy code in Superbuild

blowekamp:
By default set ITK_LEGACY_REMOVE to ON, to aid in consistent behavior
with ITK.


dave3d (approved):
So long V4.

Issue #958, opened 2020-01-28
Memory access violations in dotnet code on concurrent access to the underlying buffer of an image


vincenzoml:
I have a sample function in FSharp that implements a pixel-by-pixel copy of a monochrome image by accessing the image buffers directly. I run the function a number of times in parallel on a multi-core machine, and the program crashes with various memory access violations. The function is designed to access the underlying buffer of the source image by calling "GetBufferAs..." in each thread (which is the same for each thread). Then a new image is created, its underlying buffer is accessed, and the copy is done.

If the "GetBuffer" method of the source image is called only once, outside of the various threads, the program works. If instead, "new" is *not* called and the source image is used as destination, the program works too. So this must be a bug in the interaction between "new(a)" and "a.getBufferAs". 

Anyway, to reproduce the bug, you can clone the VoxLogicA repository and checkout the "bugsquash" branch.

Detailed instructions

1) Install dotnet core sdk (I use version 3.1) from here:

https://dotnet.microsoft.com/download

2) Clone the VoxLogicA repository: 

git clone https://github.com/vincenzoml/VoxLogicA.git

This includes my own build of simpleitk. I've tried to build 1.2.4 with itk 4.something (default from current master branch of the simpleitk repository) and nothing changed. I've tried to change the itk tag to v5.0.1 but I can't build, I get errors while compiling simpleitk. 

3) Checkout the bugsquash branch:

git checkout bugsquash

4) cd src 

5) dotnet run -c linux-x64 (or win-x64, or osx-x64, depending on your operating system)

The program may finish (and print "all done") but very often it just crashes in unpredictable ways.

Here is the source code of the test program:

https://github.com/vincenzoml/VoxLogicA/tree/bugsquash/src

 
```
#nowarn "9"

open Microsoft.FSharp.NativeInterop
open System.Runtime.InteropServices
open itk.simple

let img = SimpleITK.Cast(SimpleITK.VectorIndexSelectionCast(SimpleITK.ReadImage("test.png"),0ul),PixelIDValueEnum.sitkFloat32)
img.MakeUnique()

let buflen = int (img.GetNumberOfPixels() * img.GetNumberOfComponentsPerPixel())

let task () =           
    // Move the following two lines, and the last one of this function, to the global scope, and the program works
    let imgHandle = GCHandle.Alloc(img)
    let imgPtr = NativePtr.ofNativeInt<uint8> (img.GetBufferAsFloat())    

[...]

PR #961, opened 2020-01-31
Add renamed const buffer access methods to C# interface.

blowekamp:
The implicit modification of the image by calling MakeUnique can cause
problem in with multi-threader. See issue #958 for more details.

### DIFF
--- a/Wrapping/CSharp/CSharp.i
+++ b/Wrapping/CSharp/CSharp.i
@@ -43,6 +43,24 @@
 %CSharpTypemapHelper( float*, System.IntPtr )
 %CSharpTypemapHelper( double*, System.IntPtr )
 
+
+// CSharp does not hadle overloaded const methods. So they are
+// explicitly renamed to enable their explicit usage. These methods
+// assume that the buffer will not be modified, so MakeUnique is
+// not call.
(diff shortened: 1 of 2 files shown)


vincenzoml:
I've tested this fix and it works. This resolves the problems raised in issue #961.

Wrapping/CSharp/CSharp.i
```diff
@@ -43,6 +43,24 @@
 %CSharpTypemapHelper( float*, System.IntPtr )
 %CSharpTypemapHelper( double*, System.IntPtr )
 
+
+// CSharp does not hadle overloaded const methods. So they are
```
dave3d: 'handle'

Wrapping/CSharp/CSharp.i
```diff
@@ -43,6 +43,24 @@
...
+// explicitly renamed to enable their explicit usage. These methods
+// assume that the buffer is will not be modified, so MakeUnique is
+// not call.
+%rename("GetConstBufferAsInt8")   itk::simple::Image::GetBufferAsInt8( ) const;
```
zivy: // assume that the buffer is will not be modified, so MakeUnique is
// not call.

change to:
// assume that the buffer will not be modified, so MakeUnique is
// not called.

PR #962, opened 2020-02-03
C make cpp docs

dave3d:


docs/source/cmake_cplusplus.rst
```diff
@@ -0,0 +1,58 @@
+A Simple C++/CMake Example
+==========================
+
+Building a C++ SimpleITK application is normally done using `CMake <https://cmake.org>`_ to
```
blowekamp: "is normally done using" -> "is supported by using"

Need to make it a little stronger that this is THE way to do it.

dave3d:
The Windows config thing gets me every time.

Will refine the grammer/language some more.

PR #963, opened 2020-02-04
C make cpp docs

dave3d:


blowekamp:
Should the example directory be added to here too:
https://github.com/SimpleITK/SimpleITK/blob/master/Examples/CMakeLists.txt

That would  build it as apart of the examples. It may not be building it standalone, but that nominal testing would be a good thing.

dave3d:
I don't think so, because the example has it's own CMakeLists.txt file which doesn't conform with the other example CMakeLists.txt files. Those ones have a stuff for testing, which doesn't really apply to this example.

blowekamp:
Right, but wouldn't it be good to just compile the code?

dave3d:
I tried putting enabling CppCMake in the Examples directory.  CMake didn't like it.  I got the following error message, which I can't figure out.

CMake Error at /home/dchen/sitk/SimpleITK-build/SimpleITKConfig.cmake:70 (include):
  The file

    /home/dchen/sitk/SimpleITK-build/SimpleITKTargets.cmake

  was generated by the export() command.  It may not be used as the argument
  to the include() command.  Use ALIAS targets instead to refer to targets by
  alternative names.

  Examples/CppCMake/CMakeLists.txt:5 (find_package)

blowekamp:
Something when wrong here with git.

dave3d:
I guess I messed up.  I was trying to rebase to master and then push it up to my GitHub, but I guess I pushed it up to origin.

PR #974, opened 2020-02-12
Fix finding correct lua interpreter version

blowekamp:
The lua which is targeted for wrapping can be different than the
version used for code generation. Correctly, save the wrapping version
when searching for the code generation versions, which is more version
specific.


dave3d (approved):
OK

zivy (approved):

PR #977, opened 2020-02-14
Explicitly cast int to ITK strongly typed enums

blowekamp:
This patch is require after additional adoption of strongly typed enums in ITK: InsightSoftwareConsortium/ITK#1471

### DIFF
--- a/Code/IO/src/sitkImageReaderBase.cxx
+++ b/Code/IO/src/sitkImageReaderBase.cxx
@@ -214,7 +214,7 @@ ImageReaderBase
   if (numberOfComponents == 1 &&
       ( pixelType == itk::ImageIOBase::SCALAR || pixelType == itk::ImageIOBase::COMPLEX ) )
     {
-    outPixelType = this->ExecuteInternalReadScalar( componentType );
+    outPixelType = this->ExecuteInternalReadScalar( static_cast<int>(componentType) );
     return;
     }
   // we try to load anything else into a VectorImage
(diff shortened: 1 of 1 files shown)


Code/IO/src/sitkImageReaderBase.cxx
```diff
@@ -214,7 +214,7 @@ ::GetPixelIDFromImageIO( const ImageIOBase *iobase,
   if (numberOfComponents == 1 &&
       ( pixelType == itk::ImageIOBase::SCALAR || pixelType == itk::ImageIOBase::COMPLEX ) )
     {
-    outPixelType = this->ExecuteInternalReadScalar( componentType );
+    outPixelType = this->ExecuteInternalReadScalar( static_cast<int>(componentType) );
```
hjmjohnson: @blowekamp I'm curious why the function signature is not changed so that component type is matched to the intent of "CommonEnums::IOComponent".  

Perhaps it is for backward compatibility?  If so I'd recommend 
```
#if ITK < 5.1
using SimpleITKInternalComponentType = int;
#else
using SimpleITKInternalComponentType = CommonEnums::IOComponent;
#endif
```

Just a thought to ease migration in the future.

blowekamp: In all of the SimpleITK public headers, there is no inclusion of an ITK header.

There are some forward declarations of types used as pointers. But for this case there the enums have sizes, I don't expect a forward declaration to work.

dave3d (approved):
You have typed things **strongly**.

zivy (approved):

PR #979, opened 2020-02-14
Add move operations to the Image class

blowekamp:


Testing/Unit/sitkImageTests.cxx
```diff
@@ -1638,6 +1639,29 @@ TEST_F(Image,MetaDataDictionary)
 
 }
 
+TEST_F(Image, MoveOperations)
+{
+  sitk::Image img;
+
+  static_assert(std::is_move_constructible<sitk::Image>::value);
```
blowekamp: C+11 `static_assert` required a second string argument.

dave3d (approved):
Cool stuff

blowekamp:
One issue to discuss with this PR is: After an image has been move from the image will only be valid for destruction ( and maybe assignment ). This is different than many STL container where the moved container is empty and any valid operation can occur. With this SimpleITK image class the internal pimple pointer will be null, can most operation will cause a segmentation fault after it has been moved.

zivy (approved):

PR #982, opened 2020-02-14
Replace sitkStaticAssert with standard static_assert

blowekamp:


hjmjohnson (approved):
Nice set of changes being made recently!

Code/Common/include/sitkMacro.h
```diff
@@ -115,18 +115,6 @@ class GenericException;
 #define sitkDoMacroJoin( X, Y ) sitkDoMacroJoin2(X,Y)
 #define sitkDoMacroJoin2( X, Y ) X##Y
```
blowekamp: The above defines are no longer needed.

blowekamp: This macro is still needed for clang specific macros.

PR #993, opened 2020-02-21
Use lambda over using nested std::bind

blowekamp:
The lambda wrapping is clearer and reduces the number of typedefs and
variables required.

The std::bind is still preferred when multiple arguments are pass, as
the std::placeholders work concisely, or when a simple binding of an
argument (this) is needed.


zivy (approved):
The lambdas do make the code more readable, nice.

PR #1001, opened 2020-03-04
Add lambda commands

blowekamp:


zivy (approved):
Nice addition using modern C++.

hjmjohnson (approved):

PR #1011, opened 2020-03-18
Adding example illustrating how to read a raw image file.

zivy:


blowekamp:
Looks like a good example.

To be consistent with the other example and facilitate automatic testing, and validation:
 - please produce an output file and check with a check sum.
 - please guard the use of Show with a check for the SITK_NOSHOW environment variable like what was done here:
https://github.com/SimpleITK/SimpleITK/blob/master/Examples/DicomSeriesReader/DicomSeriesReader.py#L45
  - The other examples don't check for "__main__", please be consistent with them.

zivy:
Addressed points 2 and 3. Changing the example to write an image in the end makes this closer to a test than an example, so I'd like to keep it as is, unless you feel this is really critical.

blowekamp:
Yes, it is the way they are all written and ensures they all keep working.

zivy:
Added checks returning zero if example succeeds or one if failure. This is not validation via file writing, but is consistent with other examples.

zivy:
Add output files and comparison via md5. There is some strange behavior, the md5 for the files written via Python is different from those via R even though the pixel values are the same.

blowekamp:
I'll take a look at this behavior

blowekamp:
I have updated the PR so it works with the test system.

In the future we may add images produced by the example to the Documentation.rst files. For here adding both endings would be nice.

blowekamp:
@ziv @dave3d Please check the review the update I pushed here.

PR #1013, opened 2020-03-20
Add wrapping for DICOMOrientImageFilter

### LINKED ISSUE #644: Feature request - Please add OrientImageFilter
I want to reorient the nii image to some direction such as RAI. It can work in itk, but simple-itk does not have the filter. Could you consider to integrate it in simple-itk?

This is some script itk code
```
using FilterType = itk::OrientImageFilter< InputImageType, OutputImageType>;
orientation_filter->SetDesiredCoordinateOrientation( itk::SpatialOrientation::ValidCoordinateOrientationFlags::ITK_COORDINATE_ORIENTATION_RAI);
orientation_filter->UseImageDirectionOn();
```

blowekamp:
Adding testing, and access to convert between orientation strings to
direction cosine matrix.

Closes #644.

### DIFF
--- a/Code/BasicFilters/json/DICOMOrientImageFilter.json
+++ b/Code/BasicFilters/json/DICOMOrientImageFilter.json
@@ -1,0 +1,151 @@
+{
+  "name" : "DICOMOrientImageFilter",
+  "template_code_filename" : "ImageFilter",
+  "template_test_filename" : "ImageFilter",
+  "number_of_inputs" : 1,
+  "pixel_types" : "typelist::Append<BasicPixelIDTypeList, VectorPixelIDTypeList>::Type",
+  "custom_register" : "this->m_MemberFactory->RegisterMemberFunctions< PixelIDTypeList, 3 > ();\n",
+  "public_declarations" : "static SITKBasicFilters_EXPORT std::string GetOrientationFromDirectionCosines( const std::vector<double> &direction );\n   static SITKBasicFilters_EXPORT std::vector<double> GetDirectionCosinesFromOrientation( const std::string &str );",
+  "filter_type" : "itk::DICOMOrientImageFilter<InputImageType>",
+  "members" : [
+    {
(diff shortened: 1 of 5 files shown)


Code/BasicFilters/json/DICOMOrientImageFilter.json
```diff
@@ -0,0 +1,151 @@
+{
+  "name" : "DICOMOrientImageFilter",
+  "template_code_filename" : "ImageFilter",
+  "template_test_filename" : "ImageFilter",
+  "number_of_inputs" : 1,
+  "pixel_types" : "BasicPixelIDTypeList",
```
zivy: I think this should also be for all pixel types (including label and vector), AllPixelIDTypeList. DICOM can also contain vector/color pixels. Also, if I have a matching segmentation I'd like to apply the filter to both the original DICOM scalar volume and to the label image.

Code/BasicFilters/src/sitkDICOMOrientImageFilter_Support.cxx
```diff
@@ -0,0 +1,50 @@
...
+*  See the License for the specific language governing permissions and
+*  limitations under the License.
+*
+*=========================================================================*/
+
+#include "sitkDICOMOrientImageFilter.h"
+#include "itkDICOMOrientation.h"
+
+// This file is intended to contain the definition of static
+// membervariables needed by JSON Expand templated image filters.
```
zivy: membervariables -> member variables

zivy (changes requested):

Code/BasicFilters/json/DICOMOrientImageFilter.json
blowekamp: We can add vector support. The `BasicPixelIDTypeList` includes the integers which are commonly used for label representation the Label pixel ID refer to the label map representation ( I know it's commonly confused ).

zivy (approved):

PR #1019, opened 2020-03-25
Remove Execute member function with parameters

blowekamp:
These extra Execute member functions are being removed, to reduce the
number of Execute methods in a filter. These methods are not
significantly used. The non-member functions accomplish a similar
interface, without the ambiguity in the instance's value caused by
current object instance value or defaulted parameter value.


dave3d (approved):
Looks good.  Simplifying the interface by remove extraneous stuff is a good thing.

zivy (approved):

PR #1027, opened 2020-03-31
Check for aliasing of image's pixel container

blowekamp:
To add additional safety with image buffer aliasing, the pixel
container of the itk::Image is now checked to verify that is only has
one reference. In filter's execute method the filter needs to be
deleted before the check to ensure the filter has no internal
references to the buffer.

### DIFF
--- a/Code/Common/src/sitkPimpleImageBase.hxx
+++ b/Code/Common/src/sitkPimpleImageBase.hxx
@@ -78,6 +78,8 @@ namespace itk
                               << "SimpleITK does not support streaming or unbuffered regions!" );
           }
 
+        InternalCheckBuffer(image);
+
         const IndexType & idx = image->GetBufferedRegion().GetIndex();
         for ( unsigned int i = 0; i < ImageType::ImageDimension; ++i )
           {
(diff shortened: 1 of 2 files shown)


Code/Common/src/sitkPimpleImageBase.hxx
```diff
@@ -722,6 +724,38 @@ namespace itk
...
+      }
+
+    template< typename TPixelType, unsigned int NDimension >
+    bool
+    InternalCheckBuffer(const itk::VectorImage<TPixelType, NDimension> *image)
+      {
+        auto container = image->GetPixelContainer();
+        if (container->GetReferenceCount() != 1)
+          {
+          sitkExceptionMacro( << "The image pixel container is shared by other resources and presents aliasing issues.");
```
zivy: Maybe modify the exception message between the scalar image and vector image, just to make it easier to localize issues?

zivy (approved):
Minor comment, not sure about it myself.

dave3d (approved):
LGTM

Code/Common/src/sitkPimpleImageBase.hxx
blowekamp: The exceptions contain the line number for localization, but we can certainly add "vector image" here.

PR #1028, opened 2020-03-31
Refactor input image validation

blowekamp:
Factor method to validate inputs match dimensions, size, pixel
    
    The extracted functions provide a common place for an error
    message. The error message have been improved to uniformly describe
    the filter, and the input parameters which have a problem.

### DIFF
--- a/Code/BasicFilters/include/sitkImageFilter.h
+++ b/Code/BasicFilters/include/sitkImageFilter.h
@@ -92,9 +92,21 @@ namespace itk {
 
       }
 
-    };
+      /** Verify the dimension of image1 matches the dimension of
+       * image2, and if not then an exception is thrown.
+       */
+      void CheckImageMatchingDimension(const Image &image1, const Image& image2, const std::string &image2Name );
 
+      /** Verify the pixel type of image1 matches the pixel type of
+       * image2, and if different then an exception is thrown.
+       */
+      void CheckImageMatchingPixelType(const Image &image1, const Image& image2, const std::string &image2Name );
 
+      /** Verify the size in pixel of image1 matches the size of
+       * image2, and if different then an exception is thrown.
+       */
+      void CheckImageMatchingSize(const Image &image1, const Image& image2, const std::string &image2Name );
+  };
   }
 }
 #endif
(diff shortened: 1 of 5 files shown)


Code/BasicFilters/include/sitkImageFilter.h
```diff
@@ -92,9 +92,21 @@ namespace itk {
 
       }
 
-    };
+      /** Verify the dimension of image1 matches the dimension of
+       * image2, if not then an exception is thrown.
```
dave3d, line 96: Insert 'and' before the 'if'.  A comma by itself isn't a conjunction.

Code/BasicFilters/include/sitkImageFilter.h
```diff
@@ -92,9 +92,21 @@ namespace itk {
...
+       */
+      void CheckImageMatchingDimension(const Image &image1, const Image& image2, const std::string &image2Name );
 
+      /** Verify the pixel type of image1 matches the pixel type of
+       * image2, if different then an exception is thrown.
```
dave3d, line 101: 'and' here too

Code/BasicFilters/include/sitkImageFilter.h
```diff
@@ -92,9 +92,21 @@ namespace itk {
...
+       */
+      void CheckImageMatchingPixelType(const Image &image1, const Image& image2, const std::string &image2Name );
 
+      /** Verify the size in pixel of image1 matches the size of
+       * image2, if differenet then an exception is thrown.
```
dave3d, line 106: 'and'.  Also different has a typo.

zivy (approved):
Just looked at the changes, so I'm curious if there is a check for pixel type, dimension and size, shouldn't there be a check for matching spacing? Probably something I am missing in the context.

blowekamp:
> Just looked at the changes, so I'm curious if there is a check for pixel type, dimension and size, shouldn't there be a check for matching spacing? Probably something I am missing in the context.

These are not new checks. It refactors them to a single location and improves the error messages to be more descriptive, and gives a suggestion when the pixel types do not match as expected.

The PxielType and Dimension are handled here to give a more informative message than a dynamic_cast failure in the InternalExecute method. These are more for a template parameter checking.

The Size is checked because when the size in ITK don't match an obscure pipeline image region message if thrown from ITK.

The pixel location information is handled by the ITK filter and gives a reasonable error message. We do not which to duplicate this.

PR #1030, opened 2020-04-01
Remove specific FastMarching filter template

blowekamp:
The general image filter template work, with code to convert the
SimpleITK trial seed points to the itk node representation. This was
also done in the FastMarchingImageFilter, now the
FastMarchingBaseImageFilter uses the same approach. This reduces the
maintenance burden on the template code.


dave3d (approved):
Cool

zivy (approved):

blowekamp:
This change also removed the AddTrialPoints and ClearTrialPoints methods.

We need find a clear way to communicate these API changes to the users.

PR #1034, opened 2020-04-01
Explicitly set SimpleITKFilter git hash

blowekamp:
The module update include a fix to the initialization to the
DICOMOrientImageFilter's array parameters.


dave3d (approved):
A+

PR #1039, opened 2020-04-02
Changing the RTD documentation to use the new logo.

zivy:


blowekamp:
The problem with renaming the file is that the links will go bad, and need to be updated.

PR #1040, opened 2020-04-02
Renaming installation to getting started and some minor edits.

zivy:


blowekamp:
The problem with renaming the file is that the links will go bad, and need to be updated.

I was thinking there would be a getting started subsection at the beginning of this document.

PR #1044, opened 2020-04-07
Enable keyword argument for Resample procedures

blowekamp:
SWIG does not support keyword argument for Python with C++ overloaded
functions. Therefore, the SWIG generate procedure is remove and
replaced with a manually written method that uses the positional
argument to determine the correct method signature.


zivy (approved):
LGTM. Syntactic sugar, but will make Python programmers happy.

PR #1049, opened 2020-04-08
Use value parameter and move for Set methods of std::vectors

blowekamp:
Improves performance by removing an std::vector copy, when the
argument is initialized with an rvalue reference, and then the
parameter is moved into the member variable. Otherwise the argument
will be initialized with a copy, then a move operation performed.


blowekamp:
Unfortunately when SWIG wraps a method such as: `void SetVector(std::vector<int> v)` it creates a temporary `std::vector` (when converting for a native Python type), but the temporary is not moved into the method's parameter. If the SWIG wrapping can be improved to do the `std::move`, then an extra copy will be removed.

blowekamp:
Despite note getting any advantage from this change from SWIG (yet), it is still good to move forward with this modern best practice in our interface.

zivy (approved):

PR #1053, opened 2020-04-09
Set ExternalData path to reuse downloaded files

blowekamp:
In the manylinux distribution builds, automatically set a location to
for the CMake external data so that it can be reused between python
builds.

### DIFF
--- a/Utilities/Distribution/manylinux/imagefiles/cmd.sh
+++ b/Utilities/Distribution/manylinux/imagefiles/cmd.sh
@@ -13,6 +13,11 @@ PYTHON_VERSIONS=${PYTHON_VERSIONS:-$(ls /opt/python | sed -e 's/cp2[0-6][^ ]\+ \
 NPROC=$(grep -c processor /proc/cpuinfo)
 export MAKEFLAGS="-j ${NPROC}"
 
+# if ExternalData_OBJECT_STORES is not set by the driver script then
+# set it here to enable  reuse of downloaded files between python
+# builds.
+export ExternalData_OBJECT_STORES=${ExternalData_OBJECT_STORES:-/tmp/.ExternalData}
+mkdir -p ${ExternalData_OBJECT_STORES}
 
 function build_simpleitk {
 


Utilities/Distribution/manylinux/imagefiles/cmd.sh
```diff
@@ -5,6 +5,7 @@ set -x
 export SRC_DIR="/tmp/SimpleITK"
 export BLD_DIR="/tmp/SimpleITK-build"
 export OUT_DIR="/work/io"
+ExternalData_OBJECT_STORES=/home/blowekamp/.ExternalData
```
zivy, line 8: Should this be $HOME/.ExternalData ?

zivy (commented):
Minor potential issue.

Utilities/Distribution/manylinux/imagefiles/cmd.sh
blowekamp: Nice catch, that line should not be there.

hjmjohnson: Just a suggestion:

```bash
# Use ExternalData_OBJECT_STORES as a local cache for downloaded objects.
if [ -d ${HOME}/.ExternalData ] & [  -z "${ExternalData_OBJECT_STORES}" ]; then
  ExternalData_OBJECT_STORES=${HOME}/.ExternalData
fi

```

blowekamp: Thanks for the comment.

The line below was what I wanted to do:
```
export ExternalData_OBJECT_STORES=${ExternalData_OBJECT_STORES:-/tmp/.ExternalData}
```

This is run in a docker image, frequently driven by [this script](https://github.com/SimpleITK/SimpleITK/blob/cf5c60a4294c8409888cf03d77ab6114a6de45a2/Utilities/Distribution/manylinux/run.sh#L3), which propagates the user env variable, and mount point. When not set, the line below should set up the cache env.

I'll add a comment to clarify.

zivy (approved):

PR #1056, opened 2020-04-10
Updating Doxygen.

zivy:
Replaced old logo with new one. Updated the main page and removed old
documentation that is no longer relevant.


blowekamp:
Nice use of SVG file.

The interactive improvements on the opening section look good. Although it could be debated if a more complete rewrite is needed following the style of the new Home Page.

PR #1062, opened 2020-04-14
Tweaks to CI Docker based builds

blowekamp:
Improve consistency between Doxygen and manylinux docker build scripts.

### DIFF
--- a/Utilities/Distribution/manylinux/imagefiles/cmd.sh
+++ b/Utilities/Distribution/manylinux/imagefiles/cmd.sh
@@ -25,10 +25,14 @@ mkdir -p ${PYTHONUSERBASE}
 
 function build_simpleitk {
 
-    echo "SIMPLEITK_GIT_TAG: ${SIMPLEITK_GIT_TAG}"
 
-    git clone https://github.com/SimpleITK/SimpleITK.git ${SRC_DIR} &&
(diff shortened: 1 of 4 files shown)


Utilities/Doxygen/docker/run.sh
```diff
@@ -1,19 +1,30 @@
 #!/usr/bin/env bash
 
-TAG=${TAG:-"master"}
+
+
+SRC_DIR="/work/SimpleITK"
+BLD_DIR="/work/SimpleITK-build"
+
+# If the SRC_DIR is not mounted, then this tag will be checkouted out
```
zivy, line 8: checkouted -> checked

Utilities/Distribution/manylinux/imagefiles/cmd.sh
```diff
@@ -25,10 +25,14 @@ mkdir -p ${PYTHONUSERBASE}
 
 function build_simpleitk {
 
-    echo "SIMPLEITK_GIT_TAG: ${SIMPLEITK_GIT_TAG}"
```
zivy, line 28: Is removing the logging message intentional? No need for it, but I assume it was previously useful.

blowekamp: With the below feature implemented of using the mounted volume as the source, and taking priority over this variable, it was not worth the effort/logic to correctly report the tag.

PR #1068, opened 2020-04-15
Moderize to use = default for constructors and destructors

blowekamp:
The change was generated by:

run-clang-tidy -p /scratch/blowekamp/SimpleITK-ITK
  -extra-arg=-D__clang__  -header-filter=.* -fix
  -checks=-*,modernize-use-equals-default


zivy (approved):
Much nicer that the compiler creates the default destructor instead of the manual empty one.

PR #1069, opened 2020-04-15
Remove ImageFilter's uint template parameter

blowekamp:
This parameter is no longer being used. Removing it will reduce the
code size.


dave3d (approved):
Nice

zivy (approved):

blowekamp:
@richardbeare I had to update some of the R namespace exporting. Are there any additional cleanup we can do the the R Wrapping with SWIG 4.0 and SimpleITK 2.0?

PR #1074, opened 2020-04-20
Doxygen configuration improvements

blowekamp:
Add interactive SVG. 
Use client side MathJAX for LaTeX math rendering.

### DIFF
--- a/Utilities/Doxygen/doxygen.config.in
+++ b/Utilities/Doxygen/doxygen.config.in
@@ -2391,7 +2391,7 @@ DIRECTORY_GRAPH        = YES
 # The default value is: png.
 # This tag requires that the tag HAVE_DOT is set to YES.
 
-DOT_IMAGE_FORMAT       = png
+DOT_IMAGE_FORMAT       = svg
 
 # If DOT_IMAGE_FORMAT is set to svg, then this option can be set to YES to
 # enable generation of interactive SVG images that allow zooming and panning.
@@ -2403,7 +2403,7 @@ DOT_IMAGE_FORMAT       = png
 # The default value is: NO.
 # This tag requires that the tag HAVE_DOT is set to YES.
 
-INTERACTIVE_SVG        = NO
+INTERACTIVE_SVG        = YES
 
 # The DOT_PATH tag can be used to specify the path where the dot tool can be
 # found. If left blank, it is assumed the dot tool can be found in the path.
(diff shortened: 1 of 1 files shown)


blowekamp:
@jhlegarreta @thewtex @hjmjohnson I'm adding some modern Doxygen client side features here, that ITK may be interested in.

zivy (approved):

jhlegarreta:
:+1: for changes that make documentation pages have a nicer rendering.

Thanks Brad.

May be once this gets merged, and the new documentation for `master` is deployed, posting here a page of a previous version and a current one would be useful to see the difference.

dave3d (approved):

jhlegarreta:
Following Brad's announcement in discourse, I have visited the SimpleITK Doxygen website. I chose to visualize:
https://simpleitk.org/doxygen/latest/html/classitk_1_1simple_1_1AdditiveGaussianNoiseImageFilter.html#details

to see how equations or images are rendered. It looks like MathJax is not working out of the box for me. I tried both Chrome and Firefox and the equations are not rendered correctly. Do users need to enable some add-on?

I could not compare to former versions of the documentation: I found no easy way of visiting the documentation of former releases from the Doxygen website itself (also applies to ITK).

blowekamp:
@jhlegarreta The doxygen did not get this update last night due to the URL changes, and scripts ( see #1076 for fixes). You are looking at the output before these changes. The previous Doxygen generation here never had LaTex Math working 🤪 

https://simpleitk.org has a nice pull down menu to go to old versions. Agree it would be nice to have a similar option on the doxygen pages them selves.

blowekamp:
@jhlegarreta The Doxygen was successfully generated and updated last night. The [AdditiveGaussianNoise](https://simpleitk.org/doxygen/latest/html/classitk_1_1simple_1_1AdditiveGaussianNoiseImageFilter.html) is a good example the client side MathJAX and the interactive SVG in the inheritance and collaboration diagrams.

jhlegarreta:
> @jhlegarreta The Doxygen was successfully generated and updated last night. The AdditiveGaussianNoise is a good example the client side MathJAX and the interactive SVG in the inheritance and collaboration diagrams.

Yes, I confirm that it looks good on my side as well. Thanks Brad !

blowekamp:
Do you think we should do the same thing for ITK?

jhlegarreta:
> Do you think we should do the same thing for ITK?

I would tend to say yes. Not sure whether there are downsides to it that have prevented from doing it before. A blame on the relevant file in ITK shows that the last edition to the MathJax option, for example, was 9 years ago.

hjmjohnson:
@blowekamp @jhlegarreta FYI:  I plan to apply for a small amount of funding to help support efforts like these.  I'll put this on the list for ITK tasks to do.

PR #1079, opened 2020-04-21
Cpp c make

dave3d:
Ugh.  Git.


Examples/CppCMake/CMakeLists.txt
```diff
@@ -0,0 +1,16 @@
+
+add_executable ( sitk_example Source/sitk_example.cxx )
+target_link_libraries ( sitk_example ${SimpleITK_LIBRARIES} )
+
```
blowekamp, line 4: Please add a comment about this is for testing the cxx file, and the subdirectory should not be source for because...

PR #1084, opened 2020-04-27
Update superbuild ITK version to 5.1.0

blowekamp:


zivy (approved):

dave3d (approved):

blowekamp:
@hjmjohnson  SimpleITK experiencing some regressions with the MattesMutualInformation.
https://open.cdash.org/testDetails.php?test=876115445&build=6491154

This test failure is a very basic case of running MM with the identity transform and the same image for the input and fixed. From the above the metric value changed from -1.52 to  11.09. The difference is above numeric tolerances, and may indicate a bug with the computations. Here is the test code:

https://github.com/SimpleITK/SimpleITK/blob/0bc9517aff1eddc8472ea66f1a23ef1746ffa0be/Testing/Unit/sitkImageRegistrationMethodTests.cxx#L221
https://github.com/SimpleITK/SimpleITK/blob/master/Code/Registration/src/sitkImageRegistrationMethod.cxx#L1038-L1099

There is also a BSpline registration test failing, but that needs further examination to determine if the change is really significant. However it does have a similar change in value, where the metric was ~-1 and now it's ~10.

I am hoping the sparks some magic insight as to the source of the change, or further explaination.

PR #1090, opened 2020-05-02
Add CMake SimpleITK_MAX_DIMENSION option

blowekamp:
This option replaces the SimpleITK_4D_Image option, and will allow for instantiation of higher dimensional images and time series.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -120,15 +120,22 @@ sitk_legacy_naming(SimpleITK_INT64_PIXELIDS)
 
 
 
-set(SimpleITK_4D_IMAGES_DEFAULT ON)
+set(SimpleITK_MAX_DIMENSION_DEFAULT 4)
 # 1900 = VS 14.0 (Visual Studio 2015)
 if(MSVC AND MSVC_VERSION VERSION_LESS 1900)
-  set( SimpleITK_4D_IMAGES_DEFAULT OFF )
-  message("disable")
+  set( SimpleITK_MAX_DIMENSION_DEFAULT 3 )
 endif()
-option( SimpleITK_4D_IMAGES "Add Image and I/O support for four spatial dimensions." ${SimpleITK_4D_IMAGES_DEFAULT} )
-mark_as_advanced( SimpleITK_4D_IMAGES )
-sitk_legacy_naming(SimpleITK_4D_IMAGES)
+if (DEFINED SimpleITK_4D_Image)
+  message(WARNING "SimpleITK variable \"SimpleITK_4D_Image\" is \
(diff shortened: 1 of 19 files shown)


CMakeLists.txt
```diff
@@ -120,15 +120,22 @@ sitk_legacy_naming(SimpleITK_INT64_PIXELIDS)
...
 if(MSVC AND MSVC_VERSION VERSION_LESS 1900)
-  set( SimpleITK_4D_IMAGES_DEFAULT OFF )
-  message("disable")
+  set( SimpleITK_MAX_DIMENSION_DEFAULT 3 )
 endif()
-option( SimpleITK_4D_IMAGES "Add Image and I/O support for four spatial dimensions." ${SimpleITK_4D_IMAGES_DEFAULT} )
-mark_as_advanced( SimpleITK_4D_IMAGES )
-sitk_legacy_naming(SimpleITK_4D_IMAGES)
+if (DEFINED SimpleITK_4D_Image)
+  message(WARNING "SimpleITK variable \"SimpleITK_4D_Image\" is \
```
zivy, line 129: Maybe say "set the \"SimpleITK_MAX_DIMENSION\" to 4 instead."?

SuperBuild/SuperBuild.cmake
```diff
@@ -144,14 +144,26 @@ option(BUILD_SHARED_LIBS "Build SimpleITK ITK with shared libraries. This does n
...
 if(MSVC AND MSVC_VERSION VERSION_LESS 1900)
-  set( SimpleITK_4D_IMAGES_DEFAULT OFF )
+  set( SimpleITK_MAX_DIMENSION_DEFAULT 3 )
 endif()
-option( SimpleITK_4D_IMAGES "Add Image and I/O support for four spatial dimensions." ${SimpleITK_4D_IMAGES_DEFAULT} )
-mark_as_advanced( SimpleITK_4D_IMAGES )
-unset(SimpleITK_4D_IMAGES_DEFAULT)
+
+if (DEFINED SimpleITK_4D_Image)
+  message(WARNING "SimpleITK variable \"SimpleITK_4D_Image\" is deprecated use \"SimpleITK_MAX_DIMENSION\" instead." )
```
zivy, line 154: same comment as above

CMakeLists.txt
blowekamp: We only know that SimpleITK_4D_Image is defined and not is value, so we don't know what value to recommend it be set to.

blowekamp:
The AZP build finished are reported to a green dashboard to CDash. I'm not sure why Github did not get notification.

PR #1094, opened 2020-05-05
Extract multi dimension support

blowekamp:


blowekamp:
To enable the ExtractImageFilter in SimpleITK to extract an arbitrary dimensional image e.g. a 5D->3D, the filter needed an manual SimpleITK layer as the code generation can not handle the secondary dispatching on the output dimension.

The big item left to do is to update the filter level documentation.

zivy (approved):

PR #1096, opened 2020-05-06
Set SimpleITK_MAX_DIMENSION to 5 by default

blowekamp:


zivy (approved):
Increasing dimensionality support is awesome, slowly but surely we too will reach string theory's 26 spacetime dimensions ;) .

dave3d (approved):

PR #1105, opened 2020-05-12
Python Add Pasting to setitem operator

blowekamp:


zivy (approved):

Wrapping/Python/Python.i
```diff
@@ -577,9 +585,17 @@
 
 
         def __setitem__( self, idx, value ):
-            """Sets the pixel value at index idx to value.
+            """Sets pixel value(s) values.
```
dave3d, line 588: what does 'value(s) values' mean?  is value supposed to be there twice?

Wrapping/Python/Python.i
```diff
@@ -577,9 +585,17 @@
...
+
+            The dimension of idx should match that of the image.
 
-            The dimension of idx should match that of the image."""
+            If all indices are intengers then value should be a pixel value
```
dave3d, line 592: 'integers'

Wrapping/Python/Python.i
```diff
@@ -577,9 +585,17 @@
...
+            ( scalar or sequence for vector pixels). Value is assigned pixel.
+
+            If all indices are slice indices then, value should be an image
+            of matching size, dimension and type as the image and region. This
+            image's pixel and then assigned the value image with the paste filter.
```
dave3d, line 597: do you mean 'are' instead of 'and'?

Wrapping/Python/Python.i
blowekamp: I renamed the argument to rvalue.

PR #1116, opened 2020-05-27
Adding example illustrating conversion from DICOM.

zivy:
Example traverses a directory structure and converst all DICOM files
found. Output written to optional output directory.


zivy:
All comments from @dave3d addressed.

Examples/DicomConvert/DicomConvert.py
```diff
@@ -0,0 +1,110 @@
...
+import functools
+import pandas as pd
+import itertools
+
+def convert_image(input_file_name, output_file_name, new_width=None):
+    try:
+        image_file_reader = sitk.ImageFileReader()
+        # only read DICOM images
+        image_file_reader.SetImageIO('GDCMImageIO')
+        image_file_reader.SetFileName(input_file_name)
```
blowekamp, line 33: Please use the following code:
```
image_file_reader.ReadImageInformation()
image_size = list(image_file_reader.GetSize())
if len(image_size) == 3 and image_size[2] == 1:
    image_size[2] = 0
    image_file_reader.SetExtractSize(image_size)
```
Not only does it remove a image copy, it also preserved the DICOM meta-data dictionary in the output image,

Examples/DicomConvert/DicomConvert.py
```diff
@@ -0,0 +1,110 @@
...
+        image = image_file_reader.Execute()[:,:,0]
+        if new_width:
+            original_size = image.GetSize()
+            original_spacing = image.GetSpacing()
+            new_spacing = [(original_size[0]-1)*original_spacing[0]/(new_width-1)]*2
+            new_size = [new_width, int((original_size[1]-1)*original_spacing[1]/new_spacing[1])]
```
blowekamp, line 39: This image math does not preserve the extent of the image. With the origin pixel's extent going down to the continuous index [-0.5]*d, the new image may have a large extent than the input image.

Examples/DicomConvert/DicomConvert.py
```diff
@@ -0,0 +1,110 @@
...
+            image = sitk.Resample(image, new_size, sitk.Transform(), sitk.sitkLinear,
+                                  image.GetOrigin(), new_spacing, image.GetDirection(),0,
```
blowekamp, line 41: Please use the new keyword arguments here, the 0 parameter is very misterious.

Examples/DicomConvert/DicomConvert.py
```diff
@@ -0,0 +1,110 @@
...
+                                  image.GetPixelID())
+        # If a single channel image, rescale to [0,255]. Also modify the intensity values based
+        # on the photomertic interpretation. If MONOCHROME2 (minimum should be displayed as black) we
+        # don't need to do anything, if image has MONOCRHOME1 (minimum should be displayed as white) we flip
+        # the intensities. This is a constraint imposed by ITK which always assumes MONOCHROME2.
+        if image.GetNumberOfComponentsPerPixel() == 1:
+            image = sitk.RescaleIntensity(image, 0, 255)
+            if image_file_reader.GetMetaData('0028|0004').strip() == 'MONOCHROME1':
```
blowekamp, line 49: With the above IO region change, the meta-data may be in the image, if the resample hasn't been done. Perhaps the rescale intensities should come first? I'm not sure if the inverting the default pixel value would have and effect.

Examples/DicomConvert/DicomConvert.py
```diff
@@ -0,0 +1,110 @@
...
+#
+#=========================================================================
+
+import SimpleITK as sitk
+import multiprocessing
+import argparse
+import os
+import sys
+import functools
+import pandas as pd
```
blowekamp, line 25: Please don't use pandas in examples. We don't want the additional dependency.

Examples/DicomConvert/DicomConvert.py
```diff
@@ -0,0 +1,110 @@
...
+        file_names = [os.path.join(os.path.abspath(args.od), str(i)) for i in range(len(input_file_names))]
+    else:
+        file_names = input_file_names
+    output_file_names = [file_name+'.'+args.output_file_extension for file_name in file_names]
+
+    res = convert_images(input_file_names, output_file_names, args.w)
+    input_file_names = list(itertools.compress(input_file_names, res))
+    output_file_names = list(itertools.compress(output_file_names, res))
+
+    # save csv file mapping input and output file names
```
blowekamp, line 104: It may be a few more lines of code and a loop, but you should be able to use the lower level built-in `csv` python module.

Examples/DicomConvert/DicomConvert.py
```diff
@@ -0,0 +1,110 @@
...
+            new_size = [new_width, int((original_size[1]-1)*original_spacing[1]/new_spacing[1])]
+            image = sitk.Resample(image, new_size, sitk.Transform(), sitk.sitkLinear,
+                                  image.GetOrigin(), new_spacing, image.GetDirection(),0,
+                                  image.GetPixelID())
+        # If a single channel image, rescale to [0,255]. Also modify the intensity values based
+        # on the photomertic interpretation. If MONOCHROME2 (minimum should be displayed as black) we
+        # don't need to do anything, if image has MONOCRHOME1 (minimum should be displayed as white) we flip
+        # the intensities. This is a constraint imposed by ITK which always assumes MONOCHROME2.
+        if image.GetNumberOfComponentsPerPixel() == 1:
+            image = sitk.RescaleIntensity(image, 0, 255)
```
blowekamp, line 48: You might want to run Min/Max filter to verify that the input it high-dynamic range before rescaling.

Examples/DicomConvert/DicomConvert.py
```diff
@@ -0,0 +1,110 @@
...
+        return False
+
+
+def convert_images(input_file_names, output_file_names, new_width):
+    # Maximal number of processes in pool, which in theory we'd like to be
+    # multiprocessing.cpu_count()-1. It appears that for some systems
+    # cpu_count() does not return the correct value
+    # (https://stackoverflow.com/questions/31344582/python-multiprocessing-cpu-count-returns-1-on-4-core-nvidia-jetson-tk1).
+    # We therefore don't rely on it and set the value explicitly here.
+    MAX_PROCESSES = 15
```
blowekamp, line 64: Please skip this "detail" in the example as it add clutter. Using 16 processes on the CI system could cause problems too.

blowekamp:
The example also needs to be added to Examples/CMakeLists.txt

Examples/DicomConvert/DicomConvert.py
zivy: Done

Examples/DicomConvert/DicomConvert.py
zivy: The extent will be smaller because of the cast to int in the `new_size`. Because of the constraint of isotropic spacing and user specified width the height will almost always lead to a different image extent. Not a critical thing when the goal is display.

Examples/DicomConvert/DicomConvert.py
zivy: Done.

Examples/DicomConvert/DicomConvert.py
zivy: The meta-data is lost in 95% of the image types (jpg,png...), so not really a concern.

Examples/DicomConvert/DicomConvert.py
zivy: Done.

Examples/DicomConvert/DicomConvert.py
zivy: Done.

Examples/DicomConvert/DicomConvert.py
zivy: Not critical, so leaving as is.

Examples/DicomConvert/DicomConvert.py
zivy: Done.

blowekamp:
There is a configuration error:
```
CMake Error at /usr/local/lib/python3.8/site-packages/cmake/data/share/cmake-3.15/Modules/ExternalData.cmake:666 (message):
  Data directory referenced by argument

    DATA{/home/circleci/SimpleITK/Wrapping/Python/../../Testing/Data/Input/DicomSeries/}

```
There following CMake code may be needed:
```
DATA{${SimpleITK_DATA_ROOT}/Input/DicomSeries/,REGEX:Image[0-9]+.dcm}
```

blowekamp:
@zivy The new example didn't show up on RTD. I found the following warning in the build log:
```
  app.add_stylesheet("style.css")
../../Examples/index.rst:6: WARNING: toctree contains reference to nonexisting document 'link_DicomConvert_docs'
looking for now-outdated files... none found
```
https://readthedocs.org/projects/simpleitk/builds/11144291/

blowekamp:
I'd also recommend renaming it example to DICOM Convert and Resample... so that it's clustered/grouped with the other DICOM examples.

zivy:
Forgot to commit that file, #1119  addresses this.

PR #1119, opened 2020-05-29
Adding missing file for building documentation.

zivy:


blowekamp:
Manually having to add these files is tedious. They really should be automatically generated perhaps in conf.py.

PR #1120, opened 2020-05-29
Updated Python examples for PEP8 compliance

dave3d:


Examples/FilterProgressReporting/FilterProgressReporting.py
```diff
@@ -15,57 +15,60 @@
...
-if len ( sys.argv ) < 4:
-    print( "Usage: "+sys.argv[0]+ " <input> <variance> <output>" )
-    sys.exit ( 1 )
+if len(sys.argv) < 4:
+    print("Usage: " + sys.argv[0] + " <input> <variance> <output>")
+    sys.exit(1)
 
-##! [python director command]
+
+# #! [python director command]
```
zivy, line 31: What does this comment mean? Are these directives for sphinx? The directives appear in this file in other places too.

Examples/AdvancedImageReading/AdvancedImageReading.py
```diff
@@ -15,16 +15,16 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 #
-#=========================================================================
+# =========================================================================
 
 from __future__ import print_function
```
zivy, line 20: Possibly remove all of the `from __future__ import print_function` as we no longer support Python 2.7. Probably in all examples.

Examples/FilterProgressReporting/FilterProgressReporting.py
blowekamp: I think the ##! mark up may be important to sphinx.

dave3d: Oh, ok.  I can undo that.  pycodestyle says there should be a space, but we can ignore that.

blowekamp:
Is there a configuration file for the style that should be check in too?

zivy:
Possibly run [flake8](https://flake8.pycqa.org/en/latest/) as part of the CI?

dave3d:
I did use any configuration file.  I just ran pycodestyle and then fixed things via pycharm and by hand.

Issue #1112, opened 2020-05-18
Transform Base class is not pickable


reox:
As discussed here: https://discourse.itk.org/t/how-to-pickle-simpleitk-transforms/901/12

The Transform base class is not pickleable in 2.0.0rc1:

```
>>> import SimpleITK as sitk
>>> import pickle
>>> pickle.loads(pickle.dumps(sitk.Transform()))
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/reox/.local/lib/python3.7/site-packages/SimpleITK/SimpleITK.py", line 4562, in __init__
    _SimpleITK.Transform_swiginit(self, _SimpleITK.new_Transform(*args))
TypeError: Wrong number or type of arguments for overloaded function 'new_Transform'.
  Possible C/C++ prototypes are:
    itk::simple::Transform::Transform()
    itk::simple::Transform::Transform(itk::TransformBase *)
    itk::simple::Transform::Transform(unsigned int,itk::simple::TransformEnum)
    itk::simple::Transform::Transform(itk::simple::Image &,itk::simple::TransformEnum)
    itk::simple::Transform::Transform(itk::simple::Transform const &)
```

blowekamp:
Pickling also does not work with composite transforms represented as sitk::Transform.

reox:
Additional info: It also does not work if the transformation is returned from `ImageRegistrationMethod`. In that case also `Transform` is used and not the actual type.
Workaround: After running ImageRegistrationMethod cast the Transform into the correct form.

reox:
I wrote this little test script to check which transformations work and which not:
```
import SimpleITK as sitk
import pickle
import sys

transforms = {
    'AffineTransform': (3,),
    'BSplineTransform': (3, 3),
    'DisplacementFieldTransform': (3,),
    'Euler2DTransform': (),
    'Euler3DTransform': (),
    'ScaleSkewVersor3DTransform': (),
    'ScaleTransform': (3,),
    'ScaleVersor3DTransform': (),
    'Similarity2DTransform': (),
    'Similarity3DTransform': (),
    'Transform': (),
    'TranslationTransform': (3, ),
    'VersorRigid3DTransform': (),
    'VersorTransform': ()
}

print("System Version: ", sys.version)
print('SimpleITK Version:   ', sitk.Version_VersionString())
print('          Build Date:', sitk.Version_BuildDate())
print('ITK Version:         ', sitk.Version_ITKVersionString())
[...]

PR #1121, opened 2020-05-29
Transform downcast pickle

blowekamp:
The issue report in #1112 is further addressed in this patch, with the pickling of derived Transform classes held by the the Transform base class.


blowekamp:
The restoring of a a pickled object as a different Python type than the initial object may not be the best behavior.

Additionally, more testing needs to be added with the new `Downcast` method.

zivy (approved):

Code/Common/include/sitkTransform.h
```diff
@@ -258,6 +260,15 @@ class SITKCommon_EXPORT Transform
    */
   void MakeUnique( );
 
+  /** \brief Get the TransformEnum of the underlying Transform.
+   *
+   *  A SimpleITK Transform object can internally hold any ITK transform. This
```
dave3d, line 265: 'This returns' -> 'This method returns'

Code/Common/include/sitkTransform.h
```diff
@@ -258,6 +260,15 @@ class SITKCommon_EXPORT Transform
...
+   *  returns the TransformEnum representing the internal ITK transform. This
+   *  value may be used to identify which SimpleITK class the transfom can be
```
dave3d, line 267: 'transform' (sp)

Code/Common/src/sitkTransform.cxx
```diff
@@ -578,18 +578,24 @@ void Transform::InternalInitialization(itk::TransformBase *transform)
   visitor.transform = transform;
   visitor.that = this;
 
-  // The following list must have the children before their parents to
-  // cast to the most derived classes
-  typedef typelist::MakeTypeList<itk::IdentityTransform<double, 2>,
+  // The following list must have the *parent* classed before their children
+  // since the list is traversed end to the beginning, and t
```
dave3d, line 598: Did something get cut off at the end here?

Code/Common/include/sitkTransform.h
blowekamp: done

Code/Common/include/sitkTransform.h
blowekamp: done

dave3d (approved):

PR #1125, opened 2020-06-05
Return sitkBSplineTransform for b-splines with non-default order

blowekamp:
The sitkUnknowTransform was being returned because of the VSplineOrder
was not templated in the internal GetTransformEnum method.


zivy (approved):

dave3d (approved):
Misspelled splines 'splnes' in PR title (if we care)

PR #1130, opened 2020-06-08
Remove ITKv4 code conditionally used by ITK version

blowekamp:
ITK version 5 is now required, so this code can no longer be
successfully used.


dave3d (approved):
Looks much cleaner.  Removing ifdefs is always a good thing.

zivy (approved):

PR #1132, opened 2020-06-09
Update ProcessObject with ITKv5 threading models

blowekamp:
Expose being able to set the default multi-thread, the number of work
units, and the number of threads. The behavior of these parameters
will vary deepening the threading model used.


zivy (approved):
questions on documentation.

dave3d (approved):
Good stuff

PR #1137, opened 2020-06-16
Checkin of python-lint.yml

dave3d:
Adapted from python-app.yml Github Action file.  That file sets up
an action that runs flake8 and pytest on any pull request.

For this one, the pytest call is removed, and flake8 is only run
on the Examples sub-directory.

### DIFF
--- a/.github/workflows/python-lint.yml
+++ b/.github/workflows/python-lint.yml
@@ -1,0 +1,29 @@
+name: Python Lint
+
+on:
+  push:
+    branches: [ master, release ]
+  pull_request:
+    branches: [ master, release ]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python 3.7
+      uses: actions/setup-python@v2
+      with:
+        python-version: 3.7
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip install flake8
+    - name: Lint with flake8
+      run: |
+        # stop the build if there are Python syntax errors or undefined names
+        flake8 Examples --count --select=E9,F63,F7,F82 --show-source --statistics
+        # exit-zero treats all errors as warnings. The GitHub editor is 127 chars wide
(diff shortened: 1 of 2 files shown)


.github/workflows/python-lint.yml
```diff
@@ -0,0 +1,33 @@
...
+    - uses: actions/checkout@v2
+    - name: Set up Python 3.7
+      uses: actions/setup-python@v2
+      with:
+        python-version: 3.7
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip install flake8 pytest
+        if [ -f requirements.txt ]; then pip install -r requirements.txt; fi
```
blowekamp, line 27: This line is not needed.

dave3d: Oh, right.  I remember thinking that.  The script was autogenerated by Github.

blowekamp:
Any idea why this didn't trigger the GH Action? It would be great to check it works before we merge.

Should a root .flake8 file be added for options like line length?

```
 # stop the build if there are Python syntax errors or undefined names
 flake8 Examples --count --select=E9,F63,F7,F82 --show-source --statistics
 # exit-zero treats all errors as warnings. The GitHub editor is 127 chars wide
 flake8 Examples --count --exit-zero --max-complexity=10 --max-line-length=127 --statistics
```

This "select" is from the example, any idea if we should consider more problems errors to be more strict because if there is only a warning, the CI will not be read, and it won't be check. What do you think?

.github/workflows/python-lint.yml
```diff
@@ -0,0 +1,32 @@
...
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python 3.7
+      uses: actions/setup-python@v2
+      with:
+        python-version: 3.7
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip install flake8 pytest
```
zivy, line 26: Is pytest needed here or can it be removed?

dave3d: Yeah, you're right.  Not needed.

PR #1142, opened 2020-06-30
Updated with procedural example

dave3d:


docs/source/IO.rst
```diff
@@ -50,6 +51,21 @@ Example read and write:
...
+An more compact example using SimpleITK's procedural interface:
+
+.. code-block :: python
+
+        import SimpleITK as sitk
+
+        image = sitk.ReadImage(inputImagefileName)
+        sitk.WriteImage(image, outputImagefileName)
+
+The procedural interface does not allow for a user specified IO when reading
```
zivy, line 67: It does allow specifying the imageIO:
https://simpleitk.org/doxygen/latest/html/namespaceitk_1_1simple.html#af50ad403a7aa88acbb78e5b2d1099b9c

docs/source/IO.rst
```diff
@@ -50,6 +51,21 @@ Example read and write:
         writer.SetFileName(outputImageFileName)
         writer.Execute(image)
 
+The above example specifies using the BMPImageIO to read the file.
+If that line is omitted, SimpleITK would determine which IO to use automatically,
+based on the file name's suffix and/or the file's header.
+
+An more compact example using SimpleITK's procedural interface:
```
zivy, line 58: An more -> A more

docs/source/IO.rst
blowekamp: This is a case where the SWIG wrapping does not play nicely with the overloaded methods and keyword arguments. Ideally we'd like to be able to write:

```
sitk.ReadImage(inputImageFileName, imageIO="BMPImageIO")
```

but currently the keyword arguments don't work with this method. So it must be:

```
sitk.ReadImage(inputImageFileName, sitkUnknown, "BMPImageIO")
```

dave3d: Oh my bad.  Should I put the ImageIO in the example or leave it out?

blowekamp: This is a method where we should manually make the keywords works correctly. So please create a new feature request for that.

zivy: Please add it as apparently there are only two people in the world that knew how to use it up to now, so an explicit example will help everyone.

PR #1147, opened 2020-07-02
Doxygen add dynamic time stamp

blowekamp:
Replace inline Doxygen generated time stamp with dynamic insertion
    
    The Doxygen footer macro "$generatedby" include the generate
    time. This create excessive HTML churn and useful modification time
    every time the Doxygen updated.
    
    This patch use JQuery to dynamically insert the time stamp from a
    generate "datetime.txt" file into the footer.


dave3d (approved):
Seems plausible

zivy (approved):

PR #1163, opened 2020-08-13
Add Python DiscreteGaussian and SmoothingRecursiveGaussian procedures

blowekamp:
These manual implementation replace the overload C++ methods which
prevented keyword arguments from working. This Python implementation
directly calls the object oriented API, enabling the vector or scalars
to be used for the sigma arguments.


zivy (approved):

dave3d (approved):
Nice

PR #1166, opened 2020-08-18
Image viewer tests

dave3d:


blowekamp:
ghost flow is detecting tracing white space. We should likely follow PEP8 in the tests and run the Python linter on them too.

Testing/Unit/Python/dummy_viewer.py
```diff
@@ -0,0 +1,15 @@
...
+import sys
+import os
+
+
+out = ""
+
+if len(sys.argv) > 1:
+    for x in sys.argv[1:]:
+        out = out + ' ' + x
+
```
zivy, line 12: Instead of the loop:
" ".join(sys.argv[1:])

Testing/Unit/Python/dummy_viewer.py
```diff
@@ -0,0 +1,15 @@
...
+
+print(out)
+os.environ['DUMMY_SHOW'] = x
```
zivy, line 15: The variable 'x' doesn't exist till assigned a value in the loop. If no argument is given this causes an exception. Was the intent to use the 'out' variable and not the 'x' variable?

Testing/Unit/Python/sitkExternalViewerTest.py
```diff
@@ -0,0 +1,103 @@
...
+    """Test external viewer launch"""
+
+    IMG = None
+
+    @classmethod
+    def setUpClass(cls):
+        ExternalViewerTest.IMG = sitk.GaussianSource()
+        this_dir = os.path.dirname(os.path.abspath(__file__))
+
+        cmd = 'python ' + this_dir + '/dummy_viewer.py'
```
zivy, line 37: cmd = ' python ' + os.path.join(this_dir, 'dummy_viewer.py')

Testing/Unit/Python/dummy_viewer.py
dave3d: Clearly my brain doesn't work in a Pythonic manner.

dave3d:
> ghost flow is detecting tracing white space. We should likely follow PEP8 in the tests and run the Python linter on them too.

Where is it?  I ran flake8 on my stuff before I checked it in and didn't see anything.

Testing/Unit/Python/dummy_viewer.py
zivy: No worries. I guess you're writing C++ in Python. My original R code looked like Python in R till Richard showed me the R way of doing things. Each language and its unique quirks.

PR #1171, opened 2020-08-21
Add __version__ to Python module for PEP 376 compatibility.

blowekamp:


zivy (approved):
No more `print(sitk.Version())`, now we do `print(sitk.__version__)`, nice.

PR #1175, opened 2020-08-25
Use Readme for long description in setup.py and PyPi

blowekamp:


zivy (approved):
LGTM and consistent across descriptions.

PR #1198, opened 2020-09-15
Generate and update R documentation files

blowekamp:


zivy (approved):

blowekamp:
@richardbeare The regenerated R docs contains:
```
\%C++ includes: sitkZeroFluxNeumannPadImageFilter.h
```
Which R is giving errors like:  ExtractImageFilter.Rd:38: All text must be in a section
Before the leading '\' was not there. This seems related to 404bd6efe25789468a4b8f07e55f69cf8a67eb69, do you have any insights?

Thanks.

richardbeare:
It must be - looks like escaping the % signs outside the sections is causing problems. When outside a section the % were functioning as latex comments and removing those sections. Not sure how that problem got past last time, unless perhaps the R checking has improved.

richardbeare:
I've made some changes available in https://github.com/richardbeare/SimpleITK/tree/RB_fix_rampant_escaping

The C++ include information isn't needed for R and isn't being placed in a relevant structure at moment. Best to stop it being generated. That branch is based on blowekamp:UpdateRdfiles

dave3d (approved):

blowekamp:
Closing due to age.

PR #1212, opened 2020-09-28
Update SetApplication docs

dave3d:


Code/IO/include/sitkImageViewer.h
```diff
@@ -92,7 +92,10 @@ class SITKIO_EXPORT ImageViewer
 
   /** \brief Set the full path to the viewing application used in the command string.
    *
-   * The SetApplication method expects a full path name.
+   * The SetApplication method expects a full path name to the display
+   * application's executable.  For instance, on Windows when using Fiji, one
+   * would use the full path to the ImageJ-win64.exe such as
+   * \c C:\\Users\dave\Fiji.app\ImageJ-win64.exe
```
blowekamp, line 98: I am not sure what this double back slash is. Is it an escape sequence? Please add quotes to indicate the value of this string, and you might as well add a linux example too.

blowekamp (commented):
Thanks for improving the documentation. Consider using quotes to indicate what the value of this string should be.

Code/IO/include/sitkImageViewer.h
dave3d: My bad.  I was doing a Windows path, but I don't really speak Windows.

dave3d: fixed

zivy (approved):

PR #1214, opened 2020-09-29
BUG: GetImageFromArray accepted arrays that had zero sized dimensions.

zivy:
Throw an exception when one of the numpy array sizes is zero.


blowekamp:
What is the reason for considering this behavior a bug?

The `Image` constructors accepts dimensions with size 0. The default `Image` constructor creates a zero sized image.

zivy:
Philosophical / arbitrary decision. What does a zero sized image mean? As you pointed out, in other cases we do support creation of zero sized images, so will close this PR.

blowekamp:
> Philosophical / arbitrary decision. What does a zero sized image mean? As you pointed out, in other cases we do support creation of zero sized images, so will close this PR.

It was not arbitrary. It followed from the need to define what a default constructed image is.

PR #1217, opened 2020-09-30
For Release - Use "package_data" for SimpleITK documentation files.

blowekamp:
Place the files inside the installed package directory, this is a
change from the "data_files" behavior of the installation path being
relative to the installation prefix.


blowekamp:
This patch needs work, this change is not working in the master branch.

PR #1220, opened 2020-10-03
Added example code for Java/ImageJ

haesleinhuepf:
Hi @blowekamp ,

here comes a first version of a Java example ImageJ plugin using SimpleITK. I'm happy to adapt the code according to SimpleITK guidelines if necessary. Furthermore, depending on the changes you plan on the SimpleITK Java API, I would support maintaining this plugin code - also long term. 

In the first place, I would be happy if we could remove the dependeny to `coremem`, the library I [used](https://github.com/haesleinhuepf/SimpleITK/blob/master/Examples/Java/ImageJ/src/main/java/org/simpleitk/imagej/SimpleITKUtilities.java#L42) to wrap the SimpleITK pointer in order to copy the pixels to a Java array.

Furthermore, if you know how to replace the Maven build system with CMake, I would be happy to learn that as well. :-)

Let me know what you think!

Best,
Robert


blowekamp:
This appears to be a wonderful example of how to integrate SimpleITK with ImageJ! Thank you so much for the PR. I am not sure we are easily able to bring the ImageJ SDK into the CI/dev environments to regularly test this example. I am waiting to discuss this PR with some team members to figure outs the best thing to do with it. If you have any possible suggestions they are welcomed. Perhaps a stand alone Example/repo...

blowekamp:
Regarding CMake integration, there are unfortunately few Java examples. [This](https://github.com/SimpleITK/SimpleITK/blob/master/Examples/HelloWorld/CMakeLists.txt#L20-L22) example just uses the following to compile and add the java example as a ctest:

  sitk_add_java_test( Example.HelloWorld
    "${CMAKE_CURRENT_SOURCE_DIR}/HelloWorld.java"
    )
```
Where `sitk_add_java_test` is defined here:
https://github.com/SimpleITK/SimpleITK/blob/ba6b9c3908cb8018da581b1018a0e27e29086989/CMake/sitkAddTest.cmake#L181-L234

haesleinhuepf:
> If you have any possible suggestions they are welcomed. Perhaps a stand alone Example/repo...

Sure! We could also host it under github.com/imagej/imagej-simpleitk-template or similar. I just thought it might have higher recognition as part of simple-itk between csharp and c++ examples. I leave this decision to you and the team. 

Btw there is also [this](https://github.com/clij/clijx-assistant-simpleitk) more advanced repository, which enables simple-itk support in ImageJ/Fiji/Icy/QuPath via CLIJ. That should bring better performance, because CLIJ and SimpleITK both operate on native buffers. Just saying, there's more to come 😉

PR #1216, opened 2020-09-29
Wrap Image's GetBuffer methods for Java

blowekamp:
Helps with #1215.

PR #1227, opened 2020-10-08
Add Java Image::GetGufferAsByteBuffer method

blowekamp:
Rename the `Image::GetBufferAsVoid`, to `Image::GetBufferAsByteBuffer` while construction a Java Buffer object to access the memory buffer from SimpleITK.

### DIFF
--- a/Wrapping/Java/Java.i
+++ b/Wrapping/Java/Java.i
@@ -24,12 +24,15 @@
   }
 %}
 
+%include "JavaBufferHelper.i"
 %include "JavaDoc.i"
 
 // Make Java method names follow the naming conventions
 // See the swig.swg file, and ruby.swg for details on how this works
 // Documented in: http://www.swig.org/Doc2.0/SWIG.html#SWIG_advanced_renaming
 %rename("%(firstlowercase)s", %$isfunction ) "";
+
+%rename( getBufferAsByteBuffer ) itk::simple::Image::GetBufferAsVoid;
 
 // Enable Java classes derived from Command Execute method to be
 // called from C++
--- a/Wrapping/Java/Java.i
+++ b/Wrapping/Java/Java.i
@@ -17,15 +17,7 @@
 *=========================================================================*/
 // Java
 #if SWIGJAVA
-%include "carrays.i"
-%array_class(int8_t, int8Array);
-%array_class(uint8_t, uint8Array);
-%array_class(int16_t, int16Array);
-%array_class(uint16_t, uint16Array);
-%array_class(int32_t, int32Array);
-%array_class(uint32_t, uint32Array);
-%array_class(float, floatArray);
-%array_class(double, doubleArray);
+
 %pragma(java) jniclasscode=%{
   static {
     System.loadLibrary ( "SimpleITKJava" );
--- a/Wrapping/Java/JavaBufferHelper.i
+++ b/Wrapping/Java/JavaBufferHelper.i
@@ -1,0 +1,13 @@
+/*
+ * void * typemap for returning ByteBuffer from Image::GetBufferAsVoid
+ */
+%typemap(jni) void * "jobject"
+%typemap(jtype) void * "java.nio.ByteBuffer"
+%typemap(jstype) void * "java.nio.ByteBuffer"
+%typemap(javaout) void * {
+  return $jnicall;
+}
+%typemap(out) void * {
+  const size_t size = arg1->GetNumberOfPixels()*arg1->GetNumberOfComponentsPerPixel()*arg1->GetSizeOfPixelComponent();
+  $result = JCALL2(NewDirectByteBuffer, jenv, $1, size);
+ }


blowekamp:
@haesleinhuepf In the PR direct access with a ByteBuffer object has been added as a replacement to the getBufferAsVoid method. Hopefully the methods to convert Buffer's will be more efficient. Methods such as `ByteBuffer::asFloatBuffer` can be used.

 Also we could change the return type of methods such as `getBufferBufferAsFloat` to `FloatBuffer`. However, not all C++ type have a corresponding Java Buffer type such as unsigned 16-bit integer. 

Please let me know what will provide the best interface!

blowekamp:
The behavior of the `const` versions of the getBuffer methods needs to be reviewed for the Java interface.

zivy (approved):

haesleinhuepf:
Hey @blowekamp ,

thanks for working on this!

> Also we could change the return type of methods such as `getBufferBufferAsFloat` to `FloatBuffer`.

Hm, I'm not 100%ly sure if changing the return type is a good idea. At the moment I'm [using this interface](https://github.com/clij/clijx-assistant-simpleitk/blob/master/src/main/java/net/haesleinhuepf/clijx/simpleitk/CLIJSimpleITKUtilities.java#L52) and its performance is amazing!

> Methods such as `ByteBuffer::asFloatBuffer` can be used.

this approach has been shown to not work in my context, unfortunately:
https://github.com/SimpleITK/SimpleITK/pull/1216#issuecomment-702407476

Thus, for the moment, is it possible to build the suggested new Java wrapper so that I can test it? Your suggested changes in this PR are not in the nightly built yet, correct?

Thanks for your efforts! We'll manage to make this work :-) 

Have a great weekend!

Robert

blowekamp:
Ok.

Looks like you are making some great progress using the current implementation. Hopefully abstraction you developed will continue to work until as we make the final. Great to hear you are getting good performance. 👍 

> this approach has been shown to not work in my context, unfortunately:
> [#1216 (comment)](https://github.com/SimpleITK/SimpleITK/pull/1216#issuecomment-702407476)

I missed this!

Perhaps returning a `com.sun.jna.Pointer` would be better?

haesleinhuepf:
> Perhaps returning a `com.sun.jna.Pointer` would be better?

Yes that sounds like a great idea! I'd be happy to give this a try 🙂

blowekamp:
@haesleinhuepf 
Before abandoning this approach I did a little test:

```
    PixelIDValueEnum pixelType = PixelIDValueEnum.sitkFloat32;
    long v[] = {128, 128};
    VectorUInt32 imageSize = new VectorUInt32( v );

    Image image = new Image( imageSize, pixelType );

    ByteBuffer buffer = image.getBufferAsByteBuffer();
    
    System.out.println("ByteBuffer: " + buffer.toString());
    System.out.println("asFloatBuffer: " + buffer.asFloatBuffer().toString());

    float[] array = new float[]{1, 2, 3, 4.5f, 5, 6.5f, 7, 8};
    ByteBuffer b = ByteBuffer.allocate(8 * 4);
    System.out.println("array ByteBuffer: " + b.asFloatBuffer().toString());
    System.out.println("array asFloatBuffer: " + b.asFloatBuffer().toString());
```

Which produces the following:
```

1543: ByteBuffer.toString: java.nio.DirectByteBuffer[pos=0 lim=65536 cap=65536]
1543: ByteBuffer.toString: java.nio.DirectFloatBufferS[pos=0 lim=16384 cap=16384]
1543: array asFloatBuffer: java.nio.ByteBufferAsFloatBufferB[pos=0 lim=8 cap=8]
1543: array asFloatBuffer: java.nio.ByteBufferAsFloatBufferB[pos=0 lim=8 cap=8]
```

This seems to indicate that the buffer from JNI based buffer are direct. So that should work with the GPU library you are using?

This PR just adds the `Image::getBufferAsByteBuffer` method. It can be merged, and test/evaluated before determining if we should change the return value for the existing `getBufferAs...` methods.

haesleinhuepf:
@blowekamp that looks good!

Would you mind executing the following in your code just to check if the [jocl GPU-library would cause trouble here](https://github.com/gpu/JOCL/blob/master/src/main/java/org/jocl/Pointer.java#L224) with your buffers:

    System.out.println("buffer is direct: " + buffer.isDirect());
    System.out.println("buffer is has array: " + buffer.hasArray());

    System.out.println("buffer as floatbuffer is direct: " + buffer.asFloatBuffer().isDirect());
    System.out.println("buffer as floatbuffer is has array: " + buffer.asFloatBuffer().hasArray());

Furthermore, float-array detours don't help in my particular use-case because I'm trying to avoid that any pixel is ever copied to the JVM. I'm trying to copy from ITK buffer to JOCL buffer becaue that's super fast. Also, if you can ship the dll+jar files, I'm happy to test in more detail on my Windows :-)

blowekamp:
The output of the new code is as follows:
```
1543: buffer is direct: true
1543: buffer is has array: false
1543: buffer as floatbuffer is direct: true
1543: buffer as floatbuffer is has array: false
```

So it's a direct buffer which should be hopefully as efficient as passing around as the raw point as a long ( which seems a little hacky to me ). I'm going head with merging just this methods for evaluation.

PR #1229, opened 2020-10-14
Add Image::GetSizeOfPixelComponent method

blowekamp:
Adding convenience method to get the size of pixel component.

### DIFF
--- a/Code/Common/include/sitkImage.h
+++ b/Code/Common/include/sitkImage.h
@@ -192,20 +192,27 @@ namespace simple
 
     /** \brief Get the number of components for each pixel
      *
-     * For scalar images this methods returns 1. For vector images the
-     * number of components for each pixel is returned.
+     * For images with scalar or complex pixel types this method
+     * returns one. For images with a vector pixel type the method
+     * returns the number of vector components per pixel.
      */
     unsigned int GetNumberOfComponentsPerPixel( ) const;
 
     /** \brief Get the number of pixels in the image
      *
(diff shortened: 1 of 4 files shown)


Code/Common/include/sitkImage.h
```diff
@@ -193,20 +193,27 @@ namespace simple
     /** \brief Get the number of components for each pixel
      *
      * For scalar images this methods returns 1. For vector images the
```
zivy, line 195: Possibly rephrase:
For images with scalar or complex pixel types this method returns one. For images with vector pixel type the method returns the number of vector components.

Code/Common/src/sitkImage.cxx
```diff
@@ -139,6 +140,64 @@ namespace itk
       return this->m_PimpleImage->GetNumberOfPixels();
     }
 
+  unsigned int Image::GetSizeOfPixelComponent( ) const
+    {
+    switch( this->GetPixelIDValue() )
+      {
+      case sitkUnknown:
```
zivy, line 147: Are both the sitkUnknown and default portions of the switch required? Possibly remove the `case sitkUnknown`?

zivy (approved):
Minor comments and question.

Code/Common/include/sitkImage.h
blowekamp: done

Code/Common/src/sitkImage.cxx
blowekamp: done

Issue #578, opened 2018-09-27
N4BiasFieldCorrectionImageFilter: Getting the bias field with GetLogBiasFieldControlPointLattice 


mabelzunce:
Hi,
Using the N4BiasFieldCorrectionImageFilter, I found out that the bias field can't be obtained from the object as it can be done in ITK by calling N4BiasFieldCorrectionImageFilter::GetLogBiasFieldControlPointLattice. I think this is an important feature because allows you to use a lower resolution image (via shrinking it) to estimate the bias field and then apply it to the high resolution image. Is there any particular reason why this feature was not implemented in SimpleITK?

Best,
Martin

blowekamp:
Yes it would be good to be able to get access to the bias filed to be able to correct other resolution images. I believe I was able to do this be recovering the bias field from the results. The returned image should be `Input/exp(logBiasField)=Output` so `exp(logBiasField)=Input/Output`. You can then resample the  `exp(logBiasField)` to a higher resolutions as desired.

I agree that being able to get direct access to the field is a good feature. I'll have to look into the data structures used a little closer. The `GetLogBiasFieldControlPointLattice` method returns an unstructured set of points for BSpline interpolation. This requires the itkBSplineScatteredDataPointSetToImageFilter to get an interpolated image. I think it may be better to just return the BiasFiled image. Does that meet your needs?

mabelzunce:
I've implemented this temporal solution and the bias field can't be completely recovered because the image has zero values. So, for the purpose of correcting the image is probably fine, but if the visualization of the bias field is needed is better to generate it from the control points and BSpline interpolation.

bjw34032:
I've just come across this problem for myself.  I've previously run a separate program from the shell and then read in the corrected and bias_field files.  I would prefer an in-python solution, so adding the bias field as an output for N4BiasFieldCorrectionImageFilter would be great!

I'll give the simple division a try for now...

blowekamp:
closed by #1232

PR #1232, opened 2020-10-14
Add GetLogBiasFieldAsImage method to N4 filter

blowekamp:
This method extents the ITK GetLogBiasFieldControlPointLattice method
by evaluating the bspline on a reference image to store the log bias
correction field.

Addressed #578.


zivy (approved):
Good interface. 

I initially looked at the `BSplineLatticeToImage` function and thought it was problematic because the user provided the spline order and lattice, but then noticed that this is only for internal use and the N4BiasFieldCorrectionImageFilter only exposes the `referenceImage` parameter.

Issue #1233, opened 2020-10-15
Update the AZP SimpleITK.Package pipeline to use the same build and checkout hashes.


blowekamp:
For the SimpleITK.Package pipeline, there is a separate git hash used for the checkout of the CI code and the checkout of the SimpleITK build code. The current undesired behavior when a tag is made, the tag is used to the CI code and *master* is used for the SimpleITK hash that is compiled.

Update the behavior so that when a tag is made the *tag* hash is the code that is built.

The default AZP variables will need to be updated along with the "Configure" step in the Package pipeline.

PR #1234, opened 2020-10-15
Update the AZP SimpleITK.Package pipeline to use the same build and checkout hashes

jwj019:
This PR address https://github.com/SimpleITK/SimpleITK/issues/1233.  Submitting this PR for initial testing purposes.


blowekamp:
Looks like a good starts.

However, this change would make it so that the CI branch/hash and the build branch/hash are the same, and no longer use the AZP queue variables. Please see:

https://github.com/SimpleITK/SimpleITK/blob/master/Testing/CI/Azure/azure-pipelines-package.yml#L21

That line is where the logic can be updated to use the AZP variable to the triggering/CI hash/branch in stead of "release".

jwj019:
@blowekamp This PR is finally ready to go. If the buildhash variable is master, it uses whatever the build source branch is.  Otherwise, it uses the buildhash variable.

PR #1238, opened 2020-10-20
Update Java's Buffer interface with GetBufferAsBuffer and GetBufferAsNativePointer methods

blowekamp:
The previous developmental version of GetBufferAsType have been remove and replaced by these two methods which return a type derived from java.nio.Buffer or a long type for a native pointer.


blowekamp:
@haesleinhuepf Per our discussions, I believe this is the interface that we converged at. Please take a look.

haesleinhuepf:
@blowekamp 
That sounds great! Is there by chance a dll and a jar I can test? This is not part of the nightly build yet, right?

Thanks!

blowekamp:
@haesleinhuepf Sorry for the massive delay on this forgotten PR. Lets get this merged into master so that the binaries can be tested before 2.1rc1.

Wrapping/Java/Java.i
```diff
@@ -24,16 +24,91 @@
...
+        {
+          return b.asFloatBuffer();
+        }
+      if (getPixelID() == PixelIDValueEnum.sitkFloat64 ||
+          getPixelID() == PixelIDValueEnum.sitkComplexFloat64 ||
+          getPixelID() == PixelIDValueEnum.sitkVectorFloat64)
+        {
+          return b.asDoubleBuffer();
+        }
+       // sitkInt8 and sitkVectorInt8 are returned as ByteBuffer too.
```
zivy, line 106: Is this comment also relevant for sitkUInt8, sitkVectorUInt8? I's assume so, but not sure.

zivy (approved):

Wrapping/Java/Java.i
blowekamp: Yes. This is an inline comment, and even the doxygen like comment for the method will not be exposed in the Java docs. We probably need to explicitly document the mapping for all types in an Example.

dave3d (approved):

PR #1240, opened 2020-10-21
unpack was moved to table.unpack for Lua 5.2

dave3d:


zivy (approved):
Cool overloading.

blowekamp (changes requested):
Currently we require and are using Lua 5.3. It seems to work fine What was the error you encountered? Perhaps a more verbose commit message would better explain the situation.

blowekamp:
Hrmm... Let me see if I have figured this out some more.

We require Lua 5.3 for the code generation. And we provide that in this super build.

This PR is for testing of the Lua *wrapping*. During the configuration of SimpleITK this is usually the system Lua. We don't have such a version requirement for the wrapped version of Lua.

I have been testing with Lua 5.2 on Ubuntu 18.04 LTS, which must be compile with Lua compatibility modes enabled. I am uncertain with versions of Lua we are using in the CI.

blowekamp (changes requested):
This PR now include changes to the Wrapping/Java directory that appear to be mistakenly committed.

dave3d:
> This PR now include changes to the Wrapping/Java directory that appear to be mistakenly committed.

Oops.  I've fixed it with the latest update.

blowekamp (approved):

PR #1247, opened 2020-11-02
Fixed old,non-working Java HelloWorld example

dave3d:


Examples/HelloWorld/HelloWorld.java
```diff
@@ -24,30 +24,24 @@ public static void main(String argv[]) {
 
     // Create an image
     PixelIDValueEnum pixelType = PixelIDValueEnum.sitkUInt8;
-    VectorUInt32 imageSize = new VectorUInt32( 2 );
-    imageSize.set( 0, 128 );
-    imageSize.set( 1, 128 );
+    long[] isize = {128, 128};
```
zivy, line 27: Possibly remove the array definitions: `VectorUInt32 imageSize = new VectorUInt32( new long[] {128, 128} );`

Not sure which syntax is better/more common in java as the arrays defined in the code are just used in the constructors.

zivy (approved):
Stylistic comment which I'm not sure about.

Examples/HelloWorld/HelloWorld.java
dave3d: I tend to do it that way because that's the way my brain works.  And in the olden days you had to declare things beforehand.  Also, I find long function calls crammed with lots of stuff unreadable.  But that could just be me.

zivy: I don't have a specific preference, it's just a question what is the common java'ish style, if there is one, so that people feel the code is "good java". At least in Python and R there are very specific styles.

PR #1249, opened 2020-11-03
added SITK_NOSHOW check

dave3d:


zivy (approved):
Initially didn't get why the image title included the language, but makes sense if you are running multiple HelloWorld programs and want to see which is which.

PR #1260, opened 2020-11-17
Add missing SWIG mapping for int 64 and void pointer types

blowekamp:
The CSharp Image methods GetBufferAsVoid, GetBufferAsUInt64 and
GetBufferAsInt64 return type was SWIGTYPE_p_void, etc. This
appropriate type map is not added to return these pointers as IntPtr
types.


dzenanz (approved):

blowekamp:
The PR addresses issues reporting in ITK Discourse: https://discourse.itk.org/t/how-can-i-use-getbufferasint64-in-c/3623/4

zivy (approved):

PR #1269, opened 2020-12-01
Fix AZP Linux packaging to build configure.BuildHash

blowekamp:
The SimpleITK repository hash used be built was incorrectly set to the
same path/checkout as the CI scripts. The docker run command was
mounting the CI SimpleITK source directory overriding the
SIMPLEITK_GIT_TAG env varible. This mount is removed.


blowekamp:
The linux packages in this build are from the wrong version:
https://dev.azure.com/SimpleITK-DevOps/SimpleITK/_build/results?buildId=3881&view=results
This PR should correct it.

jwj019 (approved):
LGTM 👍

PR #1299, opened 2021-01-28
Use python venv

blowekamp:


Wrapping/Python/LegacyPackaging.cmake
```diff
@@ -90,28 +82,27 @@ if (SimpleITK_PYTHON_USE_VIRTUALENV)
...
     DEPENDS
     "${SWIG_MODULE_SimpleITKPython_TARGET_NAME}"
-    "${CMAKE_CURRENT_BINARY_DIR}/PythonVirtualEnvInstall.cmake"
-    COMMENT "Creating python virtual enviroment..."
+    COMMENT "Creating python virtual environment..."
     )
+
+  add_custom_command( TARGET PythonVirtualEnv
+    POST_BUILD
+    COMMAND "${VIRTUAL_PYTHON_EXECUTABLE}" -m pip install numpy
```
blowekamp, line 104: The prior implementation manually copied numpy from the system python, if available. Here we use pip to likely download it. There is probably a better place to put out "testing" requirements.

zivy (approved):

Wrapping/Python/LegacyPackaging.cmake
blowekamp: This "hard coding" of the dependencies can be addressed in another PR as a future improvement.

PR #1302, opened 2021-01-29
Discontinue packaging Python 3.5

blowekamp:
Python 3.5 reached end of life with the final 3.5.10 release in
September 2020.


zivy (approved):
Out with the old 👍 .

PR #1306, opened 2021-02-02
Restrict automatic found Python versions to the supported versions

blowekamp:
This enables Python 3 to automatically be found when version 2 is in
the path too.


zivy (approved):

dave3d (approved):

blowekamp:
The patch has been superseded by the update to the new CMake FindPython module.

PR #1308, opened 2021-02-03
Update to support CMake Ruby module prefix

blowekamp:
CMake 3.15, change the recommended prefix for the Ruby find package
module to match the case of the package name. The CMake code is update
to support the undated naming. This addresses configuration error
related to this name change in SimpleITK Superbuild.

The usage of the Ruby libraries is update to prefer to use the new
names.


dave3d (approved):
Looks good

zivy (approved):

PR #1310, opened 2021-02-03
Update CircleCI images to buster

blowekamp:
Update to python 3.6 version too.


zivy (approved):
CircleCI, still faster than GitHub actions. Then again, shouldn't complain about free services.

dave3d (approved):

PR #1314, opened 2021-02-07
Updating to modern Python from 2.7 to 3.x.

zivy:
Removed imports from future and replaced calls to format() with
fstring usage.


dave3d (approved):
Yet other Python print format?

blowekamp:
Thanks for making these changes. Lets see if CircleCI build OK before merging...

PR #1318, opened 2021-02-09
Set the default ITK Multithreader to Platform in Superbuild

blowekamp:
The Pool multithreader has issues when the process is forked, which is
commonly done in Python multiprocessing. Additionally, the Platform
multithreader has better progress reporting than the Pool threader.


zivy (approved):
Back to the future (I'm a fan).

dave3d (approved):

PR #1322, opened 2021-02-12
Update io docs

dave3d:
Move the example code in the doc to Examples/SimpleIO.  Put it in a Python tab.  Will add other language tabs.


blowekamp:
This PR looks like is need a rebase as it includes merges and other commits.

zivy (approved):
So many languages 🎉

blowekamp (approved):
Looks great!

PR #1331, opened 2021-02-24
Add python image meta data get item

blowekamp:


zivy (approved):
So Pythonic :scream: 

Gone are the good old days when only those who could compile ITK from source could use it. I guess that's progress.

dave3d (approved):
Sexy!

PR #1351, opened 2021-03-17
For Python Image class add support for ellipsis (...) indexing

### LINKED ISSUE #1319: Support ellipsis object for slicing images in python
Hi there,
SimpleITK already has basic support for slicing in Python: Slice objects are recognized and when asking subsets of n-dimensional it works fine. Hoewever, the ellipsis object (the infamous ` ...`) is not supported and returns a runtime error in `__getitem__`.

Numpy has very extensive support for this operator (see [here](https://github.com/numpy/numpy/blob/master/numpy/lib/index_tricks.py) for some examples). Implementing all of this might be a bit overkill for an image-processing library, but supporting ellipsis at the beginning/end of an index would be helpful.

Example code using some n-dimensional MRI:

```python
# Baseline
> image.GetSize()
(352, 352, 60, 1, 17)
# Subscripting from front-to-back works fine
> image[3].GetSize()
(352, 60, 1, 17)
# Requested behavior: Using Ellipsis does not work now.
> image[..., 12].GetSize()
IndexError: invalid index
# Work-around to get intended behavior
> image[(slice(None),)*4+(3,)].GetSize()
(352, 352, 60, 1)
# Using numpy works fine but is undesirable
# Note that dims are swapped around when getting numpy array
> sitk.GetArrayViewFromImage(image)[3,...].shape
(1, 60, 352, 352)
```

blowekamp:
Closes #1319

### DIFF
--- a/Testing/Unit/Python/sitkImageIndexingTest.py
+++ b/Testing/Unit/Python/sitkImageIndexingTest.py
@@ -113,6 +113,10 @@ class TestImageIndexingInterface(unittest.TestCase):
         # check empty image
         self.assertImageNDArrayEquals(img[-1:0,-1:0], nda[-1:0,-1:0])
 
+        self.assertImageNDArrayEquals(img[...], nda[...])
+        self.assertImageNDArrayEquals(img[...,1:3], nda[1:3,...])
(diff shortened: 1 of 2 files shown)


zivy (approved):
LGTM

dave3d (approved):
So cool.

PR #1356, opened 2021-03-23
Correct the name of the ImageIO used in the example.

blowekamp:


dave3d (approved):
Oops, my bad.

dave3d:
While you're at it, down at the Transform section I refer to the Python code.  But now I have all the langs.  So you can fix that.

PR #1357, opened 2021-03-23
Fixed C++ type assertion

dave3d:


blowekamp (approved):
Thanks for the follow up

PR #1391, opened 2021-05-20
Remove filename from sha512 has file

blowekamp:


zivy (approved):

dave3d (approved):
Oh, is that what I did wrong?  Thanks for figuring it out.

PR #1392, opened 2021-05-21
removed { in Cpp examples

dave3d:


blowekamp (commented):
Thanks for the quick fix.

blowekamp (approved):

Issue #1409, opened 2021-06-16
Document details of metric sampling strategies


blowekamp:
In the doxygen member/enum documentation in the ImageRegistrationMethod, details about the different sampling strategies should written. This should include that even with the None/Regular strategies, the sample point are randomly perturbed.

zivy:
This requires a bit of a longer "discussion" so not 100% sure if doxygen is the place, or should it go on the read-the-docs registration page. So rtd or doxygen?

blowekamp:
Both? There is nothing in the doxygen documentation. A sentence or so about each would be useful in Doxygen. Extended prose could be added to RTD too.

PR #1410, opened 2021-06-16
DOC: Improve documentation for registration sampling strategies.

zivy:
Add documentation describing the three available sampling strategies
and why the REGULAR and RANDOM strategies yield different results for
repeated registration runs, due to the use of a stochastic component
for selecting the sampled points.

Resolves: #1409

### DIFF
--- a/Code/Registration/include/sitkImageRegistrationMethod.h
+++ b/Code/Registration/include/sitkImageRegistrationMethod.h
@@ -495,6 +495,15 @@ namespace simple
     /** \brief Get the percentage of pixels used for metric evaluation.  */
     const std::vector<double> &GetMetricSamplingPercentagePerLevel() const;
 
+    /** \brief Sampling strategies for obtaining points.
+     *
+     * - NONE: use all voxels, sampled points are the voxel centers.
+     * - REGULAR: sample image voxels using a regular grid, then within
+     *            each voxel randomly perturb from center.
+     * - RANDOM: sample image voxels with replacement using a uniform
+     *           distribution, then within each voxel randomly perturb
+     *           from center.
+     */
     enum MetricSamplingStrategyType {
       NONE,
       REGULAR,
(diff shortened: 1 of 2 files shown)


Code/Registration/include/sitkImageRegistrationMethod.h
```diff
@@ -495,6 +495,15 @@ namespace simple
     /** \brief Get the percentage of pixels used for metric evaluation.  */
     const std::vector<double> &GetMetricSamplingPercentagePerLevel() const;
 
+    /** \brief Sampling strategies for obtaining points.
+     *
+     * - NONE: use all voxels, sampled points are the voxel centers.
+     * - REGULAR: sample image voxels using a regular grid, then within
+     *            each voxel randomly perturb from center.
+     * - RANDOM: sample image voxels with replacement using a uniform
```
dave3d, line 503: what does with replacement mean?  probably just me being dumb.

docs/source/registrationOverview.rst
```diff
@@ -189,6 +189,26 @@ the registration runtime. The ImageRegistration method allows you to
...
+registration code multiple times will yield different results. To remove
+the randomness, set the pseudo-random number generator's seed in the
+`SetMetricSamplingPercentage <https://simpleitk.org/doxygen/latest/html/classitk_1_1simple_1_1ImageRegistrationMethod.html#a8b891c62404a8dc5010241fea619c932>`_
+method to a constant. The default seed value is
+wall clock time.
+
+Note that using the RANDOM sampling strategy with a 100% sampling rate is not
+equivalent to using the sampling strategy of NONE. Given an image with N voxels,
+the former randomly selects N voxels with repetition and perturbs the points
+within each voxel, the latter uses the centers of all N voxels. Thus, for
```
dave3d, line 208: ^latter^later^

dave3d (approved):

docs/source/registrationOverview.rst
zivy: "latter" is the correct spelling in this case, at least according to the [Merriam Webster dictionary](https://www.merriam-webster.com/dictionary/the%20latter#:~:text=1%20%3A%20the%20second%20one%20of,I%20like%20the%20latter%20best.)

Code/Registration/include/sitkImageRegistrationMethod.h
zivy: The phrase is using the statistics nomenclature, we are performing a simple random sampling with replacement. Uniform probability for selecting each voxel and we don't remove the voxel once it is selected so "with replacement". This means the same voxel can be sampled multiple times (very small probability, but still greater than zero). If we used sampling without replacement the probability of selecting the same voxel multiple times is zero.

blowekamp:
I don't believe the description for the "REGULAR" sampling strategy as a "grid" sampling is accurate. The ITK code for the sampling is here: https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/Registration/RegistrationMethodsv4/include/itkImageRegistrationMethodv4.hxx#L992-L1021
This implementation selects every n-th pixel while traversing in scan-line order. When n is near the scan-line length it can produce undesirable..

Issue #1135, opened 2020-06-12
Exception not thrown in R


zivy:
Expected the following code to throw an exception. 
```
library(SimpleITK)
anonymous_transform_type <- Transform(TranslationTransform(2,c(1.0,0.0)))
anonymous_transform_type$GetOffset()
```
Instead get a message "Error: node stack overflow"

richardbeare:
This error usually the result of the method (GetOffset), not being present. Either caused by typo or wrapping bug.

zivy:
I'm curious why this error is different than calling a non existent function. When I do the following I can catch the exception and do something.
```
tryCatch(nonExistant(),error = function(err) {cat("The function doesn't exist.")})
```
Is this a feature of the wrapping, or is this a bug in the wrapping? This happens for other functions, I just never tried to call a non existent method up to now.

richardbeare:
Something to do with the way swig creates the class hierarchy. It effectively implements its own OO style, in which it includes lists of methods. Those get searched by the swig generated system, rather than the usual base R system. Perhaps an exception would be better.

My long term, and possibly overly hopeful, plan is to have swig generate wrapping using R6 classes, which are an internal class structure.

blowekamp:
@zivy What is the status of this issue? Can we close it? Is there some specific fix needed?

richardbeare:
I've never had a chance to look at this. I'm working on the windows build again, so may get an opportunity.

richardbeare:
I just had a quick look. The error is different for me - no more informative, but catchable:

```r
> anonymous_transform_type <- Transform(TranslationTransform(2,c(1.0,0.0)))
> anonymous_transform_type$GetOffset()
Error: C stack usage  7969364 is too close to the limit
```
this is an exception:
```r
tryCatch(anonymous_transform_type$GetOffset(),error = function(err) {cat("The function doesn't exist.")})
The function doesn't exist.
```
This may be a wrapping bug in that we're not correctly setting up the R class hierarchy.

I'll have a look into it.

richardbeare:
It seems the virtual stuff isn't the root cause. It is just that there was no allowance built into the base classes that swig creates to deal with this sort of error. We need a matching $ method in the base class to catch anything that comes through and therefore fail more gracefully. If I add the following manually:

```r
setMethod('$', "ExternalReference",
function(x, name) {
errormessage <- paste0(name, ": method doesn't exist")
stop(errormessage)
}
)
```

The following happens:

```r
[...]

PR #1411, opened 2021-06-17
BUG: Calls to nonexistent methods weren't handled gracefully

richardbeare:
Added a '$' method to the base class from which everything in
SimpleITK descends that prints a meaningful error while
raising an exception.

Eventually this piece of code should be generated by SWIG.

There's a detection section that will warn us if this happens

A simple test is included.

See #1135


zivy (approved):
LGTM, tested on my local machine and issue was resolved.

PR #1412, opened 2021-06-17
DOC: Additional information on registration sampling strategies.

zivy:


docs/source/registrationOverview.rst
```diff
@@ -209,6 +211,11 @@ within each voxel, the latter uses the centers of all N voxels. Thus, for
 repeated random sampling with 100% rate, you get different samples and likely
 none of them is of the centers of all N voxels.
 
+Combining a mask with sampling is done using a rejection approach. First a sample is generated and
+then it is accepted or rejected if it is inside or outside the mask. This may cause
+problems when the mask region occupies a very small region in the original image. In such cases it is
+better to use a cropped version of the image for registration, possibly the mask's bounding box.
+
```
gdevenyi, line 218: Can you clarify if this sampling continues until enough samples are selected, or does it just discard random samples that don't fall in the mask?

zivy: It just discards points that are outside the mask (https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/Registration/RegistrationMethodsv4/include/itkImageRegistrationMethodv4.hxx#L992-L1055). This is why a small mask inside a large image is problematic. For mutual information we end up with not enough samples for estimation which kills the registration. Some of the other metrics are slightly less sensitive to lack of samples, but they all suffer from it.

Using rejection based sampling, if the code iterated till enough samples were obtained we are in danger of creating an infinite loop (small mask inside very large image).  Not sure why the original developers used a rejection based scheme vs. iterating over the mask, but that is what was implemented.

Do you think it would be beneficial if we add this to the documentation? If yes, I'll add.

gdevenyi: I think being explicit about this pitfall is well worth documenting.

zivy: Done.

docs/source/registrationOverview.rst
```diff
@@ -209,6 +211,17 @@ within each voxel, the latter uses the centers of all N voxels. Thus, for
 repeated random sampling with 100% rate, you get different samples and likely
```
blowekamp, line 211: Avoid second person. "you get different sample"-> "different samples are produced"

docs/source/registrationOverview.rst
```diff
@@ -209,6 +211,17 @@ within each voxel, the latter uses the centers of all N voxels. Thus, for
...
 none of them is of the centers of all N voxels.
 
+Combining a mask with sampling is done using a rejection approach. First a sample is generated and
+then it is accepted or rejected if it is inside or outside the mask. This may cause
+problems when the mask region occupies a very small region in the original image. Because the sampling
+only discards data, the actual sample size is reduced. For some similarity metrics (e.g. mutual information)
```
blowekamp, line 217: "the actual sample size is reduced" -> "the sample rate within the mask can be reduced from the requested"

docs/source/registrationOverview.rst
zivy: done.

docs/source/registrationOverview.rst
zivy: done

blowekamp (approved):
Thanks for the improvement!

PR #1438, opened 2021-08-09
Add max dimension support to "projection" filters.

### LINKED ISSUE #1431: Feature Request: Wrap *ProjectionImageFilter with 4D to allow usage over multiple images
There are a number of *ProjectionImageFilter in ITK which could be used to produce various types of averages/projections across a number of images. In order for this to work however, they need to have a 4D implementation, so that the 3D images can be stacked and collapsed.

I'd like to request this wrapping be added.

blowekamp:
Closes #1431


blowekamp:
Just noting the existing 2, 3D maintain the same number of dimensions but reduce the projected dimension's size to 1. This is continued with the proposed 4,5D implementations.

zivy (approved):

PR #1450, opened 2021-09-09
Update ITK superbuild version to v5.3.rc01

blowekamp:


blowekamp:
@zivy There is a change in behavior with the number of reported iterations with the LBFGS2 optimizer. Do think it warrants further investigation?

zivy (approved):

zivy:
I peeked at the [last modification of LBFGS2Optimizerv4](https://github.com/InsightSoftwareConsortium/ITK/commit/c34fb54e10ad4cfbb44d0507e2bb693bf9467259) and the superclass was changed, which likely accounts for the change in reporting number of iterations. I think it is worth looking into this a bit more, just to make sure that the change only effected number of iterations reporting and not something more significant. Should this be an additional test in ITK, or in SimpleITK?

PR #1465, opened 2021-09-29
GitHub actions cpp only

ViktorvdValk:


blowekamp:
Looks great! Have you tested if the window builds work now?

Also please make a separate PR to the ITKIntegration code change.

ViktorvdValk:
The windows build still gives OOM issues.

blowekamp:
I see the message "space left on device" which sounds like a disk issue and not a system memory.

The first thing to try is to change the build type to "MinSizeRel".

If that doesn't work we will have to delete some object files during the build process. For example build and install the ITK target, then delete it's object files before building SimpleITK.

blowekamp:
I guess the next step is to try something like:
```
ninja ITK
rm -rf ITK-build
ninja
```

ViktorvdValk:
This is the log from the GitHub Action build.
Cmake decides to keep building a debug build even tough MinSizeRel is specified.
it seems to consider a MinSizeRel build, but then decides otherwise.

[16/44] Performing install step for 'ITK'
[0/1] Install the project...
-- Install configuration: "MinSizeRel"
[0/1] Install the project...
-- Install configuration: "Debug"

PR #1471, opened 2021-10-11
Fix configuration error when Python wrapping is enabled

lassoan:
3D Slicer often failed to build due to SimpleITK configuration error (mostly on Windows, when building in debug mode, see https://github.com/Slicer/Slicer/issues/5498) with this or similar messages:

```
  Performing configure step for 'SimpleITK'
  loading initial cache file C:/D/S4D/SimpleITK-prefix/tmp/SimpleITK-cache-Debug.cmake
  -- Selecting Windows SDK version 10.0.19041.0 to target Windows 10.0.22000.
  CMake Error at C:/Program Files/CMake/share/cmake-3.21/Modules/FindPackageHandleStandardArgs.cmake:230 (message):
    Could NOT find Python3 (missing: Python3_EXECUTABLE Python3_LIBRARIES
    Python3_INCLUDE_DIRS Interpreter Development.Module) (Required is at least
    version "3.6")
    C:/Program Files/CMake/share/cmake-3.21/Modules/FindPackageHandleStandardArgs.cmake:594 (_FPHSA_FAILURE_MESSAGE)
    C:/Program Files/CMake/share/cmake-3.21/Modules/FindPython/Support.cmake:3166 (find_package_handle_standard_args)
    C:/Program Files/CMake/share/cmake-3.21/Modules/FindPython3.cmake:485 (include)
    C:/D/S4D/VTK-build/lib/cmake/vtk-9.0/VTK-vtk-module-find-packages.cmake:303 (find_package)
    C:/D/S4D/VTK-build/lib/cmake/vtk-9.0/vtk-config.cmake:144 (include)
    C:/D/S4D/VTK-build/vtk-config.cmake:1 (include)
    C:/D/S4D/ITK-build/lib/cmake/ITK-5.2/Modules/ITKVtkGlue.cmake:30 (find_package)
    C:/D/S4D/ITK/CMake/ITKModuleAPI.cmake:76 (include)
    C:/D/S4D/ITK/CMake/ITKModuleAPI.cmake:31 (itk_module_load)
    C:/D/S4D/ITK/CMake/ITKModuleAPI.cmake:129 (_itk_module_config_recurse)
    C:/D/S4D/ITK-build/ITKConfig.cmake:80 (itk_module_config)
    SuperBuild.cmake:368 (find_package)
    CMakeLists.txt:39 (include)
  -- Configuring incomplete, errors occurred!

```

The problem is that Slicer builds its Python interpreter that FindPython3.cmake cannot find, because it is not installed in a standard locations, because it does not get the Python3_EXECUTABLE hint.
Slicer's build system sets Python3_EXECUTABLE in the top-level SimpleITK project, but the variable is not passed on to the internal SimpleITK-build project.

The problem is fixed by passing on Python3_EXECUTABLE to the internal SimpleITK-build project (similarly to how SWIG_EXECUTABLE is passed).


MarkaPola:
@blowekamp The best way to define some `HINT` is to use `Python3_ROOT_DIR`, if possible, because it is used preferably for all artifacts, regardless of which components are required: defining `Python3_EXECUTABLE` won't help if only `Development` components are searched.

Now, regarding your problems with `FindPython3` module, it is hard to diagnose any potential issue without any more information.

mathstuf:
VTK will only require dependencies of modules that are requested. Due to backwards compat, requesting nothing means "request everything". So adding a `COMPONENTS` list to the `find_package(VTK)` will make any unnecessary external searches not fail the process (they may still make configure noise, but `VTK_FOUND` is what ultimately matters).

blowekamp:
@lassoan Do you have Slicer_BUILD_ITKPython enabled for this build? That might trigger a dependency on Python for VTK ( and I'd think it'd do the same to ITK too? ).

If not it would be interesting to look at the following file:   C:/D/S4D/ITK-build/lib/cmake/ITK-5.2/Modules/ITKVtkGlue.cmake

However in SimpleITK this error occur a the following point:
```
find_package(ITK REQUIRED )
#we require certain packages be turned on in ITK
include(sitkCheckForITKModuleDependencies)
```

SimpleITK here does not a a list of components. It does not "use" or include the ITK components here. The intent is to just get a list of all ITK modules, and ITK configuration information and variables. 

 SimpleITK does't need or use the ITK/VTK modules. This error occurs during this find_package of a transient dependence in ITK's configuration. Latter on SimpleITK uses ITK like the following:
```

set(use_itk_modules ITKCommon ITKImageCompose ITKImageIntensity
  ITKLabelMap ITKTransform ITKIOTransformBase ITKDisplacementField )
find_package(ITK COMPONENTS ${use_itk_modules}  REQUIRED)

add_library ( SimpleITKCommon ${SimpleITKCommonSource} ${SimpleITKAncillarySource} )
sitk_target_use_itk (SimpleITKCommon PRIVATE ${use_itk_modules} )
```

Currently I'm testing if the initial find_package for ITK could work with just the ITKCommon component. However, I don't have the full Slicer build read to test this.

blowekamp:
Please see  #1477 for my proposed alternative.

blowekamp:
Thanks for the PR, but we have gone with this solution in #1477.

jcfr:
To follow up @mathstuf just publish a VTK merge requests (see [MR-8554](https://gitlab.kitware.com/vtk/vtk/-/merge_requests/8554)) that should address the root problem and address VTK issue [#18328](https://gitlab.kitware.com/vtk/vtk/-/issues/18328) :pray: :100:

blowekamp:
@jcfr Great work getting to the root of the problem!

PR #1498, opened 2021-11-01
Merge Lint and Spell workflows

dave3d:
python-lint.yml and python-spell.yml have been combined into one
new workflow, PythonLintAndSpell.yml.


blowekamp (commented):
L1: Update the name of this work flow
L10: The name of the job is "build", please update too

dave3d:
> L1: Update the name of this work flow
> L10: The name of the job is "build", please update too

Oops.  Should I have two separate jobs, lint and spell?

blowekamp:
> > L1: Update the name of this work flow
> > L10: The name of the job is "build", please update too
> 
> Oops. Should I have two separate jobs, lint and spell?

It will take a little more time to install the dependencies, but I think the workflow into tow jobs makes sense and should allow developers to more quickly determine the source of a CI failure.

zivy (commented):
Not too sure about these changes. Previously linting and spell checking were separate workflows which ran independently. With this change they are two jobs in a single workflow. Not sure if this change makes sense, unless the intent is to make spell checking run only if linting succeeds (add a `needs: lint` to the spell job).

dave3d:
They really are separate things, but Brad asked that I combine them.  Note that the linting is checking the formatting of the Python code, while the spell checking is done on comments the C++ header files.  It just happens that the spell checking is written in Python.

blowekamp:
@zivy With the addition of Doxygen and the Build workflows, there are starting to be to many workflows in the GHA GUI. We can kinda of consider the spell checking a type of code linting, so they can be considered the same category of task. While they are independent, placing them in the same workflow is for more of a grouping the the GUI.

zivy (approved):

zivy:
@blowekamp, thanks for explaining, sounds reasonable, approved PR.

dave3d:
https://github.com/SimpleITK/SimpleITK/issues/1497

PR #1502, opened 2021-11-03
Add to Image methods to interpolate pixels at a location

### LINKED ISSUE #654: resample a single point or point set
Add a resample method to the Image class, as suggested in [this discourse discussion](https://discourse.itk.org/t/how-can-i-get-an-interpolated-value-between-voxels/1535).

Proposed interfaces:
```
std::vector< double > GetIntensityAtContinuousIndex(const std::vector &, InterpolatorEnum)
std::vector< double > GetIntensityAtPhysicalPoint(const std::vector &, InterpolatorEnum)
```
Where the input vector is coordinates of a single/multiple points and the return vector is the intensity(s)? Always return double as the interpolation is really carried out using real values. Maybe allow the user to specify the output type?

blowekamp:
closes #654


Code/Common/include/sitkImage.h
```diff
@@ -254,6 +255,32 @@ namespace simple
...
+    /** \brief Interpolate pixel value at a continuous index.
+     *
+     * This method is not supported for Label pixel types.
+     *
+     * The valid range of continuous index is [-0.5, size-0.5] for each dimension. An exception is thrown if index is out of bounds.
+     *
+     * @param index The continuous index must be at least the length of the image dimension.
+     * @param interp The interpolation type to use, only sitkNearest and sitkLinear are supported for Vector and Complex pixel types.
+     *
+     * @return All supported pixel types are returned as an array, where complex number are returned with the real followed by the complex component.
```
zivy, line 267: number -> numbers

Code/Common/include/sitkImage.h
```diff
@@ -254,6 +255,32 @@ namespace simple
...
+     */
+    std::vector<double> EvaluateAtContinuousIndex( const std::vector<double> &index, InterpolatorEnum interp = sitkLinear) const;
+
+    /** Interpolate pixel value at a physical point.
+     *
+     * This method is not supported for Label pixel types.
+     *
+     * An exception is thrown if the point is out of the defined region for the image.
+     *
+     * @param point The physical point which is transformed to a continuous index for interpolation.
```
zivy, line 277: Possibly change to "The physical point at which the interpolation is computed."
That the physical point is converted to a continuous index is an implementation detail.

Code/Common/src/sitkCreateInterpolator.hxx
```diff
@@ -0,0 +1,269 @@
...
+  case sitkBlackmanWindowedSinc: {
+    using WindowFunction =
+        typename itk::Function::BlackmanWindowFunction<WindowingRadius, double,
+                                                       double>;
+    using InterpolatorType = itk::WindowedSincInterpolateImageFunction<
+        TImageType, WindowingRadius, WindowFunction, BoundaryCondition>;
+    return RType(ConditionalCreateInterpolator<InterpolatorType>(
+        typename IsBasic<TImageType>::Type()));
+  }
+  case sitkBSplineResamplerOrder1: {
```
zivy, line 211: Not sure if this should support resampling based on the itk::BSplineResampleImageFunction because it is specific to BSpline coefficient images and should not be used for regular images. On the other hand this is consistent with the ResampleImageFilter.

Code/Common/src/sitkCreateInterpolator.hxx
```diff
@@ -0,0 +1,269 @@
...
+    using InterpolatorType =
+        itk::NearestNeighborInterpolateImageFunction<TImageType, double>;
+    return RType(InterpolatorType::New());
+  }
+  case sitkLinear: {
+    using InterpolatorType =
+        itk::LinearInterpolateImageFunction<TImageType, double>;
+    return RType(InterpolatorType::New());
+  }
+  case sitkBSpline: {
```
zivy, line 130: Explicitly set BSpline order to 3 or add documentation stating that this is the default order.

zivy (approved):

Code/Common/src/sitkCreateInterpolator.hxx
blowekamp: That is documented in sitkInterpoaltor.h.

PR #1517, opened 2021-11-18
N4 bias example port

dave3d:
Ported the N4 bias example to C++ and CSharp.

### DIFF
--- a/Examples/N4BiasFieldCorrection/CMakeLists.txt
+++ b/Examples/N4BiasFieldCorrection/CMakeLists.txt
@@ -3,6 +3,22 @@ if(NOT BUILD_TESTING)
   return()
 endif()
 
+add_executable ( N4BiasFieldCorrection N4BiasFieldCorrection.cxx )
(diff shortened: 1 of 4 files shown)


zivy (approved):
To harmonize all examples, please remove [the call to sitk.Show](https://github.com/SimpleITK/SimpleITK/blob/6378d74836c76c58fdf0ba33d8178885fede07d1/Examples/N4BiasFieldCorrection/N4BiasFieldCorrection.py#L66) in the Python example (or add it to all others)?

dave3d:
> To harmonize all examples, please remove [the call to sitk.Show](https://github.com/SimpleITK/SimpleITK/blob/6378d74836c76c58fdf0ba33d8178885fede07d1/Examples/N4BiasFieldCorrection/N4BiasFieldCorrection.py#L66) in the Python example (or add it to all others)?

Oh right.  I think I'll added it to the others, since I'm basing off the Python.

PR #1522, opened 2021-11-22
Adding wrapping for TransformGeometryImageFilter

blowekamp:
The test was initially generated from the code generation then
modified. The code generation only handles inputs of Image types.


Testing/Unit/sitkTransformGeometryImageFilterTest.cxx
```diff
@@ -0,0 +1,144 @@
...
+
+  filter.DebugOn();
+  ASSERT_NO_THROW(output = filter.Execute(input1, input2));
+
+  // Check that the input is not modified
+  // This will fail if the filter is being run in-place
+  EXPECT_EQ(inputSHA1hash, itk::simple::Hash(input1))
+      << "Input was modified by filter.";
+
+  // Check that the input buffer is different that the output
```
zivy, line 115: different than

zivy (approved):

Testing/Unit/sitkTransformGeometryImageFilterTest.cxx
blowekamp: Done 😁

PR #1523, opened 2021-11-22
AZP Update CI to use windows 2019 and VS 16 2019

blowekamp:
AZP has deprecated window 2016:
https://github.com/actions/virtual-environments/issues/4312


dave3d (approved):
Cool

PR #1533, opened 2021-11-26
Add Python interface for EvaluateAt methods

blowekamp:
The Python methods change scalar type to be scalars (float), and
complex to be complex type, while the vector pixel types remain as
tuples.

### DIFF
--- a/Testing/Unit/Python/sitkImageTests.py
+++ b/Testing/Unit/Python/sitkImageTests.py
@@ -327,6 +327,66 @@ class ImageTests(unittest.TestCase):
         image %= 4
         self.assertEqual(image[1, 1], 2)
 
+    def test_evaluate_at_continuous_index(self):
+        """Test for EvaluateAtContinuousIndex"""
+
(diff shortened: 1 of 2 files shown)


Testing/Unit/Python/sitkImageTests.py
```diff
@@ -327,6 +327,66 @@ def test_inplace_operators_constants(self):
         image %= 4
         self.assertEqual(image[1, 1], 2)
 
+    def test_evaluate_at_continuous_index(self):
+        """Test for EvaluateAtContinuousIndex"""
+
+        for pxiel_type in (sitk.sitkFloat32, sitk.sitkFloat64,
```
zivy, line 333: consistent misspelling in the method, so not sure if it should be corrected "pxiel_type" -> "pixel_type"

zivy (approved):

Testing/Unit/Python/sitkImageTests.py
blowekamp: fixed

PR #1544, opened 2021-12-14
Fix unknown enum for smoothing DisplacementFieldTransforms

blowekamp:
Explicit implmentation of the templated dispatched GetTransformEnum
method is needed for the derived transforms. The generic templated
pointer had priority over casting the pointer to a parent class.


blowekamp:
addressed failures in #1539

zivy (approved):

PR #1558, opened 2022-01-07
WIP: Enable ITK FFT Factory registration

blowekamp:
ITK 5.3 is adding FFT factories. This needs special configuration to
ensure the generated FFT factory code is generated and registered for
the appropriate SimpleITK targets.

The difficulty is that ITK's  UseITK.cmake file set all configuration
properties at a global and directory level. This changes the
configurations for all SimpleITK libraries built in the BasicFilters
directory adding unnecessary include directories, flags, and other
compilation options to targets.

### DIFF
--- a/Code/BasicFilters/src/CMakeLists.txt
+++ b/Code/BasicFilters/src/CMakeLists.txt
@@ -169,3 +169,34 @@ add_filter_library(SimpleITKBasicFilters1 SimpleITKBasicFilters1Source no_itk_de
 
 target_link_libraries ( SimpleITKBasicFilters1
   PRIVATE ${PREV_SimpleITK_LIBRARIES} )
+
+
+
+find_package(ITK COMPONENTS ITKFFT )
+set(ITK_NO_IO_FACTORY_REGISTER_MANAGER 1)
+
+# the ITK CMake file sets directory level properties for the factories
+# that need to bedisabled and only set for the SimpleITK_ITKFFT
+# target.
+# Note: The ITK USE file also sets other directory level properties
+# such include dirs for all the FFT dependencies, and link_dirs and
+# maybe others that are not desired for all the libraries.
+include(${ITK_USE_FILE})
+
+get_property(_dir_compile_definitions DIRECTORY PROPERTY COMPILE_DEFINITIONS)
+list( REMOVE_ITEM _dir_compile_definitions
+  ITK_FFT_FACTORY_REGISTER_MANAGER )
+set_property( DIRECTORY PROPERTY COMPILE_DEFINITIONS ${_dir_compile_definitions})
+
+get_property(_dir_include_directories DIRECTORY PROPERTY INCLUDE_DIRECTORIES)
+list( REMOVE_ITEM ${_dir_include_directories}
+  "${CMAKE_CURRENT_BINARY_DIRECTORY}/ITKFactoryRegistration" )
+set_property( DIRECTORY PROPERTY INCLUDE_DIRECTORIES ${_dir_include_directories})
+
+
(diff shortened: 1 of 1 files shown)


blowekamp:
There is a struggle here trying to use the FFT factories on a target level and not on the whole directory.
@dzenanz @tbirdso

Any suggestions?

dzenanz:
I haven't done a deep dive in ITK build system in months, probably longer. Tom's mind should be fresh with FFT stuff, so he should be able to help.

tbirdso:
I was about to say that @dzenanz may know better than me, uhoh 😅 

The heart of the issue here is less about FFT filters and more about object factory infrastructure, the idea behind recent changes and improvements is that the object factory will better support swappable backends across different modules going forward. `UseITK.cmake` defines only a handful of new compiler definitions for factories and includes only a few additional configuration files that are necessary for registering default factories at runtime.

There has been discussion around refactoring the ITK build structure to rely on CMake targets rather than directories, but that will require a significant effort and I'm not up to date on whether those plans are in the pipeline beyond the idea phase. Perhaps @thewtex would be able to better comment on this discussion?

dzenanz:
> refactoring the ITK build structure to rely on CMake targets rather than directories

This is still planned. Not sure whether we are targeting 5.4 or, more likely, 6.0.

blowekamp:
I have been using this: https://github.com/SimpleITK/SimpleITK/blob/master/CMake/sitkTargetUseITK.cmake
```
# target_use_itk
#
# This function has the following form:
#   sitk_target_use_itk(<target>
#                       <PRIVATE|PUBLIC|INTERFACE>
#                       [itk_module [itk_module [...]]])
#
#   It configures the target, with the require include directories and
# link libraries to use the list of itk modules provided. The keyword
# to indicate interface of the library is also required.

function(sitk_target_use_itk target_name interface_keyword)

```

Which has been working very well up until this point. I'll see what I can figure out.

> the idea behind recent changes and improvements is that the object factory will better support swappable backends across different modules 

Does this mean a compile time configuration/options for modules or runtime options? I can certainly see a place for both.

blowekamp:
One more thing, I am interested in getting the FFT classes being instantiated in one library and just used at runtime by other modules, and not requiring library and compilation unit to recompile the FFT classes over and over again, and require the additional include and dependencies every where.

blowekamp:
@dzenanz @tbirdso I think I am going to work on a little refactoring in ITK:
 - move the factory generation code to a cmake function in the ITK configuration file ( this will enable its usage with out using ITK_USE_FILE
 -  split ITK_NO_IO_FACTORY_REGISTER_MANAGER into all the named factory types

tbirdso:
> >the idea behind recent changes and improvements is that the object factory will better support swappable backends across different modules

> Does this mean a compile time configuration/options for modules or runtime options? I can certainly see a place for both.

Both, where new factories can be configured at compile time with the `FACTORY_NAMES` field in `itk-module.cmake` to be registered at runtime init by default, or manually added to the factory list at runtime for overriding backends.

> I think I am going to work on a little refactoring in ITK

Thank you for taking a look at this!

> move the factory generation code to a cmake function in the ITK configuration file ( this will enable its usage with out using `ITK_USE_FILE`)

This sounds good to me, I assume this will reduce your need for including `UseITK.cmake` in SimpleITK?

> split ITK_NO_IO_FACTORY_REGISTER_MANAGER into all the named factory types

By "types" do you mean the existing set of factory manager files (ImageIO, MeshIO, TransformIO, FFT)? I'm not sure exactly how this should be accomplished but if you have an idea I'm happy to give feedback on a PR.

blowekamp:
Alternative implementation was merged in #1559

PR #1559, opened 2022-01-08
Use itk_generate_factory_registration and update to ITK 5.3rc03

blowekamp:
ITK 5.3 adds FFT factories which need special configuration to
ensure the generated FFT factory code is generated and registered for
the appropriate SimpleITK targets.

Use new itk_generate_factory_registration CMake function to generate
private registration methods, which find grain control of factories.


blowekamp:
Dependent on InsightSoftwareConsortium/ITK#3059

zivy (approved):

dave3d (approved):

Issue #1563, opened 2022-01-13
registration framework: return same concrete transform type as input


zivy:
In RC1 of v2.2.0, the registration framework will return the same transformation type as set in the `SetInitialTransform` if the `inPlace` flag is True. If the flag is False, the returned type is `CompositeTransform`. 

Ideally, the returned transformation type would be the same as the initial type irrespective of the `inPlace` flag value.

blowekamp:
@zivy Do you have a minimal reproducible example of this?

zivy:
MWE below. When `inPlace=False` returned type is CompositeTransform, when set to `True` it is Euler3DTransform.

```
import SimpleITK as sitk

fixed_image = sitk.Image([4,4,4], sitk.sitkFloat32)
moving_image = sitk.Image([4,4,4], sitk.sitkFloat32)
transform = sitk.Euler3DTransform()

registration_method = sitk.ImageRegistrationMethod()
registration_method.SetMetricAsCorrelation()
registration_method.SetMetricSamplingStrategy(registration_method.NONE)
registration_method.SetOptimizerAsGradientDescent(learningRate=1.0, 
                                                  numberOfIterations=100, 
                                                  convergenceMinimumValue=1e-6, 
                                                  convergenceWindowSize=10)
registration_method.SetInitialTransform(transform, inPlace=True)
print(type(registration_method.Execute(fixed_image, moving_image)))
```

blowekamp:
The default is with `inPlace=true` for wrapped languages with the [SetInitialTransfrom](https://simpleitk.org/doxygen/latest/html/classitk_1_1simple_1_1ImageRegistrationMethod.html#a6d2a1b1dd30453ca0d9b1c670b6bfd29). So the returning of the Composite, maybe commonly not used by default, and also may be frequently avoided to get the correct transform returned. It has been my intention in many of the registrations I have written to avoid the composite results.

Unfortunately, I do believe it would be better to be returning the correct transform ( not encapsulated in a Composite ), but it may be a disruptive change. However, I don't recall any user questions regarding interactive the a Composite returned transform type.

blowekamp:
The API change may be reconsidered for SimpleITK 3.0.

PR #1565, opened 2022-01-14
WIP: Return an sitk::Transform with ITK specific pimple object

blowekamp:
The returned object from ImageRegistrationMethod::Execute when the
InitialTransform had inPlace=False, was an sitk::Transform holding a
itk::CompositeTransform pimple object. The composite transform then
held an optimized transform of the initial type.

While the return type for the method remains a sitk::Transform, the
object held in the pimple transform is now the optimized transform of
the initial type.


blowekamp:
Addresses #1563

blowekamp:
No longer relevant a composite transform is not longer returned.

PR #1568, opened 2022-01-19
Add type annotation into Python wrapper

orsinium:
1. Add type annotation to pure Python function in the Python wrapper.
2. Support [pathlib.Path](https://docs.python.org/3/library/pathlib.html) as the type of the `fileName` argument.
3. Add [py.typed](https://www.python.org/dev/peps/pep-0561/#packaging-type-information) into the wrapper. I'm not sure if it's properly packaged, though, it requires to rebuild the whole SimpleITK to test which is quite demanding for resources. So I'm more guessing there.
4. Improve arguments for Resample function: make `referenceImage` and `size` explicit arguments instead of hiding it inside `**kwargs`.

### DIFF
--- a/Wrapping/Python/CMakeLists.txt
+++ b/Wrapping/Python/CMakeLists.txt
@@ -94,7 +94,7 @@ configure_file(
   "${CMAKE_CURRENT_SOURCE_DIR}/SimpleITK/_version.py.in"
   "${CMAKE_CURRENT_BINARY_DIR}/SimpleITK/_version.py" )
 
-set(SimpleITK_Py_Files "__init__.py" "extra.py" )
+set(SimpleITK_Py_Files "__init__.py" "extra.py" "py.typed" )
 
 if(DEFINED SKBUILD)
   # Currently this installation
--- a/Wrapping/Python/Packaging/setup.py.in
+++ b/Wrapping/Python/Packaging/setup.py.in
@@ -52,7 +52,7 @@ setup(
     author_email = 'insight-users@itk.org',
     ext_modules=[Extension('SimpleITK._SimpleITK', [r'SimpleITK/@SimpleITK_BINARY_MODULE@'])],
     packages= ['SimpleITK'],
-    package_data = {"SimpleITK": doc_files},
+    package_data = {"SimpleITK": doc_files + ["py.typed"]},
     platforms = [],
     description = r'SimpleITK is a simplified interface to the Insight Toolkit (ITK) for image registration and segmentation',
     long_description = long_description,
--- a/Wrapping/Python/SimpleITK/extra.py
+++ b/Wrapping/Python/SimpleITK/extra.py
@@ -16,11 +16,15 @@
 #
 # ========================================================================
 
+from pathlib import Path
 from SimpleITK.SimpleITK import *
 from SimpleITK.SimpleITK import _GetMemoryViewFromImage
 from SimpleITK.SimpleITK import _SetImageFromArray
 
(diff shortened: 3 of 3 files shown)


orsinium:
`py.typed` must be at the top level of the package, see [PEP-561](https://www.python.org/dev/peps/pep-0561/#packaging-type-information). In practice, a missed `py.typed` means that the package will be skipped by mypy and pycharm. It's ok that some functions in the module aren't annotated, `py.typed` will indicate that some already are, so the users already can benefit from the type checker when using them.

Feel free to ask any questions. Type annotations are relatively new in Python, many devs are still unfamiliar with it. Still, it already gets a wide community adoption, at least at the libraries' side. It's hard to overstate how beneficial it is for code inspection in IDEs.

The backstory is that I had noticed that there are a few functions in SimpleITK that have already been annotated and decided to contribute the rest. We actively use SimpleITK in the team and we currently have wrappers that make it a bit more Pythonic. Something like this:

```python
def read_image(path: Path) -> sitk.Image:
    return sitk.ReadImage(str(path))
```

blowekamp:
Ok, thank you for the information. I have been using type annotations in several projects for a while. I was just unaware of the effect of `py.typed`. Is there some type of test that could be unto verify the effect of `py.typed` with just the core python modules?


> The backstory is that I had noticed that there are a few functions in SimpleITK that have already been annotated and decided to contribute the rest. We actively use SimpleITK in the team and we currently have wrappers that make it a bit more Pythonic. Something like this:
> 
> ```python
> def read_image(path: Path) -> sitk.Image:
>     return sitk.ReadImage(str(path))
> ```

There has been some discussion about adding a submodule/package to SimpleITK which contains more pythonic procedure. Likely with the CamelCases methods aliased to snake_case. The hope is that it could be done with some type of introspection into the current SimpleITK module to generate the procedures for most of the cases ( as opposed to more code generation). Any thought, contribution, or prototypes to this would be welcomed.

orsinium:
> Is there some type of test that could be unto verify the effect of py.typed with just the core python modules?

Great question! I curate a list of typing-related tools and libraries which includes a section on testing:
https://github.com/typeddjango/awesome-python-typing#testing

The first tool in the list [mypy-test](https://github.com/orsinium-labs/mypy-test), is what I developed, use, and maintain. These tools are used to check that type annotations work as expected for the user. 

However, I see it fit for typing-heavy libraries. In case of SimpleITK, I think just having a few examples and running `mypy` on these examples will do. Namely, it should fail on missed `py.typed`, unless [follow_imports](https://mypy.readthedocs.io/en/stable/running_mypy.html#follow-imports) is disabled.

> Any thought, contribution, or prototypes to this would be welcomed.

I'd be glad to contribute. Is there a way to recompile the Python wrapper without recompiling the whole SimpleITK? When I follow the official build instructions, each time takes more than an hour with `-j 4` (and if I use more cores, it just kills my PC).

dave3d:
> I'd be glad to contribute. Is there a way to recompile the Python wrapper without recompiling the whole SimpleITK? When I follow the official build instructions, each time takes more than an hour with `-j 4` (and if I use more cores, it just kills my PC).

If you only make changes to the Python stuff, you shouldn't need to recompile the core SimpleITK C++ code.  So it should go much quicker than a full build.

blowekamp:
These changes look very good.

The last commit is a fix for the prior commit, could you please squash the fix so reduce noise in the history before we merge?

orsinium:
done

blowekamp (approved):
Thank you for this contribution!

PR #1572, opened 2022-01-25
Add typelist2 with variadic templates

blowekamp:
A replacement implementation of typelists and utilities implemented
with variadic templates, and pack expansion. This new implementation
is expected to generate smaller symbols and be more efficient for
compilers compared to the recursive implementation TypeList with Head
and Tail types.


blowekamp:
@N-Dekker You may find this PR interesting. It replaces a older reclusive implementation of [typelist](https://github.com/SimpleITK/SimpleITK/pull/1572/files#diff-188c3a3919c0fb37d110b3e724263615342d14c4c11b12ca8a2fc5afb854b477) with a variadic implements. Only the `index_of` utility has recursive implementation. Also the `has_type` has a non-reclusive implementation with `max(initializer list{is_same<tn,t>,...})` but that resulted in a MSVC syntax error.

If you have time your thought or comments would be appreciated.

Code/Common/include/Ancillary/type_list2.h
```diff
@@ -0,0 +1,255 @@
...
+struct visit;
+template <typename... Ts>
+struct visit<typelist<Ts...>>
+{
+
+  template <typename T, typename Predicate>
+  static int
+  _f(Predicate && visitor)
+  {
+    visitor.CLANG_TEMPLATE operator()<T>();
```
N-Dekker, line 190: Is this `.CLANG_TEMPLATE` workaround still necessary nowadays?

Code/Common/include/Ancillary/type_list2.h
```diff
@@ -0,0 +1,255 @@
...
+namespace typelist2
+{
+
+/** \struct typelist
+ * \brief an non-instantiable holder of variadic template parameters to represent a list of types
+ *
+ * Can be used as a not and not an object. It enables complex compile-time manipulation of sets of types.
+ */
+template <typename... Ts>
+struct typelist;
```
N-Dekker, line 34: I find it a bit difficult to distinguish `typelist2` and `typelist`. Maybe follow the code convention of having namespace name in lower case, and template name in CamelCase? Like `typelist2::TypeList`?

Code/Common/include/Ancillary/type_list2.h
```diff
@@ -0,0 +1,255 @@
...
+
+/** \brief Provides access to the number of types in a typelist.
+ *
+ * Uses std::integral_constant to define member type "value_type" and member constant "value".
+ */
+template <typename Typelist>
```
N-Dekker, line 40: I think the ITK code convention would be to have an extra `T` in front of names of template arguments, like `TTypelist`, or `TTypeList`. But it's up to you, of course.

Code/Common/include/Ancillary/type_list2.h
```diff
@@ -0,0 +1,255 @@
...
+template <typename... Ts>
+struct visit<typelist<Ts...>>
+{
+
+  template <typename T, typename Predicate>
+  static int
+  _f(Predicate && visitor)
+  {
+    visitor.CLANG_TEMPLATE operator()<T>();
+    return 0;
```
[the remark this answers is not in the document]
blowekamp: To be used with an initializer_list it can't return void. Know of an alternative?

Code/Common/include/Ancillary/type_list2.h
blowekamp: Yea, the naming is a bit horrible. `TypeList` was the name of the old class. It was easier to search and change to a different name. I think it will be each to change back to TypeList now.

Code/Common/include/Ancillary/type_list2.h
blowekamp: Good suggestion.

Code/Common/include/Ancillary/type_list2.h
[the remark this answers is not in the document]
N-Dekker: > Know of an alternative?

Still trying to find a better alternative myself 🙂 

I think I would define an empty `EmptyStruct` (whose size is typically just one byte). Then the variadic pack could be expanded by passing it to the C-style array of those structs:

    struct EmptyStruct {};
    using EmptyStructArray = EmptyStruct[];
    (void)EmptyStructArray{ (_f<Ts>(visitor), EmptyStruct{})... };

Which would avoid the dependency from `std::initializer_list`. `_f` could then simply return `void`. And `EmptyStruct` makes the intention slightly more clear. Well... only _slightly_  ;-) 

But your approach will also work fine, I guess.

N-Dekker:
Very interesting. 😃  Did you already have a look at https://github.com/facebook/folly/blob/main/folly/detail/TypeList.h ? Might be interesting to see if there's any relevant difference.

zivy (approved):

PR #1587, opened 2022-02-03
BUG: variable name was wrong and misleading.

zivy:
Introduced in previous version of the N4 example, "improving" variable
names was wrong. This time it is the correct name.


dave3d (approved):

blowekamp (commented):
looks good

PR #1588, opened 2022-02-07
Updating documentation to v.2.2.0 functionality.

zivy:
Starting with v2.2.0 ReadTransform in Python automatically downcasts,
so the return type is the basic type and not the generic Transform().


dave3d (approved):
LGTM

PR #1603, opened 2022-02-23
Add to AZP Packaging CSharp build for OSX

blowekamp:


blowekamp:
@vincenzoml The file name from the MacOS packaging is `SimpleITK-2.2.0rc2.post23-CSharp-unknown-anycpu.zip`. The platform name and architecture will need to be fixed. But this seems like a good start.

https://dev.azure.com/SimpleITK-DevOps/SimpleITK/_build/results?buildId=5082&view=results

vincenzoml:
Thanks a lot! This is great news to me. Additionally: from last time I re-compiled simpleitk for osx (shamefully in voxlogica it's still at 1.0) I recall I had to rename the ".so" file to ".dylib" and maybe to add a prefix to the file name. When a binary is ready to be tested I will report more precisely on whether that is still needed. Also AFAIK the .dll file ought to be interchangeable between platforms. Don't know if it will be bit-by-bit equal though.

blowekamp:
The first step is to get the current CSharp package/dist being built on the CI for the basic platforms. From there hopefully you can contribute to improving and standardizing the packages.

dave3d (approved):

vincenzoml:
I am following this  but did not understand how I can help right now :) Sorry I'm a bit slow ... please do not hesitate to just give orders to me.

blowekamp:
@vincenzoml On the nightly [latest](https://github.com/SimpleITK/SimpleITK/releases/tag/latest) releases page there is no the [SimpleITK-2.2.0rc2.post32-CSharp-macosx-10.9-anycpu.zip](https://github.com/SimpleITK/SimpleITK/releases/download/latest/SimpleITK-2.2.0rc2.post32-CSharp-macosx-10.9-anycpu.zip) package. Can you please download and test. Also the naming convention for the file could be improved. Do you have any suggestions?

vincenzoml:
@blowekamp thanks a lot. I've downloaded the binary, but as anticipated I had to rename it from ".dylib" to ".so". Once this has been done I have a typecast problem that I have to investigated. Would it be easier for you to provide also a linux binary to test, I think the three platforms should be tested together. I also need to find out whether the .dll is the same or it is equivalent nevertheless. I think so.

blowekamp:
@vincenzoml Thanks for testing. Can you please create an new issues regarding the Mac extension issue. Also please provide information on how to reproduce it, as the SimpleITK CSharp tests seem to pass with the current extension.

The Linux CSharp packaging is being worked on in #1605.

PR #1608, opened 2022-02-25
Update ITK version to post ITKv5.3rc3 with fix for SLIC filter

blowekamp:
Update SLIC tests baseline images for fix in SLIC filter in the
initialization of the clusters.


dave3d (approved):
Looks good. The tests all pass on M1 and Linux aarch64.

PR #1610, opened 2022-03-01
DOC: Adding example of using Python logging with SimpleITK.

zivy:
ITK issues warnings for several situations which are relevant for
debugging. These warnings are written to stderr or stdout. This is
primarily relevant for registration where the warning that not enough
points are available for estimating the similarity metric is
issued. This example illustrates how to re-route the file descriptors
so that the warnings are written to a log file.

### DIFF
--- a/Examples/PythonFileLogging/CMakeLists.txt
+++ b/Examples/PythonFileLogging/CMakeLists.txt
@@ -1,0 +1,9 @@
+
+if ( NOT BUILD_TESTING )
+  return ()
+endif ()
+
+
+sitk_add_python_test( Example.PythonFileLogging
+  "${CMAKE_CURRENT_SOURCE_DIR}/FileLogging.py"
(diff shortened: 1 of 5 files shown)


blowekamp:
We already have an example for integrating ITK logging and Python logging here:
https://github.com/SimpleITK/SimpleITK/blob/master/Examples/Logging/Logging.py

Having two would be confusing.

If there are messages that are occurring in ITK directly to stdout/stderr perhaps ITK should be fixed to use it's own logging system that we currently support.

blowekamp:
There is also a pure Python logging module solution to the issue of redirecting stdout/stderr to a Python logging object.
https://stackoverflow.com/questions/19425736/how-to-redirect-stdout-and-stderr-to-logger-in-python
This more a Python logging solution and at SimpleITK one.

PR #1613, opened 2022-03-17
bugfix (typo in sitkUint8)

m-reuter:
If passing a mask, this code failed due to the typo.


dave3d (approved):
Thanks for finding this bug.

PR #1615, opened 2022-03-19
fix for unused variable warning

dave3d:


zivy (commented):
Several requests after looking at the code and read-the-docs:

1. Add the C# example to the RTD pages (currently only Python and R tabs are visible).
2. Modify all examples so that they write the full res image, `corrected_image_full_resolution`, using the argv[2] parameter instead of the resized image which is done in the current code.
3. If the program received a shrink factor, argv[3], write the shrunk image, `corrected_image`, to disk too using a constant name.

This would address the warning and the code would make more logical sense.

dave3d:
> Several requests after looking at the code and read-the-docs:
> 
> 1. Add the C# example to the RTD pages (currently only Python and R tabs are visible).
> 2. Modify all examples so that they write the full res image, `corrected_image_full_resolution`, using the argv[2] parameter instead of the resized image which is done in the current code.
> 3. If the program received a shrink factor, argv[3], write the shrunk image, `corrected_image`, to disk too using a constant name.
> 
> This would address the C# warning and the code would make more logical sense.

That makes sense.  I'll make the changes.

I didn't know the code, so I wasn't sure why the resized image was written by default.

zivy (approved):
Minor observation, possibly use underscore in the "shrunk" file names instead of dot? The dot separator seems strange and I don't know what windows does with it in terms of display of file extension, maybe it does the right thing.

dave3d:
> Minor observation, possibly use underscore in the "shrunk" file names instead of dot? The dot separator seems strange and I don't know what windows does with it in terms of display of file extension, maybe it does the right thing.

Looking at the other testing output files on Windows, they tend to use dash, '-'.  So I'll go with that.  I don't think Windows really cares, though, since some do have  dots, and the look ok in Windows.

PR #1629, opened 2022-04-19
DOC: Adding an entry to the FAQ.

zivy:
Explanation about image appearing blank/black/empty because of
visualization and not actual data.


blowekamp (approved):
Looks good. Possibly add a reference to the Mask common convention: https://simpleitk.readthedocs.io/en/master/conventions.html#mask-image-types-and-default-values

dave3d (approved):

zivy:
Added reference to mask images.

blowekamp:
Thanks.

PR #1638, opened 2022-05-05
disallow slashes and non-ascii character in temp file name

dave3d:
I found a bug where '/' was allowed in the temporary file name.  That messes up the writing of the temporary file.

### DIFF
--- a/Testing/Unit/Python/sitkExternalViewerTest.py
+++ b/Testing/Unit/Python/sitkExternalViewerTest.py
@@ -70,7 +70,8 @@ class ExternalViewerTest(unittest.TestCase):
 
         try:
             viewer = sitk.ImageViewer()
-            viewer.SetTitle("Basic Image Viewer Test")
+            # Some funky characters in the title to test the temp file name filter
+            viewer.SetTitle("Basic Image Viewer Test/:\'")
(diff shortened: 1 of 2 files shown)


dave3d:
> Please add some tests. Also would '' be a problem on windows? or ':'?

will do.  yeah I can exclude those characters too.

dave3d:
Ping @blowekamp @zhangknih

zivy (approved):

Testing/Unit/Python/sitkExternalViewerTest.py
```diff
@@ -70,7 +70,8 @@ def test_image_viewer(self):
 
         try:
             viewer = sitk.ImageViewer()
-            viewer.SetTitle("Basic Image Viewer Test")
+            # Some funky characters in the title to test the temp file name filter
+            viewer.SetTitle("Basic Image Viewer Test/:\'")
```
blowekamp, line 74: Is there a GetTitle method that could be used to verify the operation completed as expected? Perhaps testing a couple cases?

dave3d: It doesn't filter the title string.  It only filters the temporary file name string, which we don't provide access to.

Issue #1636, opened 2022-05-02
ImageSeriesReader reading images incorrectly


six6egrees:
Hi,

I am using SimpleITK 2.1.1.1 build py37h82bb817_0. I am trying to read a DICOM series with python containing a binary mask from a segmentation. ImageSeriesReader returns an empty image with only zeroes. The image can be correctly opened with 3D Slicer or ImageJ. Using ImageFileReader on a single slice, the resulting image is better, although instead of zeros and ones I get zeros and 0.9999.... which is also surprising, given that the RescaleSlope of the file is 1. Here is the code I use:

```
import SimpleITK as sitk
import os 

test_path = '.\\SE5'

reader = sitk.ImageSeriesReader()  
dicom_names = reader.GetGDCMSeriesFileNames(test_path)
reader.SetFileNames(dicom_names)
image = reader.Execute()

reader2 = sitk.ImageFileReader()
reader2.SetImageIO('GDCMImageIO')
reader2.SetFileName(os.path.join(test_path, '0000006K'))
image2 = reader2.Execute()

filt = sitk.MinimumMaximumImageFilter()
filt.Execute(image)  
print(filt.GetMaximum()) #0.0

filt.Execute(image2)
print(filt.GetMaximum()) #0.9999999975098001
```

Here are the relevant DICOM attributes of the image (I have omitted attributes unrelated to pixel data):

[0020,0032]	ImagePositionPatient -348.632812, -348.632812, -872.330017	
[0020,0037]	ImageOrientationPatient	 1, 0, 0, 0, 1, 0	
[0028,0002]	SamplesPerPixel	1	
[0028,0004]	PhotometricInterpretation	MONOCHROME2	
[0028,0010]	Rows	256	
[0028,0011]	Columns	256	
[0028,0030]	PixelSpacing 2.734375, 2.734375	
[0028,0100]	BitsAllocated	16	
[0028,0101]	BitsStored	16	
[0028,0102]	HighBit	15	
[0028,0103]	PixelRepresentation	1	
[0028,1050]	WindowCenter	0.5	
[0028,1051]	WindowWidth	1	
[0028,1052]	RescaleIntercept	0	
[0028,1053]	RescaleSlope	1	
[0054,0081]	NumberOfSlices	259

zivy:
Hello @six6egrees,

[...]

PR #1640, opened 2022-05-09
BUG: Test for reading a DICOM series with variable rescale slope.

zivy:
The output pixel type of the image is determined by the rescale slope
type. If only looking at a single image from a series, the first
image,  the type may be too restrictive (unsigned int) when other
images have a float. The image data is then truncated (e.g. 4.5 is
mapped to 4).


blowekamp:
@zivy Thank for making a PR to reproduce the issue. What there a reported issue # associated with this?

What it the plan for the fix or the proposed work around for users?

zivy:
This is with respect to issue #1636, see our discussion there.

PR #1643, opened 2022-05-20
make IsBadCharacter func static

dave3d:


N-Dekker:
@dave3d Just wondering, is there a specific compilation or linker problem that could be solved by the `static` keyword that you're proposing here?

In general, I think it should be sufficient to place such a local helper function inside an unnamed (anonymous) namespace. Right?

As suggested by C++ Core Guidelines "Use an unnamed (anonymous) namespace for all internal/non-exported entities", https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rs-unnamed2 

It seems to me that `IsBadCharacter` is already inside an unnamed namespace.

PR #1652, opened 2022-06-21
DOC: Updating the documentation for a corner case.

### LINKED ISSUE #1636: ImageSeriesReader reading images incorrectly

zivy:
When reading a series of images, if the caller does not specify
the pixel type for the returned image, it is deduced from the
first image in the series. If this results in a narrowing
conversion, the returned image is not valid. The solution is
to have the user explicitly specify the wider pixel type. We do
not modify the approach because it would significantly increase the
computational cost while the use case is very rare.

closes #1636

### DIFF
--- a/Code/IO/include/sitkImageSeriesReader.h
+++ b/Code/IO/include/sitkImageSeriesReader.h
@@ -40,6 +40,17 @@ namespace itk {
      * Once the image series is read the meta-data is directly
      * accessible from the reader.
      *
+     * \note If the pixel type for the returned image is not specified
+     * it is deduced from the first image in the series. This approach
+     * is computationally efficient and assumes that all images in a
+     * series have the same pixel type. In rare situations this is not
+     * the case, not all images have the same pixel type.
+     * If this leads to a narrowing conversion (e.g. first image pixel
+     * type is unsigned int and others are float) the returned
+     * image does not represent the data correctly. To resolve such
+     * situations, explicitly specify the expected pixel type via the
+     * SetOutputPixelType method before reading the series.
+     *
      * \sa itk::simple::ReadImage for the procedural interface
      **/
     class SITKIO_EXPORT ImageSeriesReader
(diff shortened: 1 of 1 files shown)


Code/IO/include/sitkImageSeriesReader.h
```diff
@@ -40,6 +40,17 @@ namespace itk {
      * Once the image series is read the meta-data is directly
      * accessible from the reader.
      *
+     * \note If the pixel type for the returned image is not specified
```
blowekamp, line 43: technically the pixel type is implicitly specified as "sitkUnknown", to indicate that is should be deduced.

blowekamp (approved):
A little lengthly and specific for this particular issue. It is likely there would be another issue a user could encounter when it would be required to specify the output pixel type too.

PR #1655, opened 2022-06-28
Fix examples for flake8

dave3d:
In looking at the SimpleElastrix PR, I saw a bunch of Flake8 warnings.  This PR addresses most of those.

The only warnings I didn't fix have to do with comment lines like these:

https://github.com/SimpleITK/SimpleITK/blob/a450bdd0425fb3e5d50eac70e0dfb4fbc530046e/Examples/FilterProgressReporting/FilterProgressReporting.py#L41

Flake8 does not like the '##' that starts the line. It says comments should have a space after the first hash.

Is that line some sort of directive that needs to stay like that?


dave3d:
> Did you try running black?

I haven't.  I can give it a try.

blowekamp (approved):
Looks good! Thank for improving the style.

PR #1663, opened 2022-07-19
STYLE: Use `unique_ptr` for Elastix and Transformix Pimpl data members

N-Dekker:
Following C++ Core Guidelines, April 10, 2022, "A raw pointer (a T*) is non-owning":
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#r3-a-raw-pointer-a-t-is-non-owning

### DIFF
--- a/Code/ElastixTransformixWrappers/include/sitkTransformixImageFilter.h
+++ b/Code/ElastixTransformixWrappers/include/sitkTransformixImageFilter.h
@@ -23,6 +23,7 @@
 #include "sitkImage.h"
 
 #include <map>
+#include <memory> // For unique_ptr.
 #include <string>
 #include <vector>
 
@@ -122,7 +123,7 @@ public:
 private:
 
   class TransformixImageFilterImpl;
-  TransformixImageFilterImpl* m_Pimple;
+  const std::unique_ptr<TransformixImageFilterImpl> m_Pimple;
 
 };
 
--- a/Code/ElastixTransformixWrappers/include/sitkElastixImageFilter.h
+++ b/Code/ElastixTransformixWrappers/include/sitkElastixImageFilter.h
@@ -23,6 +23,7 @@
 #include "sitkImage.h"
 
 #include <map>
+#include <memory> // For unique_ptr.
 #include <string>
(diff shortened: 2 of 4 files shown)


Code/ElastixTransformixWrappers/include/sitkTransformixImageFilter.h
```diff
@@ -122,7 +123,7 @@ class SITKElastix_EXPORT TransformixImageFilter
 private:
 
   class TransformixImageFilterImpl;
-  TransformixImageFilterImpl* m_Pimple;
+  const std::unique_ptr<TransformixImageFilterImpl> m_Pimple;
```
blowekamp, line 126: We have previously had problems using forward declared/undefined classes as arguments to these types of smart pointers.

N-Dekker: Thanks for your comment, @blowekamp Do you still remember what kind of problems? Because `std::unique_ptr` is really _meant_ to support forward declared Pimpl classes, as shown by the Pimpl example in the C++ Core Guidelines, at https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Ri-pimpl

In order for this to work, the destructor of the exported class (`TransformixImageFilter` in our case) must be implemented in the CXX file (or CPP file), not in the header file. Or more precisely, the Impl class must be fully defined (rather than just forward-declared) at the point of the destructor implementation. As `~TransformixImageFilter()` is implemented in de CXX file that does `#include "sitkTransformixImageFilterImpl.h"`, it should be just fine.  OK?

blowekamp:
It appears that this updated pattern should work with the modern compilers.

Despite SimpleITK having a large number of ITK filters, this code is relatively small and homogeneous. In order to keep SimpleITK consistent, and not have this directory diverge please update the other SimpleITK Pimpl classes to this new pattern.

Also consider if these elastix Pimpl classes are actually needed. The other SITK pimple classes are base classes with derived templated classes for different type. I don't see that used here.

blowekamp (approved):

PR #1669, opened 2022-07-21
BUG: Fix GenericException assignment by using `shared_ptr` internally

N-Dekker:
The assignment operator of `GenericException` originally just copied the "raw" internal pointer (`m_PimpleException`), causing `delete` to be called twice on the very same pointer, when both the original and the copy of the GenericException are destructed. Which would have undefined behavior. Fixed by using `shared_ptr` instead of a raw pointer.

Used `= default` to implement its copy-constructor and its destructor, as `shared_ptr` now takes care of doing the right thing "by default".

### DIFF
--- a/Code/Common/include/sitkExceptionObject.h
+++ b/Code/Common/include/sitkExceptionObject.h
@@ -21,6 +21,8 @@
 #include "sitkMacro.h"
 #include "sitkCommon.h"
 
+#include <memory> // For shared_ptr.
+
 #ifndef sitkMacro_h
 #error "sitkMacro.h must be included before sitkExceptionObject.h"
 #endif // sitkMacro_h
@@ -98,7 +100,7 @@ public:
   virtual unsigned int GetLine() const;
 
 private:
-  const ExceptionObject *m_PimpleException;
+  std::shared_ptr<const ExceptionObject> m_PimpleException;
 };
 
 #ifdef _MSC_VER
--- a/Code/Common/src/sitkExceptionObject.cxx
+++ b/Code/Common/src/sitkExceptionObject.cxx
@@ -33,28 +33,13 @@ GenericException::GenericException() noexcept
   : m_PimpleException( nullptr )
 {}
 
-GenericException::GenericException( const GenericException &e ) noexcept
-
-  : std::exception( e )
-{
-  try
-    {
-    if ( e.m_PimpleException != nullptr )
-      {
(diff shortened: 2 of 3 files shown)


N-Dekker:
FYI, the (originally troublesome) GenericException assignment operator is here:

https://github.com/SimpleITK/SimpleITK/blob/fb426f2811c7ef1853436168d3afdb72e9581b56/Code/Common/src/sitkExceptionObject.cxx#L99-L104 

The original destructor did unconditionally delete the internal pointer, potentially causing undefined behavior when one exception was assigned to another:

https://github.com/SimpleITK/SimpleITK/blob/fb426f2811c7ef1853436168d3afdb72e9581b56/Code/Common/src/sitkExceptionObject.cxx#L93-L96

blowekamp:
Nice fix! SimpleITK tries to follow best testing practices by adding regression tests for these types of bugs when they are encounter. Please add an appropriate test:
https://github.com/SimpleITK/SimpleITK/blob/master/Testing/Unit/sitkExceptionsTests.cxx

Also we don't use the ITK categories in the commit message. Please feel free to make the best use of those recovered characters in the the first line of the comment message.

blowekamp (changes requested):
Please add testing.

blowekamp (approved):
Thank you for the fix and adding the test! Looks go to go.

blowekamp:
@N-Dekker Do you have permissions to hit the green merge button?

N-Dekker:
> Do you have permissions to hit the green merge button?

Yes, I think so, but I wouldn't mind if you press the button either 😄 It feels like a reward to me to see someone actually merging my pull request, rather than having to merge my own pull request, myself. But if necessary, yes I can merge my own PR! 

BTW Personally I would prefer the "Rebase and merge" option from the button, as it would avoid an extra "merge commit" in the git log history. But it appears that "Rebase and merge" is disabled, is that intentional?

blowekamp:
> Yes, I think so, but I wouldn't mind if you press the button either 😄 It feels like a reward to me to see someone actually merging my pull request, rather than having to merge my own pull request, myself. But if necessary, yes I can merge my own PR! 

OK I'll do it.

> BTW Personally I would prefer the "Rebase and merge" option from the button, as it would avoid an extra "merge commit" in the git log history. But it appears that "Rebase and merge" is disabled, is that intentional?

Yes.

PR #1670, opened 2022-07-21
STYLE: Remove include guards from cxx files

N-Dekker:
Include guards are only meant to be used in header files (h and hxx files), as they aim to prevent compile errors when the very same header file is `#include`d multiple times.


blowekamp (approved):
Thank you for cleaning this up.

PR #1671, opened 2022-07-21
ENH: Make Elastix and Transformix `GetName()` members const-correct

N-Dekker:
Fixed the location of the `const` keyword of `GetName()` member function declarations, allowing to retrieve the name from a `const` ElastixImageFilter or TransformixImageFilter object.


blowekamp (approved):

N-Dekker:
Honestly I wonder if these member functions are very useful. They just return "ElastixImageFilter" or "TransformixImageFilter", respectively.  So they might even be declared `static`. Maybe Kasper (@kaspermarstal) still knows their purpose...?

blowekamp:
The method are in all the other SimpleITK messages. Mostly use in printing debug, and warning messages.

PR #1673, opened 2022-07-21
Use unique_ptr over raw in Transform class

blowekamp:


zivy (approved):
Less code with same functionality, I'll call that an improvement.

Code/Common/include/sitkTransform.h
```diff
@@ -93,7 +93,7 @@ class SITKCommon_EXPORT Transform
    */
   template<unsigned int NDimension>
   explicit Transform( itk::CompositeTransform< double, NDimension >* compositeTransform )
-    : m_PimpleTransform( nullptr )
+  : Transform()
```
N-Dekker, line 96: The original line of code `: m_PimpleTransform( nullptr )` worked well already, right?

Code/Common/src/sitkTransform.cxx
```diff
@@ -128,19 +128,16 @@ bool initialized = RegisterMoreTransforms<2>() && RegisterMoreTransforms<3>();
 //
 
 Transform::Transform( )
-  : m_PimpleTransform( nullptr )
+  : m_PimpleTransform( new PimpleTransform<itk::IdentityTransform< double, 3 > >() )
```
N-Dekker, line 131: Please consider `std::make_unique<itk::IdentityTransform< double, 3 > >` instead. The idea is that it's best to avoid _both_ low-level memory management commands, `delete` _and_ `new`, when possible.

Code/Common/include/sitkTransform.h
blowekamp: This where the compilation error I encountered before occoured. Since this in in the header, the std::unique_ptr can not be initialized/assigned/constructed without the Pimple being defined. Hence by calling the compiled default constructor from cxx, the definition of the Pimple is not needed.

N-Dekker: ah, that's clever 👍

Code/Common/src/sitkTransform.cxx
blowekamp: The code should actually be: `std::make_unique< PimpleTransform<itk::IdentityTransform< double, 3 > > >()` which is getting a bit longer, and uglier.

> The idea is that it's best to avoid both low-level memory management commands, delete and new, when possible.

There appears to be less benefits for for this change than using the std smart pointers. Please keep improvements to genuine improvements  not just following current C++ "ideas".

N-Dekker: @blowekamp Seriously, manually doing `delete` _and_ `new` is a major cause of bugs in old C++ code. Please re-consider: _Use `make_unique()` to construct objects owned by `unique_ptr`s_, https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rh-make_unique

blowekamp: There are many benefits of using object for handle the `delete` with object scope and exception handling.

But the linked reasoning for `make_unique` is rather poor:
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rr-make_unique

And the resulting code style is convoluted.

PR #1678, opened 2022-07-25
Disable InPlace option in Cast filter

blowekamp:
Test for potential cause of  BasicFilters.Cast of Float to VectorInt8


blowekamp:
Test fixed with: #1680

PR #1682, opened 2022-08-02
Add self hosted ARM build to github actions.

blowekamp:


.github/workflows/Build.yml
```diff
@@ -25,11 +25,15 @@ jobs:
           - os: windows-2019
```
jwj019, line 25: It may be worth increasing the `max-parallel` here to `4` so all jobs run at the same time.  Otherwise, LGTM

jwj019 (approved):

PR #1694, opened 2022-08-16
WIP: print locale on circle ci

blowekamp:


blowekamp:
the locale LANG was reported as UTF-8

PR #1706, opened 2022-09-08
Update Java.i to set endian type to native before exposing typed buffers

joeyearsley:
I recently had an issue when trying to write into the FloatBuffer from a Tensorflow JVM Tensor FloatBuffer, the issue was caused by the endianess not matching the default OS's endianess. 

I propose we cast into the native OS's endian order before returning these typed buffers, as we can't change the endianess from these typed buffers directly, instead needing to manually perform the added steps.

### DIFF
--- a/Wrapping/Java/Java.i
+++ b/Wrapping/Java/Java.i
@@ -75,7 +75,8 @@
    */
   public java.nio.Buffer getBufferAsBuffer()
     {
-      java.nio.ByteBuffer b = getBufferAsByteBuffer();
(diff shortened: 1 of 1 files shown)


blowekamp:
It is not clear to me know the SimpleITK image could not be in the native byte order. Perhaps the byte order was not correct when initially setting the SimpleITKImage?

gdevenyi:
Is there a test case that could expose this?

joeyearsley:
@gdevenyi Thanks for recommending me to make a test case, it has highlighted the issue is with the Tensorflow JVM library and it's handling of unsafe memory during read/writes rather than SimpleITK, closing this as it's not an issue with SimpleITK.

PR #1711, opened 2022-09-13
DOC: Updating registration overview, return type.

zivy:
Provide a description of the possible return types from the Execute method. General TransformType and underlying concrete types.

### DIFF
--- a/docs/source/registrationOverview.rst
+++ b/docs/source/registrationOverview.rst
@@ -81,7 +81,7 @@ transformations include:
 * `AffineTransform <https://simpleitk.org/doxygen/latest/html/classitk_1_1simple_1_1AffineTransform.html>`_.
 * `BSplineTransform <https://simpleitk.org/doxygen/latest/html/classitk_1_1simple_1_1BSplineTransform.html>`_.
 * `DisplacementFieldTransform <https://simpleitk.org/doxygen/latest/html/classitk_1_1simple_1_1DisplacementFieldTransform.html>`_.
-* `Composite Transform <https://simpleitk.org/doxygen/latest/html/classitk_1_1simple_1_1Transform.html>`_.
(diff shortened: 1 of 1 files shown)


dave3d (approved):

docs/source/registrationOverview.rst
```diff
@@ -235,6 +233,16 @@ The relevant methods are `SetOptimizerScalesFromPhysicalShift <https://simpleitk
 In many cases this scaling is what determines if the the optimization converges to the
 correct optimum.
 
+Results
+======
+
+The `Execute <https://simpleitk.org/doxygen/latest/html/classitk_1_1simple_1_1ImageRegistrationMethod.html#a3ba46dbf038828c8f2952875b8f6e0de>`_ method returns the generic
+`Transform <https://simpleitk.org/doxygen/latest/html/classitk_1_1simple_1_1Transform.html>`_ type. The actual underlying type of the transformation depends on
+the boolean value of the `SetInitialTransform's <https://simpleitk.org/doxygen/latest/html/classitk_1_1simple_1_1ImageRegistrationMethod.html#a6f8d5b96434b73bc9d4c0afacfc2beaf>`_
+`inPlace` parameter. If `True` then the underlying type matches the type of the transformation given as input in the `SetInitialTransform` (e.g. `Euler3DTransform`), otherwise the underlying type is `CompositeTransform`.
```
blowekamp, line 242: I am not sure this explains things the term "generic transform" and "underlying type" don't explain the SimpleITK base transform class as a holder of an arbitrary transform which can be converted or down cast. There is a lot to explain on how the Transform classes behavior, then we have the added complication of the CompositeTransform and the InPlace option too.

zivy: Updated the language to say that the Transform holds an arbitrary transform type.

Not sure what you mean by "There is a lot to explain on how the Transform classes behavior...". If the current change is sufficient then please accept. If not, please edit to add the information you think is missing.

PR #1718, opened 2022-09-16
Midas link requires HTTPS

dave3d:


zivy:
This file [should not be edited](https://github.com/SimpleITK/SimpleITK/blob/ddcb3f43ef5a698ef530c1c1f4afb0ca683484c7/Code/BasicFilters/include/sitkLandmarkBasedTransformInitializerFilter.h#L22) it is auto-generated from ITK. Please update docs in ITK.

blowekamp:
> This file [should not be edited](https://github.com/SimpleITK/SimpleITK/blob/ddcb3f43ef5a698ef530c1c1f4afb0ca683484c7/Code/BasicFilters/include/sitkLandmarkBasedTransformInitializerFilter.h#L22) it is auto-generated from ITK. Please update docs in ITK.

Unfortunately that is not true. It was initially generated with the scrips, but diverged due to separate requirements that could not be done by the generators.

@dave3d Please remove the lines @zivy references too.

blowekamp:
Also the ITK Docs are wrong and those need to be updated too as Ziv said.

dave3d:
BTW: There are a lot of HTTP links throughout the code that should probably be changed to HTTPS.  The link to the Apache License, for instance.

dave3d:
I've fixed the link and squashed the multiple commits.  Also, I have a fix in to ITK.

blowekamp (approved):

Issue #1733, opened 2022-10-19
LBFGSB Bounded optimizer should take in arrays for bounds


hjmjohnson:
The LBFGSB should take in arrays for the bounds.

Providing bounds for the parameters of optimized transforms almost always requires different bounds for each parameter type.

For example, a Euler3D transform is (radians, radians, radians, mm, mm, mm)

One could easily bound to +- 15 mm for the transform, but 15 radians does not make any sense for optimization purposes.

One would like to create bounds like

```python
tendegrees= 10*pi/180

upperbound=[tendegrees, tendegrees, tendegrees, 15, 15,15]
 lowerbound= [ -x for x in upperbound] 
```

```diff
--- a/Code/Registration/src/sitkImageRegistrationMethod_CreateOptimizer.cxx
+++ b/Code/Registration/src/sitkImageRegistrationMethod_CreateOptimizer.cxx
@@ -217,28 +217,30 @@ namespace simple
       #define UPPERBOUND  2 // 10
       #define BOTHBOUND   3 // 11
 
-      unsigned char flag = NOBOUND;
       const unsigned int sitkToITK[] = {0,1,3,2};
-      if ( this->m_OptimizerLowerBound != std::numeric_limits<double>::min() )
-        {
-        flag |= LOWERBOUND;
-        }
-      if ( this->m_OptimizerUpperBound != std::numeric_limits<double>::max() )
+      _OptimizerType::BoundSelectionType boundSelection( numberOfTransformParameters );
+      for(int i =0 ; i < numberOfTransformParameters; ++i)
         {
-        flag |= UPPERBOUND;
+        unsigned char flag = NOBOUND;
+        if ( this->m_OptimizerLowerBound[i] != std::numeric_limits<double>::min() )
+          {
+          flag |= LOWERBOUND;
+          }
+        if ( this->m_OptimizerUpperBound[i] != std::numeric_limits<double>::max() )
+          {
+          flag |= UPPERBOUND;
+          }
+         boundSelection[i] = sitkToITK[flag];
         }
 
-      _OptimizerType::BoundSelectionType boundSelection( numberOfTransformParameters );
-      _OptimizerType::BoundValueType lowerBound( numberOfTransformParameters );
-      _OptimizerType::BoundValueType upperBound( numberOfTransformParameters );
-
-      boundSelection.Fill( sitkToITK[flag] );
-      lowerBound.Fill( this->m_OptimizerLowerBound );
[...]

PR #1746, opened 2022-10-27
Add lbfgsb per parameter

blowekamp:
Addresses #1733


blowekamp:
@hjmjohnson  Do you have a sample program which works with Euler3D and the LBFGSB optimizer? The registration in the test does not seem to work and the optimizer report "no feasible solution", and no line searches perform.

blowekamp:
@hjmjohnson  Any feedback or interest in this enhancement?

blowekamp:
@hjmjohnson I am going to close the addition of this feature since we are not able to create a test or an example which successfully uses bounds per parameters. Contributions are welcomed :)

PR #1750, opened 2022-10-31
WIP: initial checkin

dave3d:
So I wrote this simple landmark registration example.  Basically I ported it from the following ITK example:
https://examples.itk.org/src/registration/common/registerimagetoanotherusinglandmarks/documentation?highlight=landmark

However, the last part isn't working right, the resample with a transform.  The resample moving image seems to be the same as the original moving image.  So the transform isn't being used or something?

@zivy @blowekamp Can you tell me what I did wrong?

Thanks,
Dave


zivy:
Will take a look and get back to you shortly.

zivy (changes requested):
Pleas use more pythonic style (e.g. fixedImage -> fixed_image, fixedLandmarkds -> fixed_landmarks)

Examples/ImageRegistrationMethodLandmark/ImageRegistrationMethodLandmark.py
```diff
@@ -0,0 +1,68 @@
...
+
+landmark_initializer.SetFixedLandmarks(fixed_landmarks)
+landmark_initializer.SetMovingLandmarks(moving_landmarks)
+
+weights = [1, 1, 1]
+landmark_initializer.SetLandmarkWeight(weights)
+
+transform = sitk.Euler2DTransform()
+# transform = sitk.Similarity2DTransform()
+
```
zivy, line 48: Not sure if the intent is to use a similarity or Euler transform? Once that is clear, delete the irrelevant row.

Examples/ImageRegistrationMethodLandmark/ImageRegistrationMethodLandmark.py
```diff
@@ -0,0 +1,68 @@
...
+
+output_image = sitk.Resample(
+    moving_image, fixed_image, transform=output_transform, defaultPixelValue=150
+)
+
+sitk.WriteImage(output_image, "landmark_example.png")
+# sitk.WriteImage(moving_image, "moving_image.png")
+# sitk.WriteImage(fixed_image, "fixed_image.png")
+
```
zivy, line 62: Please delete commented out code.

zivy (approved):
In the overview section of the docs, say something about the minimal number of landmarks required, which is transformation dependent. Also list the transforms that are supported by the landmark based initializer. 

One last thing that I'm not sure about, I don't think that the weights are used for the 2D rigid case. I vaguely remember the weights were only used in the estimation of affine transforms. Please check and update, or not.

blowekamp:
Doesn't `Examples/CMakeLists.txt` need to be includes to that the tests will run?

dave3d:
> Doesn't `Examples/CMakeLists.txt` need to be includes to that the tests will run?

You're right.  I forgot to add/commit that file.

blowekamp:
@dave3d  Please rebase onto master, which should help some CI failures and address the ghost flow error so that the example is an executable.

blowekamp (changes requested):
a docs/source/link_ImageRegistrationMethodLandmark.rst file with the appropriate "include" line is also needed.

blowekamp (approved):

PR #1786, opened 2022-12-01
Sphinx version and en lang

mbopfNIH:
Modified sphinx version in requirements.txt to ~5.3.0 and ran "make html"
Also changed "language" in conf.py from None to "en" to fix warning.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -64,7 +64,9 @@ master_doc = 'index'
 #
 # This is also used if you do content translation via gettext catalogs.
 # Usually you set "language" from the command line for these cases.
-language = None
+# Setting default to English to avoid warning.
+#language = None
+language = 'en'
 
 # List of patterns, relative to source directory, that match files and
 # directories to ignore when looking for source files.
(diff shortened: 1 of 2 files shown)


docs/source/conf.py
```diff
@@ -64,7 +64,8 @@
 #
 # This is also used if you do content translation via gettext catalogs.
 # Usually you set "language" from the command line for these cases.
-language = None
+# Mike Bopf - 01Dec2022 - setting default to English to avoid warning.
```
zivy, line 67: Please change to "Setting default to English to avoid warning" without the prefix information. Who made the change and when is recorded via the git commits. 

On the command-line if you run `git blame file_name` you will see who was the last person to touch each line of the file. On github you can navigate to the source file and on the upper right corner there is a "blame" button which will display this info.

zivy (changes requested):

blowekamp:
Please squash the last two commits together.

blowekamp (approved):
Thanks for fixing up the commit history.
The PR looks good to go now.

zivy (approved):

PR #1799, opened 2022-12-19
Typo in DicomSeriesFromArray example

umasehs:
Currently, the DicomSeriesFromArray  example says to set the DICOM metadata key "0020,0013" for the instance number. This does not work with the SetMetaData function, and should instead be "0020|0013". I hope this is useful, but no worries if it's not.


zivy:
Hello @umasehs,

Good catch, thanks for fixing it. Once testing passes will merge.

Issue #1798, opened 2022-12-16
Embed the output of DicomImagePrintTags in Sphinx


blowekamp:
This example only contains text output and is a good starter where the existing Sphinx extensions doctest or other extension should work.
Part of #1797

mbopfNIH:
Having difficulty finding a good Sphinx extension for running the Example code. I thought [sphinx-exec-code](https://sphinx-exec-code.readthedocs.io/en/latest/index.html) would work, but it doesn't allow me to reference a subset of the Example code file.

mbopfNIH:
I now have sphinx-exec-code is now running, but requires a DICOM image as input, and I don't see any in the "docs/images" path in the sphinx build - it's all jpg, png, and svg. Will look into how to get on in there and find an appropriate image.

zivy:
Hi @mbopfNIH, 

Possibly use the files that are referenced in the testing of the examples which use DICOM (e.g. [DicomConvert](https://github.com/SimpleITK/SimpleITK/blob/master/Examples/DicomConvert/CMakeLists.txt)). These are images that are already in the binary data store, see their hash files under the [Testing/Data/Input](https://github.com/SimpleITK/SimpleITK/tree/master/Testing/Data/Input/DicomSeries) directory. 

If they are not visually compelling, you may need to add several DICOM images to the binary data store under the docs/images path.

mbopfNIH:
Thanks, Ziv. I’m trying to understand the make/cmake build process and how the CMake ExternalData module knows where our images are stored. A make -debug shows that they are coming from an AWS s3 bucket, but I can’t see where that is set – I suspect some “magic” is going on (which magic defined as anything I don’t understand).

In any case, I’ll take a look at the 3 DICOMs I can find in Testing/Data/Input.

From: Ziv Yaniv ***@***.***>
Date: Tuesday, December 27, 2022 at 10:50 AM
To: SimpleITK/SimpleITK ***@***.***>
Cc: Bopf, Mike (NIH/NIAID) [C] ***@***.***>, Mention ***@***.***>
Subject: [EXTERNAL] Re: [SimpleITK/SimpleITK] Embed the output of DicomImagePrintTags in Sphinx (Issue #1798)

[...]

PR #1803, opened 2023-01-03
Include the DicomImagePrintTags.py output in Sphinx docs

mbopfNIH:
Added sphinx_exec_code directive and used it to run the Python example as a subprocess in the rst file using the Image0075.dcm as input.

Fixes  issue #1798.


Examples/DicomImagePrintTags/Documentation.rst
```diff
@@ -32,3 +32,17 @@ Code
...
+Sample Output
+-------------
+
+Running the above Python script with the image  "docs/images/Image0075.dcm" produces the following output:
+
+.. exec_code::
+
+    # --- hide: start ---
+    import subprocess
+    result = subprocess.call(['python',
```
blowekamp, line 45: Please see this change for a better way to get the current python executable.
64bd1f84

blowekamp:
Looks pretty good.

mbopfNIH:
I've changed the code to eliminate the trailing whitespace, and improve the method of determining the python executable. But I don't know how to get the ghostflow check to re-run with my new commit. Also, is there a way to stop the old tests and restart with the new code?

mbopfNIH:
Made suggested changes to get correct python executable, fixed ghostflow issue, and squashed commits. Build appears to be failing only because of that numpy read-only issue. (At least I think so).

docs/requirements.txt
```diff
@@ -2,3 +2,4 @@ sphinx-tabs
 cmake
 sphinx_rtd_theme
 sphinx~=5.3.0
+sphinx_exec_code~=0.8
```
blowekamp, line 5: SimpleITK is also going to be needed here.

mbopfNIH: Right. Should I specify a version? Maybe >=2.2 ?

blowekamp: How about "~=2.2.0"

mbopfNIH: Okay, but that wouldn't match 2.3 and above, right? So would that require an update this requirements.txt each time we push a minor upgrade?

blowekamp: Yes. But it would also ensure if an incompatible change was made the older versions could still be built.

mbopfNIH:
I may have done an unnecessary merge to get my branch synced with the upstream master. But the code change is simply the extra simpleitk~=2.2.0 in requirements.txt.

blowekamp:
The way to update a topic or PR branch with the changes of master is to rebase it onto the master branch. We don't want the commit history to have unnecessary merges in the history which would create confusion and clutter. Also, the changes from master were not needed for this topic/PR.

PR #1810, opened 2023-01-06
Exclude problematic numpy versions

blowekamp:
See SimpleITK#1805 for details.

Explude version of numpy which cause Python.Test.ArrayView to fail.


zivy (approved):

mbopfNIH (approved):
I'm still relearning cmake syntax, but it looks like avoiding those broken versions of numpy in the test environment should work. Will this allow the PRs to complete?

blowekamp:
Yes, this should make the CI for the PR green again. If a PR is rebased onto master.

PR #1811, opened 2023-01-09
Teach Python masked assign to support vector images

blowekamp:
Updated MaskedAssignImageFilter to have an optional AssignImage, and support constant assignments. This  duplicates the functionality of the MaskNegatedImageFilter, but support vector images.

### DIFF
--- a/Code/BasicFilters/json/MaskedAssignImageFilter.json
+++ b/Code/BasicFilters/json/MaskedAssignImageFilter.json
@@ -1,15 +1,14 @@
 {
   "name" : "MaskedAssignImageFilter",
-  "template_code_filename" : "DualImageFilter",
+  "template_code_filename" : "ImageFilter",
   "template_test_filename" : "ImageFilter",
-  "filter_type" : "itk::MaskedAssignImageFilter<InputImageType, InputImageType2, OutputImageType>",
+  "filter_type" : "itk::MaskedAssignImageFilter<InputImageType, itk::Image<std::uint8_t, InputImageType::ImageDimension>, OutputImageType>",
   "number_of_inputs" : 0,
   "doc" : "",
   "include_files" : [
     "sitkToPixelType.hxx"
   ],
   "pixel_types" : "NonLabelPixelIDTypeList",
-  "pixel_types2" : "MaskedPixelIDTypeList",
   "inputs" : [
     {
       "name" : "Image",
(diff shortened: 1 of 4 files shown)


Testing/Unit/Python/sitkImageTests.py
```diff
@@ -467,6 +467,21 @@ def test_masked_assign_constant(self):
...
+        for pixel_type in (sitk.sitkVectorFloat32, sitk.sitkVectorFloat64,
+                           sitk.sitkVectorUInt8, sitk.sitkVectorInt8,
+                           sitk.sitkVectorUInt16, sitk.sitkVectorInt16,
+                           sitk.sitkVectorUInt32, sitk.sitkVectorInt32,
+                           sitk.sitkVectorUInt64, sitk.sitkVectorInt64, ):
+            image = sitk.Image([2, 2], pixel_type)
+
+            image[mask_image] = 2
+            self.assertTrue(all(i == 0 for i in image[0,0]))
+            self.assertTrue(all(i == 2 for i in image[1,0]), image[1,0])
```
zivy, line 480: Statement looks strange, not sure if this is correct.

zivy (commented):
Not sure what is meant by "Teach" in the commit message, seems strange possibly remove?

Code/BasicFilters/json/MaskedAssignImageFilter.json
```diff
@@ -42,24 +53,55 @@
...
       "inputs" : [
         "Input/cthead1-Float.mha",
         "Input/cthead1-mask.png",
         "Input/cthead1-Float.mha"
       ]
     },
     {
-      "tag" : "rgb",
-      "description" : "rgb",
+      "tag" : "cthead1_consant",
```
dave3d, line 64: should that be cthread1_constant?

Code/BasicFilters/json/MaskedAssignImageFilter.json
```diff
@@ -42,24 +53,55 @@
...
+      "md5hash" : "25219307a68bebbe202040ea61c7687d",
       "inputs" : [
         "Input/VM1111Shrink-RGB.png",
         "Input/VM1111Shrink-mask.png",
         "Input/VM1111Shrink-RGB.png"
       ]
+    },
+    {
+      "tag" : "rgb_constant",
+      "description" : "assign consant to rgb",
```
dave3d, line 92: constant (sp)

Testing/Unit/Python/sitkImageTests.py
blowekamp: The second argument is a message. In this case it would just be the actual value. It is not consistent with the other vector comparisons. But it is OK.

zivy (approved):

PR #1819, opened 2023-01-12
ENH: Dependency between GetGDCMSeriesIDs and GetGDCMSeriesFileNames.

zivy:
The GetGDCMSeriesFileNames supports a useSeriesDetails flag which appends tag based information to the seriesUID. This "decorated" seriesUID is then used to identify files as belonging to the series. The GetGDCMSeriesIDs did not support the useSeriesDetails flag, so always returned the non-decorated seriesUID which in combination with the decorated seriesUID used by
GetGDCMSeriesFileNames returns no matching files. This commit adds the flag to the GetGDCMSeriesIDs so that returns "decorated" seriesUIDs when it is set to true and will work as expected with GetGDCMSeriesFileNames when that is called with the flag set to true too.

### DIFF
--- a/Code/IO/include/sitkImageSeriesReader.h
+++ b/Code/IO/include/sitkImageSeriesReader.h
@@ -99,7 +99,10 @@ namespace itk {
        * \param seriesID          Set the name that identifies a
        * particular series. Default value is an empty string which
        * will return the file names associated with the first series
-       * found in the directory.
+       * found in the directory. If the parameter value was obtained
+       * from a call to GDCMSeriesIDs, the value of the
+       * useSeriesDetails parameter must be the same here and in the
+       * call to GDCMSeriesIDs.
        * \param useSeriesDetails  Use additional series information
        * such as ProtocolName and SeriesName to identify when a single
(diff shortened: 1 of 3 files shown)


Code/IO/include/sitkImageSeriesReader.h
```diff
@@ -119,7 +119,8 @@ namespace itk {
        * \param directory  The directory that contains the DICOM data set
        * \sa itk::GDCMSeriesFileNames
        **/
-      static std::vector<std::string> GetGDCMSeriesIDs( const std::string &directory );
+      static std::vector<std::string> GetGDCMSeriesIDs( const std::string &directory,
```
blowekamp, line 122: Should we document in the function documentation that the value of `useSeriesDetails` should match when the GetGDCMSeriesIDs/FileNames is used?

zivy: Good catch. Documentation updated.

dave3d (approved):

blowekamp (approved):

Issue #1797, opened 2022-12-16
For Sphinx docs run example and embed output images and text into docs


blowekamp:

mbopfNIH:
Working on a solution that uses the Matplotlib.sphinxext.plot_directive sphinx extension. This is used by scikit-image as well as Matplotlib itself to display images in the documentation. May need to build a wrapper script since many of the Examples automatically show the image using Fiji which needs to be turned off via the SITK_NOSHOW environment variable.

mbopfNIH:
**Summary Solution**
The current code uses three different Sphinx extensions to produce output images and text by running the Example python code and Documentation.rst (included in conf.py and requirements.txt):
- `matplotlib.sphinxext.plot_directive`: used to create the plots and make them available to the Sphinx build
- `sphinx_exec_code`: used to run the Python code to include any text output, if desired
- `sphinx_toolbox.collapse`: allows the user to hide long sections of text output by collapsing it

**Code Structure**
- Modified the Example python code to use a standard `def main(args):` routine which is called from within the RST
- Created a helper module, `docs/source/example_utils.py` which contains two functions:
1. `plot_image(image: SimpleITK.Image, cmap: str = 'gray'):` which uses Matplotlib to output a SimpleITK image to make it accessible to Sphinx
2. `run_example(module_name: str, func_name: str, args: list) -> dict:` which encapsulates the standard `main(args)` routine. Since it takes a `func_name`, using "main" is just a convention, and the code will accept other names.

The above routines are called from within the RST file for each example using the code section of the ..plot:: directive as so:
`.. plot::
   :align: center
   :nofigs:

    import os
    import sys

    from example_utils import run_example, plot_image
    args = [arg1, arg2, arg3...]
    
    return_dict = run_example(<ExampleName>, 'main', args)
    for key, value in return_dict.items():
        plot_image(value)`

Note that the run_example expects the main routine to return a `dict[str, SimpleITK.Image]` which is iterated through and all desired images displayed using the `plot_image` routine.
Also note that the `:nofigs:` option is usually included in the plot directive to avoid directly including the images in the output. This is done to provide more control over the display of the images, such as:
`.. list-table::

    * - .. figure:: link_ImageRegistrationMethod1_docs-1_00.png
           :scale: 100%

           **Fixed Image**

      - .. figure:: link_ImageRegistrationMethod1_docs-1_01.png
           :scale: 100%

           **Moving Image**`

**Issues**
[...]

PR #1823, opened 2023-01-13
Used matplotlib Sphinx plot directive to embed output image in the docs.

mbopfNIH:
First fix for #1797

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -45,8 +45,13 @@ extensions = [
     'sphinx.ext.mathjax',
     'sphinx.ext.ifconfig',
     'sphinx_exec_code',
-    'sphinx_tabs.tabs'
+    'sphinx_tabs.tabs',
+    "matplotlib.sphinxext.plot_directive",
 ]
+
+# Suppress hyperlinks in the display of matplotlib.sphinxext.plot_directive plots
+plot_html_show_source_link = False
+plot_html_show_formats = False
 
 # Add any paths that contain templates here, relative to this directory.
 templates_path = ['_templates']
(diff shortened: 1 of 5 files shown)


mbopfNIH:
Comments on this fix:

- Added function call similar to DicomPrintImageTags solution, but could avoid this since there are no arguments passed
- Made assumption of 96dpi display to get plt.imshow to present at "correct" size
- Added extra conf.py parameters to avoid hyperlinks to the source code and images
- The function name get_smiley() isn't "professional", but overloading main() could be confusing. Suggestions?

docs/source/conf.py
```diff
@@ -45,9 +45,14 @@
     'sphinx.ext.mathjax',
     'sphinx.ext.ifconfig',
     'sphinx_exec_code',
-    'sphinx_tabs.tabs'
+    'sphinx_tabs.tabs',
+    "matplotlib.sphinxext.plot_directive",
 ]
 
+# Supress hyperlinks in the display of matplotlib.sphinxext.plot_directive plots
```
dave3d, line 52: suppress (sp)

dave3d (approved):

docs/source/conf.py
mbopfNIH: Oops. I'll fix it next time I touch that file.

zivy (commented):
Generally looks good, but I would aim for a more generic code layout. 

Instead of `create_smiley`, have all examples that return an image use the function signature `main(args)` and the `if __name__==main` pattern. In this particular case the args are ignored. 

The wrapper script then becomes generic, always importing main from the example module (`HelloWorld.py` etc.), forwarding the commandline arguments to the main function and getting the resulting image from it. No need for separate `ExampleX_Wrapper.py` per example.

blowekamp:
@Zivy Generalizing sounds like a good second step. Let's get this case working in RtD, and then generalize it to more examples.

PR #1836, opened 2023-01-23
Adding documentation on specifying pixel type as part of reading.

### LINKED ISSUE #1835: Add pixel output type for ReadImage in Sphinx docs
It is common to convert images to a specific type when reading. This could be to convert a color image to a scalar when the expected input is a binary image, or to convert images to a floating point type.

An additional code block and accompanying prose should be added to IO.rst explain how this could be done and listing some of the convent reasons why this should be done.

zivy:
resolves #1835

### DIFF
--- a/docs/source/IO.rst
+++ b/docs/source/IO.rst
@@ -145,6 +145,29 @@ A more compact example using SimpleITK's procedural interface:
 Similarly, if the imageIO parameter is omitted, SimpleITK will determine
 which IO to use automatically.
 
+Finally, one can also specify the desired pixel type for the returned image,
+regardless of the underlying pixel type. This is done via the ``SetOutputPixelType`` of
+the ``ImageFileReader`` or setting the ``ReadImage``'s ``outputPixelType`` parameter.
+
+When reading a scalar image and specifying a different scalar type from the underlying one,
+the returned image's pixel type will be a simple casting of the actual type. This casting can result
+in loss of data when performing a narrowing conversion. For example, if
+the underlying pixel type is ``sitk.sitkFloat32`` and the specified pixel type
+is ``sitk.sitkUInt8``. Explicitly specifying a pixel type is commonly done in conjunction
+with the SimpleITK registration framework. This framework only accepts images with
+float pixel types. Therefore, it is common to read the images and specify their pixel type as
+``sitk.sitkFloat32`` or ``sitk.sitkFloat64``.
+
+When reading a vector image and specifying the pixel type as a scalar, the returned pixel
+values are computed based on the number of channels in the original image. This conversion makes
+several assumptions, three channels mean an RGB image, four channels an RGBA image. These assumptions are
+useful, but may not be what you need if the original image was not in these color-spaces (e.g. HSV).
+
+For those looking for further details, the ITK class which deals with these conversions is
+`ConvertPixelBuffer <https://itk.org/Doxygen/html/classitk_1_1ConvertPixelBuffer.html>`_.
+
+
+
 .. _transformation-io:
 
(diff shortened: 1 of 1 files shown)


docs/source/IO.rst
```diff
@@ -145,6 +145,28 @@ A more compact example using SimpleITK's procedural interface:
...
+in loss of data when performing a narrowing conversion. For example, if
+the underlying pixel type is ``sitk.sitkFloat32`` and the specified pixel type
+is ``sitk.sitkUInt8``. Explicitly specifying a pixel type is commonly done in conjunction
+with the SimpleITK registration framework. This framework only accepts images with
+float pixel types. Therefore, it is common to read the images and specify their pixel type as
+``sitk.sitkFloat32`` or ``sitk.sitkFloat64``.
+
+When reading a vector image and specifying the pixel type as a scalar, the returned pixel
+values are computed based on the number of channels in the original image. This conversion makes
+several assumptions, three channels mean an RGB image, four channels an RGBA image. This is useful,
```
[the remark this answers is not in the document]
zivy: done

docs/source/IO.rst
```diff
@@ -145,6 +145,28 @@ A more compact example using SimpleITK's procedural interface:
 Similarly, if the imageIO parameter is omitted, SimpleITK will determine
 which IO to use automatically.
 
+Finally, one can also specify the desired pixel type for the returned image,
+regardless of the underlying pixel type.
+
+When reading a scalar image and specifying a different scalar type from the underlying one,
+the returned image's pixel type will be a simple casting of the actual type. This can result
```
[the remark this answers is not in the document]
zivy: done

dave3d (approved):

docs/source/IO.rst
```diff
@@ -145,6 +145,28 @@ A more compact example using SimpleITK's procedural interface:
...
+float pixel types. Therefore, it is common to read the images and specify their pixel type as
+``sitk.sitkFloat32`` or ``sitk.sitkFloat64``.
+
+When reading a vector image and specifying the pixel type as a scalar, the returned pixel
+values are computed based on the number of channels in the original image. This conversion makes
+several assumptions, three channels mean an RGB image, four channels an RGBA image. These assumptions are
+useful, but may not be what you need if the original image was not in these color-spaces (e.g. HSV).
+
+For those looking for further details, the ITK class which deals with these conversions is
+``ConvertPixelBuffer``.
```
blowekamp, line 166: Please add a line to the ITK Doxygen for this class?

docs/source/IO.rst
```diff
@@ -145,6 +145,28 @@ A more compact example using SimpleITK's procedural interface:
 Similarly, if the imageIO parameter is omitted, SimpleITK will determine
 which IO to use automatically.
 
+Finally, one can also specify the desired pixel type for the returned image,
+regardless of the underlying pixel type.
```
blowekamp, line 149: How?

blowekamp (commented):
Perhaps the second procedural examples should be updated to set the `outputPixel` Type.

docs/source/IO.rst
zivy: done

docs/source/IO.rst
zivy: updated

zivy:
I don't think we should change the current examples. 

The explicit specification of the output pixel type is the exception and not the rule. Don't want to teach folks to do this when they start learning SimpleITK. For less experienced users will likely result in bugs, unexpected mismatch between actual and read pixel values.

blowekamp (approved):

PR #1842, opened 2023-01-26
Adding GHA file to perform rolling builds

blowekamp:


.github/workflows/BatchBuild.yml
```diff
@@ -0,0 +1,104 @@
+name: Batch Build and Test
+
+on:
+  push:
+   branches: ['master', 'release']
+  pull_request:
+    paths: .github/workflows/BatchBuild.yml
+
```
zivy, line 8: Would it be useful to also allow manual triggering? 
```
on:
  push:
   branches: ['master', 'release']
  pull_request:
    paths: .github/workflows/BatchBuild.yml
  # Enable manual running of workflow
  workflow_dispatch:
```

zivy (approved):

.github/workflows/BatchBuild.yml
blowekamp: What is the use case where this would be useful?

zivy: Had the following experience:
I limited the triggers too much, to specific file level, and made a PR with a new file which I forgot to add to the triggers. Testing didn't run because the added file wasn't in the list. Triggered manually to see if things were ok, which they were. Then fixed the workflow. Probably should have just fixed the workflow.

blowekamp: OK, we want to manually be able to trigger it for PR as needed 👍

PR #1846, opened 2023-01-27
Examples words

dave3d:


.github/workflows/additional_dictionary.txt
```diff
@@ -99,6 +99,7 @@ Centroid
 Chamfer
 CharEqn
 Charvat
+Clunie's
```
blowekamp, line 102: Are `'s' automatic? Could just "Clunie" be added?

PR #1852, opened 2023-01-30
Generalized Sphinx code to display inputs and outputs across Examples.

### LINKED ISSUE #1851: In Sphinx docs, create run_example to generically run examples.
The existing `gaussian_utils.py` file should be generalized. There should be a method `run_example` which takes a module and function as string(s), imports then and then executes with the remaining arguments of `run_example`.

As a second example, use hello world to also call this method.

mbopfNIH:
closes #1851


mbopfNIH:
This doesn't include **kwargs inputs, but it does include a dict for the return values. This was necessary for it to work with both HelloWorld and SimpleGaussian. It can still use a little tweaking, but it "works" and should be somewhat extensible.

docs/source/example_utils.py
```diff
@@ -21,13 +20,14 @@ def plot_image(image: SimpleITK.Image):
     plt.imshow(nda, cmap='gray')
 
 
-def run_gaussian():
+def run_example(module_name, func_name, args):
+    sys.path.append(f'../../Examples/' + module_name)
```
blowekamp, line 24: Please document this functions and arguments with doc strings: https://peps.python.org/pep-0257/

zivy (approved):

docs/source/example_utils.py
blowekamp: You might as well use a standard as oppose to something arbitrary.

mbopfNIH:
On the *args, **kwargs change, this will break any existing code, including any tests, that depend on the current input method. Not sure how much that matters, but I'll need to figure out how to update the tests.

blowekamp:
The changes needed for args kwargs can be done in a separate PR.

docs/source/example_utils.py
```diff
@@ -0,0 +1,58 @@
...
+import SimpleITK
+import matplotlib.pyplot as plt
+import importlib
+
+# Hard-coding a screen dpi of 96: calculating it appears to be difficult
+SCREEN_DPI = 96
+
+
+def plot_image(image: SimpleITK.Image, cmap='gray'):
+    """This takes a SimpleITK Image and plots it using Maptplotlib.
```
blowekamp, line 15: sp:  Maptplotlib->Matplotlib

blowekamp (commented):
LGTM

PR #1857, opened 2023-02-01
Added images to the FastMarchingSegmentation example.

### LINKED ISSUE #1855: Update the FastMarchingSegmentation example to contain images
This example was based on the ITK example: https://itk.org/Doxygen/html/Examples_2Segmentation_2FastMarchingImageFilter_8cxx-example.html

Which contains more information. Also the ITK example writes several images which the SimpleITK example should also display such as the speed image and the time-crossing map.

mbopfNIH:
closes #1855

### DIFF
--- a/Examples/FastMarchingSegmentation/FastMarchingSegmentation.py
+++ b/Examples/FastMarchingSegmentation/FastMarchingSegmentation.py
@@ -21,65 +21,79 @@
 import SimpleITK as sitk
 import sys
 
-if len(sys.argv) < 10:
-    print(
-        "Usage:",
-        sys.argv[0],
-        " <inputImage> <outputImage> <seedX> <seedY> <Sigma>",
-        "<SigmoidAlpha> <SigmoidBeta> <TimeThreshold>",
-    )
-    sys.exit(1)
 
-inputFilename = sys.argv[1]
-outputFilename = sys.argv[2]
+def main(args):
+    if len(args) < 10:
+        print(
+            "Usage:",
+            "FastMarchingSegmentation",
+            "<inputImage> <outputImage> <seedX> <seedY> <Sigma>",
+            "<SigmoidAlpha> <SigmoidBeta> <TimeThreshold>",
+            "<StoppingTime>"
+        )
+        sys.exit(1)
 
-seedPosition = (int(sys.argv[3]), int(sys.argv[4]))
+    inputFilename = args[1]
+    outputFilename = args[2]
 
-sigma = float(sys.argv[5])
-alpha = float(sys.argv[6])
-beta = float(sys.argv[7])
-timeThreshold = float(sys.argv[8])
-stoppingTime = float(sys.argv[9])
+    seedPosition = (int(args[3]), int(args[4]))
 
-inputImage = sitk.ReadImage(inputFilename, sitk.sitkFloat32)
+    sigma = float(args[5])
+    alpha = float(args[6])
+    beta = float(args[7])
+    timeThreshold = float(args[8])
+    stoppingTime = float(args[9])
 
-print(inputImage)
+    inputImage = sitk.ReadImage(inputFilename, sitk.sitkFloat32)
 
-smoothing = sitk.CurvatureAnisotropicDiffusionImageFilter()
-smoothing.SetTimeStep(0.125)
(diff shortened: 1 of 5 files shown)


mbopfNIH:
Initial cut so you can take a look. Naming of the images may not be correct - I used the ITK comments to hopefully figure out the "SpeedImage" and the "TimeCrossingMap". Added a "Title" to the plot_image() routine and used the dict keys. For some reason the Title is getting chopped in the HelloWorld example...

mbopfNIH:
Also, I didn't separate the inputs and outputs because that currently would require me to rerun the Example code again. But maybe not worth the savings. I commented out the sigmoid.DebugOn() call because it was showing up in the build. I'm currently not displaying the `print(inputImage)` in the docs, but I could do that with exec_code().

docs/source/example_utils.py
```diff
@@ -25,8 +25,10 @@ def plot_image(image: SimpleITK.Image, cmap: str='gray'):
     size_pixels = image.GetSize()
 
     nda = SimpleITK.GetArrayFromImage(image)
-    plt.figure(figsize=(size_pixels[0] / SCREEN_DPI, size_pixels[1] / SCREEN_DPI))
+    plt.figure(figsize=(size_pixels[1] / SCREEN_DPI, size_pixels[0] / SCREEN_DPI))
```
blowekamp, line 28: I am finding the images in the other examples a little small. This formula seem to let the size of the figure vary based on the size of the simpleitk image. Do we want that?

Should the images always take up about the same size in the document? 

Also the `dpi` argument is not specified.

mbopfNIH: I wrote that code to get the HelloWorld image to display at about the same size as the sitk.Show, at least with Fiji. It may be that the matplotlib plot_directive is scaling the images, which would be "bad".

Examples/FastMarchingSegmentation/FastMarchingSegmentation.py
```diff
@@ -21,65 +21,81 @@
...
+    result = thresholder.Execute(fastMarchingOutput)
+
+    sitk.WriteImage(result, outputFilename)
+
+    image_dict = {"InputImage": inputImage,
+                  "SmoothingOutput": smoothingOutput,
+                  "GradientOutput": gradientMagnitudeOutput,
+                  "SpeedImage": sigmoidOutput,
+                  "TimeCrossingMap": fastMarchingOutput,
+                  "Segmentation": result,
```
blowekamp, line 95: We don't need to display the smoothing or the gradient output images. Please remove them from the list.

docs/source/example_utils.py
blowekamp: There is also the image directives matplotlib_plot support such as `:width:` and `:height:` which could effect with how the rendered figure is displayed on the sphinx html document.

mbopfNIH: I guess my goal was to avoid the resampling by creating a display size similar to the original image. But on second thought that isn't really necessary. I tried the :width: and :height: options on HelloWorld, but they didn't seem to have an effect, but I'll try again with FastMarching. I could also just pass in another argument to plot_image() to control it from the RST file.

blowekamp: The align is not working as expected either. We should figure out how to get those image directives to work.

mbopfNIH: It strange, the :width: has an effect, although it's after the image creation, so the title ends up shrinking, too. And while :align: right has an effect, it gets partially covered by the code. I also tried putting the whole thing in a Sphinx container, but the plot_directive doesn't seem to want to nest.

mbopfNIH:
This version removes the extra images and uses a fixed Figsize (3,3). This could be sent to plot_image() as another parameter. Alternatively, I could possibly send plot_image a **kwargs that would get passed to Matplotlib.

blowekamp (approved):
LGTM

mbopfNIH:
Hopefully it's okay to merge this without it passing all the tests.

mbopfNIH:
Oops - it looks like merging *stopped* the tests. I'm guessing that's not what we wanted.

blowekamp:
It is fine. New ones got triggered on the merge.

PR #1861, opened 2023-02-02
Changed images to render 1:1 and display centered.

### LINKED ISSUE #1831: Reformat Output in Sphinx
The the Sphinx example the output/input should be before the sample code. Output could be an image or text.

Also images should be centered.

### LINKED ISSUE #1859: Improve formatting for plotted images in Sphinx docs.
The SimpleITK images should be rendered to PNG files with 1:1 pixels and no interpolations performed. The images should be centered. And The size of the images should be controlled with the sphinx mark up, so the images don't look so small in the document.

mbopfNIH:
Resolves #1859, resolves #1831

### DIFF
--- a/Examples/FastMarchingSegmentation/Documentation.rst
+++ b/Examples/FastMarchingSegmentation/Documentation.rst
@@ -48,11 +48,9 @@ Running the Python code with the following inputs:
 
 produces the input & output images below.
 
-Input/Output Images
-^^^^^^^^^^^^^^^^^^^
-
 .. plot::
    :align: center
+   :nofigs:
 
     from example_utils import run_example, plot_image
     args = ['../../docs/images/BrainProtonDensitySlice.png',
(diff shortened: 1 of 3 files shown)


mbopfNIH:
Another attempt. I read that matplotlib fig.figimage produces 1:1 images, but without titles, axes or anything. Also, needed to use DPI=80 for this to work, which is the plot_directive default for PNG output. Also changed the RST to manually display the images written to build/html by the plot_directive instead of letting plot do it. This allowed the :align: center to work. The size can also be controlled by :width: & :height: or even better, by using :scale:.
However, Sphinx doesn't allow captions using .. image:: and the title is gone for the images. Also relying on the ..plot:: naming convention (link_FastMarchingSegmentation_docs-1_00.png) to access the images in order.

blowekamp:
Please update the PR message to include keywords to automatically close the issue:
https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword#linking-a-pull-request-to-an-issue-using-a-keyword

blowekamp:
Can you post a screen shot of the html sphinx rendering?

mbopfNIH:
Screen Shot 2023-02-03 at 8 46 32 AM

mbopfNIH:
On the auto-close comments, I did not intend for these changes to close the issue, but I wanted you to see the code/results so we could discuss. I guess it won't close the issue until I merge, so I'll change it now.

mbopfNIH:
Sphinx is annoying sometimes. Images don't allow a caption; a Figure does accept a caption, but doesn't allow the :align: option. Sigh. I'm looking into putting them into some sort of container to get a better display.

blowekamp:
Some things may be dictated by the "style" template of the sphinx document. Let us not try to force things too much. I think if we can get the images to be displayed at a reasonable size that is OK.

mbopfNIH:
My current solution provides us complete control of the size of the images from within the Sphinx markup. The drawback is not being able to center the text descriptions, and having to rely on the ordering and naming conventions of the images created by the plot_directive, e.g.:
../build/html/link_FastMarchingSegmentation_docs-1_00.png
../build/html/link_FastMarchingSegmentation_docs-1_01.png
../build/html/link_FastMarchingSegmentation_docs-1_02.png

are **assumed** to be the Input Image, the Speed Image, etc. because of the order.

mbopfNIH:
Moving the images back to the left allowed me to use ..figure:: instead of ..image:: which allows for a caption but still can use :scale: the same way. I think it's a little cleaner this way, but let me know if you disagree.

Screen Shot 2023-02-03 at 1 21 00 PM

blowekamp (approved):
LGTM

PR #1868, opened 2023-02-06
Add self-hosted-x64

jwj019:
This PR adds Monarch self-hosted Github runners on x64 architecture.

### DIFF
--- a/.github/workflows/Build.yml
+++ b/.github/workflows/Build.yml
@@ -99,7 +99,7 @@ jobs:
     - name: Install Monarch self-hosted build dependencies
       if: matrix.os == 'self-hosted-x64'
       run: |
-        apt install -y \
+        apt-get install -y \
           mono-devel \
           r-base \
           r-base-dev \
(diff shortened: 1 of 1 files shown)


.github/workflows/Build.yml
```diff
@@ -88,6 +96,15 @@ jobs:
       run: |
         python -m pip install --upgrade pip
         python -m pip install ninja scikit-ci-addons cmake~=3.18.0
+    - name: Install Monarch self-hosted build dependencies
+      if: matrix.os == 'self-hosted-x64'
+      run: |
+        apt install -y \
```
blowekamp, line 102: Why apt over apt-get? I see a warning indicating that apt's interface is not stable.

blowekamp (approved):
Awesome!

.github/workflows/Build.yml
jwj019: Fingers work faster than brain sometimes.  I'll change this to apt-get then merge.

PR #1869, opened 2023-02-06
Spell check the ".py" and ".cxx" files

dave3d:
Also the entire repository is spell checked, not just the files in the Code sub-directory.


zivy (approved):
Very nice. We should have done this years ago.

PR #1874, opened 2023-02-08
Adding images to the N4 Bias Example

### LINKED ISSUE #1873: Improve the N4 bias field correction example to include images in the output.
https://github.com/SimpleITK/SimpleITK/blob/master/Examples/N4BiasFieldCorrection/Documentation.rst

The inputImage, the log_bias_field, and the corrected_image_full_resolution should be displayed. The bias field may need a color map.

mbopfNIH:
Resolves #1873

### DIFF
--- a/Examples/N4BiasFieldCorrection/Documentation.rst
+++ b/Examples/N4BiasFieldCorrection/Documentation.rst
@@ -37,6 +37,55 @@ the limit of iterations per resolution level, thereby setting both the
 iterations and the number of scales ( from the length of the array
 ). The output of the filter is the bias corrected input image.
 
+Example Run
+-----------
+Running the Python code with the following inputs:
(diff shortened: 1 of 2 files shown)


mbopfNIH:
The input image isn't a very good example case for this algorithm, at least not with the defaults. We should find a better input that can benefit from the correction.

I was able to add Ziv's suggestion for the ..list-table and I think it looks pretty good!
Screen Shot 2023-02-08 at 6 53 04 PM

mbopfNIH:
I displayed the default mask, but can remove it if you prefer. I don't think the Log Bias Image needs a color map.

blowekamp (commented):
LGTM

dave3d (approved):
Very nice.

PR #1886, opened 2023-02-15
Use exec_code configuration for paths

blowekamp:
Make the working directory the same for exec_code and plot, makes the input files relative to the same path.

Remove unneed imports.
Change working directory to import an Example instead of modifying the python path.


mbopfNIH (approved):
At one point I tried changing the `exec_code_working_dir` but it didn't seem to have any effect. But I wouldn't have changed it to '.' since I assumed that was the default. This is a cleaner solution.

zivy (approved):

dave3d (approved):

PR #1899, opened 2023-03-01
BUG: Fixed incorrect tag usage.

zivy:
The Study Instance UID tag was given as 0020|000D and should be 0020|000d. The Series Description tag, 0008|103e, is optional and may not exist. Changes address both issues.


blowekamp:
The DICOM tags are case sensitive? Is this the correct behavior? Are other DICOM packages tags case sensitive? Is the behavior documented?

zivy:
The tags are hexadecimal numbers so they should not be case sensitive when using letters (f or F both represent 15). 

In ITK they are strings and then [converted to hexadecimal numbers](https://github.com/InsightSoftwareConsortium/ITK/blob/208902e16a9103d486c7078a747d04044b780441/Modules/IO/GDCM/src/itkGDCMImageIO.cxx#L868) via GDCM. The GDCM conversion function [looks good](https://github.com/malaterre/GDCM/blob/f68ecaaf3e0769a965c2fa7caa4b2efe877cf3b9/Source/DataStructureAndEncodingDefinition/gdcmTag.cxx#L45.), case insensitive. Not sure what is going on in ITK.

dave3d (approved):

PR #1930, opened 2023-05-17
Add Casting for vector pixel types to scalar image with dimensions +1

blowekamp:
Change the CastImageFilter to support converting between scalar images and vector images, where the dimension is changed. The component is added/removed from the fastest dimensions. The Image buffer remains the same between the two images, but it currently is copied.


Code/Common/include/sitkImageConvert.h
```diff
@@ -31,7 +31,9 @@ namespace simple
 {
 
 
-/** \brief A utility method to help convert between itk image types efficiently.
+/** \brief A utility methods to  convert between itk image types efficientl by
```
dave3d, line 34: efficiently

dave3d (approved):

blowekamp:
Considering #1936 as an alternative or in addition to this feature...

PR #1936, opened 2023-05-25
Adding Image::ToVector and Image::ToScalar

blowekamp:
Add new method to efficiently convert between scalar and vector images, where the vector component is converted to a image dimension. This enable conversion of images when the fastest axis is no used for channels and other operations dones on a per-channel basis.

### DIFF
--- a/Code/Common/include/sitkImage.h
+++ b/Code/Common/include/sitkImage.h
@@ -371,6 +371,38 @@ namespace simple
 
     std::string ToString( ) const;
 
+    /** \brief Convert the first dimension to the components for image with vector pixel type.
+     *
+     * This method will convert a scalar image to a vector image with
+     * the number of components equal to the size of the first
+     * dimension. If the image is already a vector image then the
+     * image is returned.
+     *
+     * An exception is thrown if the image is 2D or if the pixel type is a label or complex pixel type.
+     *
+     * \param inPlace If true then the image is made unique and converted in place updating this image,
+     * otherwise a copy of the image is made and returned.
+     *
+     * \sa ToScalarImage
+     */
+    Image ToVectorImage(bool inPlace = true);
+
+    /** \brief Convert a image of vector pixel type to a scalar image with N+1 dimensions.
+     *
+     * This method will convert a vector image to a scalar image with
+     * the size of the first dimension equal to the number of
+     * components. If the image is already a scalar image then the
+     * image is returned.
+     *
+     * An exception is thrown if the image is has SITK_MAX_DIMENSION dimensions or if the pixel type is a label or complex pixel type.
+     *
+     * \param inPlace If true then the image is made unique and converted in place updating this image,
+     * otherwise a copy of the image is made and returned.
+     *
+     * \sa ToVectorImage
+     */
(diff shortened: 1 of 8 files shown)


CMake/sitkUseGTest.cmake
```diff
@@ -18,7 +18,7 @@ include(GoogleTest)
 
 if (TARGET GTest::Main AND TARGET GTest::GTest)
     message(STATUS "GTest targets already defined.")
-#    return()
+    return()
```
blowekamp, line 21: This change should not be part of this PR.

Code/Common/include/sitkImage.h
```diff
@@ -371,6 +371,38 @@ namespace simple
 
     std::string ToString( ) const;
 
+    /** \brief Convert the a first dimension to the components for image with vector pixel type.
```
zivy, line 374: "the a" -> "the"

Code/Common/include/sitkImageConvert.h
```diff
@@ -31,7 +31,9 @@ namespace simple
 {
 
 
-/** \brief A utility method to help convert between itk image types efficiently.
+/** \brief A utility methods to  convert between itk image types efficiently by
```
zivy, line 34: "methods" -> "method"

Code/Common/include/sitkImage.h
```diff
@@ -371,6 +371,38 @@ namespace simple
...
+     *
+     * This method will convert a scalar image to a vector image with
+     * the number of components equal to the size of the first
+     * dimension. If the image is already a vector image then the
+     * image is returned.
```
zivy, line 379: Should it also say something about loosing the metadata information for the first axis? What I mean is that the operation is lossy with respect to metadata: `scalarImage.ToVectorImage().ToScalarImage() + scalarImage` will throw an exception because the metadata is different.

zivy (approved):
LGTM. Some minor comments with respect to documentation.

Code/Common/include/sitkImage.h
blowekamp: This is a more complicated issues than just documentation. The spacing and origin have a straight forward way to drop  a dimension and prepend a 1 or 0 respectfully when adding a dimensions. However the Direction matrix can have some degenerative cases when the extract sub-matrix is not appropriate.

The most similar thing that exists in ITK is the ExtractImageFilter, which has multiple direction collapse strategies and not working default.

zivy: A possible punting approach, which I think is valid, is to limit these operations to images with identity direction cosine matrices. I suspect this will address all relevant use cases. 

Extend the capabilities of these functions to non identity cosine matrices only if the need arises. Add the functionality only when there is a concrete use case.

blowekamp: >A possible punting approach, which I think is valid, is to limit these operations to images with identity direction cosine matrices. I suspect this will address all relevant use cases.

Perhaps just the "component" dimension needs to be identity. This would follow the intended functionality of converting to and from vector images.

zivy (approved):

PR #1941, opened 2023-06-13
Test out setting OSX_ARCHITECTURES for arm64

blowekamp:


blowekamp:
Parts of the ITK build require running compiled executable, unfortunately intel Macs can not run ARM code with Rosetta.

PR #1946, opened 2023-06-16
Add win32 build in batch

blowekamp:
This will replace the AZP batch build. The win32 build is interesting because the floating point operations are different that x64 intel.


zivy (commented):
Is there a need for using Win32 build given that with [release 2.0](https://github.com/SimpleITK/SimpleITK/releases/tag/v2.0rc1) we stopped providing 32 bit binaries? What am I missing?

blowekamp:
The usefulness of the x86 instruction set is that it has a different set of FPU instructions which may give different computation results, and help with setting reasonable tolerances to tests.

zivy (approved):

blowekamp:
win32 build are no longer of interest

PR #1950, opened 2023-06-21
GHA Bactch Build use self hosted images first

blowekamp:
Swtich some build combinations to use the self-hosted images, and put them first since they run the quickest.


jwj019 (approved):
LGTM! There does seem to be an issue with a test on the x64 runner for some reason.  Do you know what is causing that?

zivy:
The failed test is for the `LaplacianSharpeningImageFilter` which was modified recently, so likely due to that.

PR #1959, opened 2023-07-12
Testing the sitk.ImageSeriesReader_GetGDCMSeriesIDs method.

zivy:
Test that the method indeed returns the series that are in a directory.

### DIFF
--- a/.github/workflows/additional_dictionary.txt
+++ b/.github/workflows/additional_dictionary.txt
@@ -1,3 +1,5 @@
+UIDs
+unittest
 API
 AUL
 AULL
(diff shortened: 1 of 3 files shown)


zivy:
In SWIG 4.0.1, the underscore approach to referencing static class methods is discouraged (requires usage of -flatstaticmethod option). Test code was changed to use the dot access.

blowekamp:
This PR no longer tests the existing expected behavior.

blowekamp:
It looks like there are updates to the dictionary that likely were not intentional.

blowekamp (approved):

PR #1960, opened 2023-07-13
Restore functions for static members like ImageSeriesReader_GetGDCMSe…

blowekamp:
…riesIDs

SWIG 4.1.0 change behavior to removing flattened static methods for objects. The compatible behavior can be restored by adding `-flatstaticmethod` argument to SWIG. The
SimpleITK_PYTHON_FLATSTATICMETHOD CMake variable has been added to control the usage of this flag. It is currently enabled by default and is planned to default to OFF in future releases.

### DIFF
--- a/Wrapping/Python/CMakeLists.txt
+++ b/Wrapping/Python/CMakeLists.txt
@@ -17,6 +17,8 @@ include_directories ( ${SimpleITK_INCLUDE_DIRS} ${CMAKE_CURRENT_SOURCE_DIR} )
 #
 # Options
 #
+option ( SimpleITK_PYTHON_FLATSTATICMETHOD "Enable object_method functions for static class members. \
+This feature will be removed in the future. To disable requires SWIG>=4.1.0" ON )
 option ( SimpleITK_PYTHON_THREADS "Enable threaded python usage by unlocking the GIL." ON )
 mark_as_advanced( SimpleITK_PYTHON_THREADS )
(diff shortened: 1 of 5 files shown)


blowekamp:
@dzenanz @thewtex FYI, this change in behavior with SWIG 4.1.0 may be of interest.

dzenanz:
Thanks for the notice. I don't have time to deal with it now, so I created an issue to keep track of it: https://github.com/InsightSoftwareConsortium/ITK/issues/4106.

Wrapping/Python/CMakeLists.txt
```diff
@@ -17,6 +17,8 @@ include_directories ( ${SimpleITK_INCLUDE_DIRS} ${CMAKE_CURRENT_SOURCE_DIR} )
 #
 # Options
 #
+option ( SimpleITK_PYTHON_FLATSTATICMETHOD "Enable object_method functions for static class members. \
+This feature will be removed in the future. To disable requires SWIG>=4.1.0" ON )
```
zivy, line 21: Possibly add an example of a flat static method to the help text so that others understand what this means. Change to :"... for static class members (e.g. ImageSeriesReader_GetGDCMSeriesIDs)."

zivy (approved):
Minor comment on CMake help message, not sure if change is needed or not.

dave3d (approved):

Wrapping/Python/CMakeLists.txt
blowekamp: The "object_method" was intended to be an example.

PR #1970, opened 2023-08-02
BUG: Renamed CommentSpellCheck repo

dave3d:


dave3d:
> Will the new tag work?

Dunno?

.github/workflows/PythonLintAndSpell.yml
```diff
@@ -42,18 +42,18 @@ jobs:
...
       uses: actions/setup-python@v4
       with:
-        python-version: 3.8
+        python-version: 3.9
 
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
-        python -m pip install git+https://github.com/SimpleITK/SimpleITKSpellChecking@v0.1
+        python -m pip install git+https://github.com/SimpleITK/CommentSpellCheck
```
blowekamp, line 53: Is there a new tag that we could pin?

dave3d: Yeah, it's currently at release v0.2.0.  I can pin it to that.

blowekamp:
Looks good. Can we squash some of those extra commits before merge?

blowekamp (approved):
Thanks for the update!

PR #1986, opened 2023-09-08
add sitkLabelLinear based on LabelImageGenericInterpolateImageFunction

dyollb:
blindly adding sitkLabelLinear where sitkLabelGaussian was found (except in ExpandImageFilter.json).

### DIFF
--- a/Code/Common/include/sitkCreateInterpolator.hxx
+++ b/Code/Common/include/sitkCreateInterpolator.hxx
@@ -20,13 +20,16 @@
 
 
 #include "sitkInterpolator.h"
-#include <itkNearestNeighborInterpolateImageFunction.h>
-#include <itkLinearInterpolateImageFunction.h>
 #include <itkBSplineInterpolateImageFunction.h>
+#include <itkBSplineResampleImageFunction.h>
 #include <itkGaussianInterpolateImageFunction.h>
 #include <itkLabelImageGaussianInterpolateImageFunction.h>
+#ifdef SITK_GENERIC_LABEL_INTERPOLATOR
+#include <itkLabelImageGenericInterpolateImageFunction.h>
+#endif
+#include <itkLinearInterpolateImageFunction.h>
+#include <itkNearestNeighborInterpolateImageFunction.h>
 #include <itkWindowedSincInterpolateImageFunction.h>
-#include <itkBSplineResampleImageFunction.h>
 
 namespace itk
 {
@@ -183,6 +186,16 @@ CreateInterpolator(const TImageType *image, InterpolatorEnum itype) {
     p->SetSigma(sigma);
     p->SetAlpha(1.0);
     return RType(p);
+  }
+  case sitkLabelLinear: {
+#ifdef SITK_GENERIC_LABEL_INTERPOLATOR
+    using InterpolatorType = itk::LabelImageGenericInterpolateImageFunction<TImageType, itk::LinearInterpolateImageFunction, double>;
+
+    typename InterpolatorType::Pointer p = InterpolatorType::New();
+    return RType(p);
+#else
+      sitkExceptionMacro( "sitkLabelLinear is not supported in this build. Compile ITK with Module_GenericLabelInterpolator=ON." );
+#endif
   }
   case sitkHammingWindowedSinc: {
 
--- a/Code/Common/include/sitkInterpolator.h
+++ b/Code/Common/include/sitkInterpolator.h
@@ -80,6 +80,13 @@ enum InterpolatorEnum {
    */
   sitkLabelGaussian = 5,
 
+  /** \brief Linear interpolation for multi-label images
+   *
+   * \sa itk:LabelImageGenericInterpolateImageFunction
+   * \sa itk:LinearInterpolateImageFunction
+   */
+  sitkLabelLinear = 26,
+
   /** \brief Windowed sinc interpolation
    *
    * \f[ w(x) = 0.54 + 0.46 cos(\frac{\pi x}{m} ) \f]
--- a/Testing/Unit/sitkCommonTests.cxx
+++ b/Testing/Unit/sitkCommonTests.cxx
@@ -739,6 +739,9 @@ TEST( Interpolator, Test1 )
   ss.str("");
   ss << sitk::sitkBlackmanWindowedSinc;
   EXPECT_EQ("BlackmanWindowedSinc", ss.str());
+  ss.str("");
+  ss << sitk::sitkLabelLinear;
+  EXPECT_EQ("LabelLinear", ss.str());
 
 }
 
--- a/Code/Common/CMakeLists.txt
+++ b/Code/Common/CMakeLists.txt
@@ -7,6 +7,10 @@ add_subdirectory( src )
 set( SITK_INT64_PIXELIDS ${SimpleITK_INT64_PIXELIDS} )
 set( SITK_EXPLICIT_INSTANTIATION ${SimpleITK_EXPLICIT_INSTANTIATION} )
 set( SITK_USE_ELASTIX ${SimpleITK_USE_ELASTIX} )
+set( SITK_GENERIC_LABEL_INTERPOLATOR OFF)
+if (GenericLabelInterpolator IN_LIST ITK_MODULES_ENABLED)
+  set( SITK_GENERIC_LABEL_INTERPOLATOR ON)
+endif()
 
 include(sitkCheckTemplateDisambiguation)
 
--- a/Code/Common/src/CMakeLists.txt
+++ b/Code/Common/src/CMakeLists.txt
@@ -52,7 +52,6 @@ target_compile_options( SimpleITKCommon
     ${SimpleITK_PRIVATE_COMPILE_OPTIONS} )
 target_compile_features( SimpleITKCommon PUBLIC cxx_std_17 )
 
-
 sitk_install_exported_target( SimpleITKCommon )
 
 # Add custom command that will delete java files which need to be rebuilt when changes
--- a/Code/Common/src/sitkConfigure.h.in
+++ b/Code/Common/src/sitkConfigure.h.in
@@ -39,6 +39,8 @@
 
 #cmakedefine SITK_USE_ELASTIX
 
+#cmakedefine SITK_GENERIC_LABEL_INTERPOLATOR
+
 // Include ITK version reported in CMake with SITK prefix, so that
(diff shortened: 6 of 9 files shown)


dyollb:
@blowekamp I think my C++ (library) changes are correct and make sense.  I am not sure what the ExpandImageFilter.json is used for. Should I add something there, similar to [here](https://github.com/SimpleITK/SimpleITK/blob/master/Code/BasicFilters/json/ExpandImageFilter.json#L138)?

blowekamp:
The module will also need to be added here: https://github.com/SimpleITK/SimpleITK/blob/master/Code/Common/src/CMakeLists.txt#L34

The question we have is should this remote module be required to build SimpleITK or should it be an option? If an option, more logic will need to be added.

dyollb:
> The module will also need to be added here: https://github.com/SimpleITK/SimpleITK/blob/master/Code/Common/src/CMakeLists.txt#L34
> 
> The question we have is should this remote module be required to build SimpleITK or should it be an option? If an option, more logic will need to be added.

The module is only a single filter, so I personally I would vote for requiringit. My personal goal would be to install SImpleITK from PyPI and use the `sitkLabelLinear` option.

But you maintainers will decide. I remember there being a discussion which remote modules should be moved to ITK proper (@thewtex). Maybe this could be a target.

If I should make it optional, is there any example filter that I could model this after, or should I just define some macro which enables/disables the code?

blowekamp:
Yes, we can enable in the python distribution. But for other third party builds, that are against a system ITK or other ITK distribution I don't think we want this to be a deal breaker.

The other optional ITK modules in the SimpleITK Filter directory. The logic for detecting if the module is enabled in ITK is here: https://github.com/SimpleITK/SimpleITK/blob/master/Code/BasicFilters/src/CMakeLists.txt#L135C22-L135C22

For this case the only file which used that module is `[sitkCreateInterpolator.hxx](https://github.com/SimpleITK/SimpleITK/pull/1986#diff-9c773eceeb2e184b22faa0c661daae104e41cdef3291a34b01686992ca1b7f0e)` so I think a private "target_compile_definition" can be used to enable this code, otherwise an exception can be thrown.

gdevenyi:
> Yes, we can enable in the python distribution. But for other third party builds, that are against a system ITK or other ITK distribution I don't think we want this to be a deal breaker.

Maybe its time for this filter to be enabled by default? Its pretty useful and enabled by default in ANTs builds.

blowekamp:
> Maybe its time for this filter to be enabled by default? Its pretty useful and enabled by default in ANTs builds.

 ITK Remote modules are not enabled by default, the code would need to be moved into ITK proper. This can be discussed on ITK discourse or a ITK PR can be made to move the module into the ITK repository.

blowekamp:
@dyollb  you can find more detailed build logs/errors on dash: https://open.cdash.org/index.php?project=SimpleITK

dyollb:
@blowekamp I nearly have the logic ready for optionally compiling with `Module_GenericLabelInterpolator=ON`. 

- Would a user compiling using the SuperBuild strategy simply set the initial cache variable `Module_GenericLabelInterpolator=ON`  or do I need to add an option, e.g. `SIMPLEITK_USE_GENERIC_LABEL_INTERPOLATOR`?
- Where would you add this option for the CI to add the feature in python modules (on PyPI)?

blowekamp:
> @blowekamp I nearly have the logic ready for optionally compiling with `Module_GenericLabelInterpolator=ON`.
> 
> * Would a user compiling using the SuperBuild strategy simply set the initial cache variable `Module_GenericLabelInterpolator=ON`  or do I need to add an option, e.g. `SIMPLEITK_USE_GENERIC_LABEL_INTERPOLATOR`?
> * Where would you add this option for the CI to add the feature in python modules (on PyPI)?

The Superbuild can set `Module_GenericLabelInterpolator=ON` by default if it is not already set. The ITK configuration will control if SimpleITK uses this module/feature. No need for an additional option.

For example the condo-forge[ libitk ](https://github.com/conda-forge/libitk-feedstock) package currently does not have this module enabled. When SimpleITK is build against this library, this feature would be disabled. But if the feed stock is reconfigured to enable this module, then SimpleITK would enable this feature.

gdevenyi:
> For example the condo-forge libitk package currently does not have this module enabled. 

I think it does?

https://github.com/conda-forge/libitk-feedstock/blob/078ea8e5dc5e713eaa03a8bbf7731a8b049c6961/recipe/build.sh#L47

blowekamp:
Yes, it does appear to be already enable the the libitk feedstock. The comments were more an illustrative example of how the SimpleITK configuration should respond to the ITK configuration. As opposed a specific request for a change in that feedstock.

Code/Common/include/sitkCreateInterpolator.hxx
```diff
@@ -184,6 +187,16 @@ CreateInterpolator(const TImageType *image, InterpolatorEnum itype) {
     p->SetAlpha(1.0);
     return RType(p);
   }
+  case sitkLabelLinear: {
+#ifdef SITK_GENERIC_LABEL_INTERPOLATOR
+    using InterpolatorType = itk::LabelImageGenericInterpolateImageFunction<TImageType, itk::LinearInterpolateImageFunction, double>;
```
blowekamp, line 192: Does ANTs make available other interpolator functions beside Linear for use with this type?

dyollb: just linear: https://github.com/ANTsX/ANTs/blob/46ab4e7c161151732624b8662be8023999dbfe78/Examples/make_interpolator_snip.tmpl

dyollb: but `genericlabel[linear]` apparently generally works well: https://github.com/ANTsX/ANTs/issues/677

Code/Common/include/sitkInterpolator.h
```diff
@@ -80,6 +80,13 @@ enum InterpolatorEnum {
    */
   sitkLabelGaussian = 5,
 
+  /** \brief Smoothly interpolate multi-label images
+   *
+  /** \brief Linear interpolation for multi-label images
```
blowekamp, line 85: There is a mistake in the comments with two opening "/**"

Testing/Unit/sitkCommonTests.cxx
```diff
@@ -739,6 +739,8 @@ TEST( Interpolator, Test1 )
   ss.str("");
   ss << sitk::sitkBlackmanWindowedSinc;
   EXPECT_EQ("BlackmanWindowedSinc", ss.str());
+  ss << sitk::sitkLabelLinear;
```
blowekamp, line 742: Missing `ss.str("")'

Test failure:
https://open.cdash.org/test/1211771265
[ RUN      ] Interpolator.Test1
/home/circleci/SimpleITK/Testing/Unit/sitkCommonTests.cxx:743: Failure
Expected equality of these values:
  "LabelLinear"
  ss.str()
    Which is: "BlackmanWindowedSincLabelLinear"

blowekamp (changes requested):

Testing/Unit/sitkCommonTests.cxx
dyollb: thanks, I was still trying to setup a proper build (not SuperBuild), but I cmake doesn't like my lua executable.

Code/Common/include/sitkInterpolator.h
dyollb: fixed

Code/Common/include/sitkCreateInterpolator.hxx
gdevenyi: The original paper that introduced genericlabel showed linear was the best choice as well

dyollb:
I can reproduce the linker issue on my mac. But I get the same error for the master branch. Not sure where this is coming from.

[build-master-full.log](https://github.com/SimpleITK/SimpleITK/files/12565227/build-master.log)
[build-master-truncated.log](https://github.com/SimpleITK/SimpleITK/files/12565236/build-master-truncated.log)

dyollb:
I measured some execution times for a toy problem.

Resampling an image with 2 labels [0, 1] from 200 x 200 x 200 to 600 x 600 x 600 voxels:
| Method | Time [s] |
| --- | --- |
| sitkNearestNeighbor | 0.6 |
| sitkLabelLinear | 2.9 |
| sitkLabelGaussian | 70.0 |

Resampling an image with 20 labels [0, 20) from 200 x 200 x 200 to 600 x 600 x 600 voxels:
| Method | Time [s] |
| --- | --- |
| sitkNearestNeighbor | 0.52 |
| sitkLabelLinear | 22.5 |
| sitkLabelGaussian | 79.2 |

For a large number of labels, sitkLabelLinear starts getting slowish. But for many cases the speed-up is great.

The quality is also nice (and much less dependent on the target resolution, more on the source resolution):
| sitkNearestNeighbor | sitkLabelGaussian | sitkLabelLinear |
| --- | --- | --- |
|  nn_8 | gauss_8 | linear_8 |

dyollb:
> I can reproduce the [linker issue](https://open.cdash.org/viewBuildError.php?buildid=8976888) on my mac. But I get the same error for the master branch. Not sure where this is coming from.
> 
> [build-master-full.log](https://github.com/SimpleITK/SimpleITK/files/12565227/build-master.log) [build-master-truncated.log](https://github.com/SimpleITK/SimpleITK/files/12565236/build-master-truncated.log)

needed to merge upstream master

blowekamp:
Very nice.

We like to keep a clean history. So no merges of master into feature branches as rebases are preferend. Also please condense the history in to a few concise commits. The SimpleITK git history does not need to contain the evolution of this feature.

dyollb:
Different example (interpolating a thin line `np.eye(20)`):

| sitkNearestNeighbor | sitkLabelGaussian | sitkLabelLinear |
| --- | --- | --- |
| nn_8 | gauss_8 | linear_8 |

dyollb:
Hi @blowekamp , i implemented the changes you requested and squashed the commits to keep the history clean.

blowekamp (approved):
Great contribution! Everything looks in order.

If you have additional time, in a follow up PR it would be good to add a test which runs the interpolator and checks the results.

dave3d (approved):

zivy (approved):
Nice addition for label resampling, faster than sitkLabelGaussian and more accurate results.

PR #1994, opened 2023-09-26
Tidy modernize avoid bind

blowekamp:
Use clang-tidy with modernize-avoid-bind to migrate to lambda functions over std::bind. A few manual clean up were needed  along with some manual migration of code generation and a couple files.

### DIFF
--- a/Code/Registration/src/sitkImageRegistrationMethod_CreateOptimizer.cxx
+++ b/Code/Registration/src/sitkImageRegistrationMethod_CreateOptimizer.cxx
@@ -101,16 +101,16 @@ namespace simple
       optimizer->SetDoEstimateLearningRateOnce( this->m_OptimizerEstimateLearningRate==Once );
       optimizer->SetMaximumStepSizeInPhysicalUnits( this->m_OptimizerMaximumStepSizeInPhysicalUnits );
 
-      this->m_pfGetMetricValue = std::bind(&_OptimizerType::GetCurrentMetricValue,optimizer.GetPointer());
-      this->m_pfGetOptimizerIteration = std::bind(&CurrentIterationCustomCast::CustomCast,optimizer.GetPointer());
-      this->m_pfGetOptimizerPosition = std::bind(&PositionOptimizerCustomCast::CustomCast,optimizer.GetPointer());
-      this->m_pfGetOptimizerLearningRate = std::bind(&_OptimizerType::GetLearningRate,optimizer.GetPointer());
-      this->m_pfGetOptimizerConvergenceValue = std::bind(&_OptimizerType::GetConvergenceValue,optimizer.GetPointer());
+      this->m_pfGetMetricValue = [capture0 = optimizer.GetPointer()] { return capture0->GetCurrentMetricValue(); };
+      this->m_pfGetOptimizerIteration = [capture0 = optimizer.GetPointer()] { return CurrentIterationCustomCast::CustomCast(capture0); };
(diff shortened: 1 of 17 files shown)


blowekamp:
@N-Dekker You may find this migration of std::bind to lambda interesting.

dave3d (approved):
Fancy!

zivy (approved):

Code/Registration/src/sitkImageRegistrationMethod_CreateOptimizer.cxx
```diff
@@ -157,16 +157,16 @@ namespace simple
       optimizer->SetDoEstimateLearningRateOnce( this->m_OptimizerEstimateLearningRate==Once );
       optimizer->SetMaximumStepSizeInPhysicalUnits( this->m_OptimizerMaximumStepSizeInPhysicalUnits );
 
-      this->m_pfGetMetricValue = std::bind(&_OptimizerType::GetCurrentMetricValue,optimizer.GetPointer());
-      this->m_pfGetOptimizerIteration = std::bind(&CurrentIterationCustomCast::CustomCast,optimizer.GetPointer());
-      this->m_pfGetOptimizerPosition = std::bind(&PositionOptimizerCustomCast::CustomCast,optimizer.GetPointer());
-      this->m_pfGetOptimizerLearningRate = std::bind(&_OptimizerType::GetLearningRate,optimizer.GetPointer());
-      this->m_pfGetOptimizerConvergenceValue = std::bind(&_OptimizerType::GetConvergenceValue,optimizer.GetPointer());
+      this->m_pfGetMetricValue = [capture0 = optimizer.GetPointer()] { return capture0->GetCurrentMetricValue(); };
```
N-Dekker, line 160: Is the `capture0` really useful? I mean, why not just capturing ` optimizer`, as in:

    this->m_pfGetMetricValue = [optimizer] { return optimizer->GetCurrentMetricValue(); };

?

blowekamp: Good question. What we want to avoid is a copy of the "optimizer" variable as this would add an additional reference to the ITK object. But the the "optimizer" variable is local to the function so it has improper duration. The itk::SmartPointer which owns the object at class scope level is of a base class type, so that does not work either. Hence, a why a raw pointer is used.

N-Dekker: OK, so I understand you intended to avoid extra reference-counting! If you _know_ that `optimizer` is not NULL, you might still consider passing it by (C++) reference:

    this->m_pfGetMetricValue = [&optimizerRef = *optimizer] { return optimizerRef.GetCurrentMetricValue(); };

Anyway, just a suggestion. My 2 cents  :-)

N-Dekker: I'm just suggesting to capture `*optimizer` by reference, instead of capturing `optimizer->GetPointer()`. `optimizer` is a SmartPointer, so `*optimizer` returns the reference to the object that it has inside. Just like `optimizer->GetPointer()` returning a (raw) pointer to  the object that it has inside.

blowekamp: Yea, using a reference to a object derived from `itk::Object` is rather atypical for ITK. The code like `itk_instance.GetSomething()` seems odd itk stylisticly.

PR #1998, opened 2023-09-27
Apply clang tidy modernize-concat-nested-namespaces

blowekamp:
And manually update expand templates.

### DIFF
--- a/Code/BasicFilters/src/sitkCastImageFilter-3.cxx
+++ b/Code/BasicFilters/src/sitkCastImageFilter-3.cxx
@@ -19,9 +19,7 @@
 #include "sitkCastImageFilter.hxx"
 
 
-namespace itk
-{
-namespace simple
+namespace itk::simple
 {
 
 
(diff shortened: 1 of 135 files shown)


zivy (approved):

Code/BasicFilters/src/sitkCastImageFilter-3.cxx
```diff
@@ -38,5 +36,4 @@ void CastImageFilter::RegisterMemberFactory3()
 
 }
 
-} // end namespace simple
 } // end namespace itk
```
dave3d, line 39: Should these end comments be changed to itk::simple?  And others to itk::simple::detail?

dave3d (approved):

Code/BasicFilters/src/sitkCastImageFilter-3.cxx
blowekamp: Good check. I think the comments were needed before to keep track of the nested '}', so I think I'll remove them... *invoking sed*....

PR #2002, opened 2023-09-28
Add Nightly Action for coverage

blowekamp:


zivy (approved):
Phrasing is a bit weird. The mix of daily and nightly. File is "Daily.yml", cron is set to run in the morning so daily makes sense but the workflow name is "Nightly Build".  I believe "nightly" is the ITK tradition.

Doesn't matter in terms of functionality but possibly consider renaming things for consistency.

jwj019 (approved):

PR #2003, opened 2023-09-29
Update Elastix with clang-tidy

blowekamp:
Applied modernize-make-unique, modernize-concat-nested-namespaces, and modernize-avoid-bind.

Note this was done by setting
  CMAKE_CXX_CLANG_TIDY:STRING=/usr/bin/clang-tidy-14;-checks=-*,modernize-make-unique,modernize-concat-nested-namespaces,modernize-avoid-bind;--header-filter=Code/;-fix

### DIFF
--- a/Code/ElastixTransformixWrappers/include/sitkElastixImageFilter.h
+++ b/Code/ElastixTransformixWrappers/include/sitkElastixImageFilter.h
@@ -27,8 +27,7 @@
 #include <string>
 #include <vector>
 
-namespace itk {
-  namespace simple {
+namespace itk::simple {
 
 /** \class ElastixImageFilter
 * \brief The class that wraps the elastix registration library.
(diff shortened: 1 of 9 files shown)


N-Dekker (approved):
`modernize-avoid-bind` didn't make any changes here, right? Then why do you mention this option in the commit message? Approved anyway, thanks!

blowekamp:
It was on the command line options when I ran it.

N-Dekker:
> It was on the command line options when I ran it.

Someone must have put it on the command line, then! Right? While apparently it didn't do anything! 🤔 

By the way, it appears that `modernize-avoid-bind` does not automatically replace each `std::bind` call. This one had to be done "by hand":
- https://github.com/InsightSoftwareConsortium/ITK/pull/4232

Does that look fine to you as well?

blowekamp:
Yes, That look good. I had a case where I had to add the return type of the lambda here:
https://github.com/SimpleITK/SimpleITK/blob/master/Code/Common/include/sitkMemberFunctionFactoryBase.h#L189


There is probably a large refactor which could be done in the member functions factory classes to further modernize/improve the classes.

PR #2005, opened 2023-10-03
Try new  macos-latest-xlarge M1 runner

blowekamp:


blowekamp:
GHA M1 agents are pay only.

PR #2017, opened 2023-10-11
DOC: Get CommentSpellCheck from PyPI

dave3d:
Get the spell checking package from PyPI instead of Github, and update to v0.2.1.

### DIFF
--- a/.github/workflows/PythonLintAndSpell.yml
+++ b/.github/workflows/PythonLintAndSpell.yml
@@ -50,7 +50,7 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
-        python -m pip install git+https://github.com/SimpleITK/CommentSpellCheck@v0.2.0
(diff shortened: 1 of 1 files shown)


zivy (commented):
When installed via pip, the instructions with respect to standard running `python comment_spell_check.py...` listed on pypi don't work.  I believe the command is `python -m comment_spell_check...` (sans the `.py`). When I ran this with `--version` the result was `unknown`. Possibly need for updating the docs or code or both? Possibly I'm missing something.

dave3d:
> When installed via pip, the instructions with respect to standard running `python comment_spell_check.py...` listed on pypi don't work. I believe the command is `python -m comment_spell_check...` (sans the `.py`). When I ran this with `--version` the result was `unknown`. Possibly need for updating the docs or code or both? Possibly I'm missing something.

Oh, you're right.

blowekamp (approved):

PR #2018, opened 2023-10-11
Add nightly ITK coverage valgrind job

blowekamp:


jhlegarreta:
> @dzenanz @jhlegarreta @thewtex I have added running the ITK nightly valgrind and coverage to SimpleITK GHA. It utilizes self-hosted runners to perform these actions. The will replace the "AWS BATCH" hosts on the ITK dashboard.

OK. Not sure why the checks belonging to ITK are added here instead of in the ITK repository. I guess you already thought about it.

I am not seeing the reports being submitted to CDash. Will we get them back, maybe upon merging the PRs?

> The valgrind process was taking too long. So the tests with the label "RUNS_LONG" are excluded from the valgrind. Alternatively the debug -O0 flag could be changed to -O1 but the reported lines of problems may be less accurate due to optimizations.

Did it timeout? If not, maybe two different jobs could be run so that the longer ones get also tested and if we know that the long tests only dwell (for now) in the default modules, turn off all unnecessary CMake flags?

Thanks for the effort @blowekamp.

jhlegarreta:
> They are run here because the self-hosted agents are available in this repo and not the ITK repo. Not ideal, but what we can do. The community can still see the configuration and update the configuration as needed.

OK.

> Here is the coverage report: https://open.cdash.org/index.php?project=Insight&date=2023-10-11

Had seen the build a few days ago and sent a message to the ITK discourse (staff): coverage has dropped importantly (90.11% to 89.39%).

> With all the test it ran for 10hrs and timed out. These tests run too long and there is little benefit to running them.

OK.

zivy (approved):

dave3d (approved):

PR #2033, opened 2023-11-20
DOC: Update documentation with Python version support policy.

zivy:


blowekamp:
I don't think under building is the place to describe binary packing policies.

zivy:
[Building](https://simpleitk.readthedocs.io/en/master/building.html) was the closest relevant page I could identify. Possibly under the [Developer page](https://simpleitk.readthedocs.io/en/master/FilterTemplates.html#developer) but I don't think that is a page users will read. If you think that is the more appropriate location, then I'll move the text there.

dave3d (approved):

PR #2044, opened 2024-01-29
BUG: Refactor ImageViewer class to fix initialization problem.

dave3d:


blowekamp:
Please follow the singleton design pattern in the answer here:
https://stackoverflow.com/questions/1008019/how-do-you-implement-the-singleton-design-pattern

Also please add a test for this specific issue, C++ and python.

Code/IO/src/sitkImageViewer.cxx
```diff
@@ -84,28 +84,43 @@ namespace
   void ExecuteCommand( const std::vector<std::string> & cmdLine, const unsigned int timeout=500 );
 }
 
+class GlobalConfig {
```
blowekamp, line 87: Place this class in the anonymous namespace too.

Code/IO/src/sitkImageViewer.cxx
```diff
@@ -152,66 +167,66 @@ void ImageViewer::initializeDefaults()
...
 
-  m_GlobalDefaultSearchPath.emplace_back("/opt/" );
-  m_GlobalDefaultSearchPath.emplace_back("/usr/local/" );
+  m_DefaultSearchPath.emplace_back("/opt/" );
+  m_DefaultSearchPath.emplace_back("/usr/local/" );
 
 #endif
 
-  localDebugMacro( << "Default search path: " << m_GlobalDefaultSearchPath << std::endl );
+  //localDebugMacro( << "Default search path: " << m_DefaultSearchPath << std::endl );
```
blowekamp, line 206: Maybe add logic to the getInstance which prints the default configuration the first time its called in debug mode.

Code/IO/src/sitkImageViewer.cxx
```diff
@@ -220,109 +235,119 @@ void ImageViewer::initializeDefaults()
 //
 ImageViewer::ImageViewer()
```
blowekamp, line 236: Mode declaration down after all GlobalConfig methods are defined.

blowekamp (commented):
Looks good. Some minor clean up comments. Also you may want to add class level documentation to this new class.

Code/IO/src/sitkImageViewer.cxx
dave3d: No, that can't happen.  Debug mode can't happen the first time getInstance is called.  At the start debug is false.  To set it to true, you'd have to call SetGlobalDefaultDebug, which calls getInstance already.  I'll make it so if you call SetGlobalDefaultDebug with true, it prints config.

Code/IO/src/sitkImageViewer.cxx
```diff
@@ -152,107 +173,96 @@ void ImageViewer::initializeDefaults()
...
-  m_GlobalAreDefaultsInitialized = true;
+  m_DefaultApplication = FindViewingApplication();
   }
 
-
-//
-// Constructor
-//
-ImageViewer::ImageViewer()
+void GlobalConfig::PrintConfig()
```
blowekamp, line 234: consider changing to operator<<

Code/IO/src/sitkImageViewer.cxx
```diff
@@ -152,107 +169,127 @@ void ImageViewer::initializeDefaults()
...
+  if (first_time)
     {
-    initializeDefaults();
+    if (instance.m_DefaultDebug)
+      {
+      std::ostringstream msg;
+      msg << instance;
+      ::itk::OutputWindowDisplayDebugText( msg.str().c_str() );
+      }
+    first_time = false;
```
blowekamp, line 243: Please use std::call_once with a lambda to avoid a race condition with first_time.

blowekamp:
@dave3d Can you please squash that last commit into the previous? Then I think the PR is ready to go!

PR #2048, opened 2024-02-07
DOC: DICOM meta-data tags are lower case.

### LINKED ISSUE #2047: simpleITK python acts differently while reading metadata with upper or lower case hexadecimal label
**Describe the bug**
I can only read series description ('0008|103e' or '0008|103E') with lower case ('0008|103e').

**To Reproduce**

1. Operating system, version, and architecture
 - OS: Windows 11 PRO 21H2
 - Architecture: Intel x86_64
2. Programming language, and version: Conda Python 3.8.17
3. Version of SimpleITK [e.g. 1.1.0], output of the SimpleITK Version() function:  2.3.1 (ITK 5.3)
4. How was SimpleITK installed?
 - [x] binary distribution [e.g. python -m pip install SimpleITK]

5. A minimal working example which causes the error.
```Python
import SimpleITK as sitk

def read_dicom_series_description(dicom_path, key):
    reader = sitk.ImageFileReader()
    reader.SetFileName(dicom_path)
    reader.LoadPrivateTagsOn()
    reader.ReadImageInformation()
    
    if reader.HasMetaDataKey(key):
        series_description = reader.GetMetaData(key)
    else:
        series_description = 'NA'
    
    return series_description

dicom_path = r'path\to\your\dicom'
key = "0008|103E"
print(f"Series Description with {key}:", read_dicom_series_description(dicom_path, key))
key = "0008|103e"
print(f"Series Description with {key}:", read_dicom_series_description(dicom_path, key))
print(sitk.Version())
```
**Expected behavior**

> Series Description with 0008|103E: 3 Plane Localizer 
> Series Description with 0008|103e: 3 Plane Localizer 
> SimpleITK Version: 2.3.1 (ITK 5.3)
> Compiled: Nov  1 2023 12:40:43

**Unexpected behavior**

> Series Description with 0008|103E: NA
> Series Description with 0008|103e: 3 Plane Localizer 
> SimpleITK Version: 2.3.1 (ITK 5.3)
> Compiled: Nov  1 2023 12:40:43

zivy:
DICOM tags are hexadecimal numbers so there is no difference between upper and lower case representations. The ITK dictionary is string based, so there is a difference and the tags are in lower case. Updated the documentation to point this out.

resolves #2047

### DIFF
--- a/Code/IO/include/sitkImageFileReader.h
+++ b/Code/IO/include/sitkImageFileReader.h
@@ -55,6 +55,18 @@ namespace itk {
      * when multiple ImageIOs "can read" the file and the user wants
      * to select a specific IO (not the first).
      *
+     * \note DICOM tags are represented as strings in the meta-data
+     * dictionary(s), therefore "0020|000D" and "0020|000d" are
+     * different when accessing the tag value. This differs from
+     * the hexadecimal numbers they represent, 0020|000D and 0020|000d
+     * are equivalent. The ITK meta-data dictionary is string based
+     * and uses lower case to represent the hexadecimal number read
+     * from disk, so 0020|000d will work as a key and 0020|000D will
+     * not be found in the dictionary (results in an exception if
+     * attempting to access).
+     * It is recommended to use lower case when setting and accessing
+     * DICOM tags.
+     *
      * \sa itk::simple::ReadImage for the procedural interface
      */
     class SITKIO_EXPORT ImageFileReader
--- a/Code/IO/include/sitkImageSeriesReader.h
+++ b/Code/IO/include/sitkImageSeriesReader.h
@@ -39,6 +39,18 @@ namespace itk::simple {
      * Once the image series is read the meta-data is directly
      * accessible from the reader.
      *
+     * \note DICOM tags are represented as strings in the meta-data
+     * dictionary(s), therefore "0020|000D" and "0020|000d" are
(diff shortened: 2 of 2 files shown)


dave3d (approved):

Code/IO/include/sitkImageFileReader.h
```diff
@@ -55,6 +55,12 @@ namespace itk {
      * when multiple ImageIOs "can read" the file and the user wants
      * to select a specific IO (not the first).
      *
+     * \note DICOM tags represent hexadecimal numbers,
```
blowekamp, line 58: Please begin with stating how things ARE. e.g. DICOM tags are index as strings in the meta-data dictionary(s), therefore "0020|000D" and "0020|000d" are different when accessing the tag.

When setting a tag does the capitalization matter?

zivy: When setting the tag the capitalization doesn't matter. Writing is done correctly no matter if you set the tag via "0020|000D" or "0020|000d". 

I would recommend to always use lower case, because of the inconsistency that occurs when setting the tag using upper case and then accessing it (it is upper case in the meta-data dictionary) vs. setting the tag using upper case, writing to disk, reading from disk and then accessing it (it will be lower case in the meta-data dictionary).

zivy: Language updated.

PR #2049, opened 2024-02-13
Updated github actions versions.

zivy:
Updated to the newer versions of github actions which use Node.js
20. Older versions used Node.js 16 which is deprecated.


blowekamp (approved):
Thanks for the update!

PR #2057, opened 2024-03-08
BUG: Example include a DICOM tag number with upper case letter.

zivy:
DICOM tags should be all in lower case letters to match the conversion done by the ITK reader.

### DIFF
--- a/Examples/DicomSeriesReadModifyWrite/DicomSeriesReadModifySeriesWrite.py
+++ b/Examples/DicomSeriesReadModifyWrite/DicomSeriesReadModifySeriesWrite.py
@@ -79,7 +79,7 @@ tags_to_copy = [
     "0010|0010",  # Patient Name
     "0010|0020",  # Patient ID
     "0010|0030",  # Patient Birth Date
-    "0020|000D",  # Study Instance UID, for machine consumption
+    "0020|000d",  # Study Instance UID, for machine consumption
     "0020|0010",  # Study ID, for human consumption
     "0008|0020",  # Study Date
     "0008|0030",  # Study Time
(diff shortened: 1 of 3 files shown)


Examples/DicomSeriesReadModifyWrite/DicomSeriesReadModifySeriesWrite.py
```diff
@@ -79,7 +79,7 @@
     "0010|0010",  # Patient Name
     "0010|0020",  # Patient ID
     "0010|0030",  # Patient Birth Date
-    "0020|000D",  # Study Instance UID, for machine consumption
+    "0020|000d",  # Study Instance UID, for machine consumption
```
blowekamp, line 82: On line 105 we are using `tolower`, which I guess is not technically needed any more.

Should the "Note" down below be promoted to the Documentation.rst?

zivy: Agreed on both counts. Will remove the `tolower` and *copy* the note to the Documentation.rst. Hopefully if it appears as is in enough places folks will be aware of the case sensitivity issue.

blowekamp: Maybe start the note with "Alway write DICOM tags in lower case hexadecimals."

zivy: Done and force pushed. Please merge once passes testing.

blowekamp (approved):
Thanks!

PR #2058, opened 2024-03-11
Update superbuild SWIG to 4.2.1

blowekamp:


blowekamp:
After running git bisect on SWIG, looks like SWIG/SWIG@64c9720a7ac8acbb54690c64e61d1b5abc8f3d29 is causing the following error for a few tests like " R.Example.DemonsRegistration2":
```
1100: Error in f(x, ...) : 
1100:   cannot find overloaded function for ProcessObject_AddCommand with argtypes (_p_itk__simple__FastSymmetricForcesDemonsRegistrationFilter, character, function)
1100: Calls: <Anonymous> -> f
```
@richardbeare  I am digging into understanding the issue, if you provide some assistance, it would be appreciated.

blowekamp:
Issue related to the `R.PixelIndexing` test failure is due to std::vector<uint64_t> and std::vector<int64_t> not being wrapped correctly. This issue has been reported to upstream swig/swig#2847

Wrapping/R/R.i
```diff
@@ -196,19 +196,27 @@ itk::simple::Image ArrayAsIm(SEXP arr,
...
 %extend itk::simple::ProcessObject {
   int AddCommand( itk::simple::EventEnum e, SEXP obj )
  {
    // make sure that the CommandCallable is in fact callable
    if (!Rf_isFunction(obj))
      {
-           sitkExceptionMacro(<<"R object is not a function, "
-                              <<"or it has not been set.");
+       void *argp3 = 0 ;
+       int res3 = SWIG_R_ConvertPtr(obj, &argp3, SWIGTYPE_p_itk__simple__Command,  0 );
```
blowekamp, line 212: Support for the direct sitk::Command class is not needed and does not need to be supported here.

zivy (approved):

PR #2063, opened 2024-03-11
Add pre-commit for black, yaml and json checks

blowekamp:


.pre-commit-config.yaml
```diff
@@ -0,0 +1,20 @@
...
+fail_fast: true
+
+repos:
+- repo: https://github.com/pre-commit/pre-commit-hooks
+  rev: v4.5.0
+  hooks:
+   - id: detect-private-key
+   - id: check-json
+   - id: check-yaml
+- repo: https://github.com/psf/black
```
zivy, line 16: Possibly add a pre-commit to preclude commits of large files (mistakenly committing an image instead of its hash file)? Not sure what size should be considered large, used 200 below as a magic number.
```
- id: check-added-large-files
     args: ['--maxkb=200']
```

dave3d (approved):

zivy (approved):
Minor suggestion for additional pre-commit setting.

.pre-commit-config.yaml
blowekamp: There are a couple large files *.I, that are above that limit. Currently there is some check with Kitware's GhostFlow checker and old local hooks that do some large file checks. These also have per file configuration in .gitattribute so that certain files can be allowed to be large. I don't see that with any built in pre-commit checks.

PR #2069, opened 2024-03-14
WIP: Adding script to setup precommit.

blowekamp:
Using a temporary repo to hold Kitware hooks with a configuration for pre-commit. Replaces the system to clone the hook repo/branch into the .git/hooks directory, which just using pre-commit.


blowekamp:
@bradking  Any interesting in adding a configuration file to Kitware's hooks to enable usage with [pre-commit](https://pre-commit.com) framework built on Python?

https://github.com/blowekamp/demo_kwhooks/blob/main/.pre-commit-hooks.yaml

bradking:
> Any interesting in adding a configuration file to Kitware's hooks

Please open an issue in our [GitSetup project](https://gitlab.kitware.com/utils/gitsetup/-/issues/) to propose that.

Utilities/GitSetup/setup-precommit
```diff
@@ -0,0 +1,55 @@
...
+if [ ! -d venv ]; then
+  echo "Setting up python venv..." &&
+  python3 -m venv venv
+fi
+) &&
+# activate the venv and install pre-commit with the min version in subshell
+(
+    echo "Setting up pre-commit..." &&
+    source ${git_dir}/hooks/venv/bin/activate &&
+    python -m pip install --disable-pip-version-check -U pre-commit>=$MIN_PRECOMMIT_VERSION &&
```
bradking, line 53: Some distros don't offer a plain `python` binary.  They offer `python3` (and `python2`).

bradking: ..or does `venv` always create a `python`?

.pre-commit-config.yaml
```diff
@@ -31,3 +31,10 @@ repos:
   hooks: # check conformance to black formatting
    - id: black
      args: ['--diff', '-v'] # if run without arguments, will fail and will format the files
+
+- repo: https://github.com/blowekamp/demo_kwhooks
+  rev: v0.1.0
+  hooks:
+    - id: legacy-pre-commit
+    - id: legacy-prepare-commit-msg
+    - id: legacy-commit-msg
```
bradking, line 40: I'd rather have the legacy/simple/bash hooks have a way to launch the python ones.  That way the weight of python is needed only when one wants to use it.

blowekamp: yes the kw hooks are nice and light weight, the direct installation can still be done with the additional of the yaml configuration file. 

I started adding pre-commit with installation on top of the kw hooks. The pre-commit installation step renamed existing kw hooks in .git/hook with the addition of '-legacy" to their names and chained to execute them.

I suppose something like `python -mpre_commit  run --hook-stage pre-commit` could be used or added .hook-config.bash.

zivy (approved):
Nice, detect issues as early as possible. pre-commit combined with the original KW hooks will keep us on the straight and narrow on the local machines before any PR hits the CI.

PR #2084, opened 2024-03-27
In R, remove overloaded AddCommand method

blowekamp:
SWIG 4.2 is unable to resolve the overloaded method taking SEXP, and sitk::Command.


blowekamp:
Breaking out prerequisite for #2058, also ensure changes are compatible with SWIG 4.1.

zivy (approved):
Nice use of closures.

PR #2094, opened 2024-04-12
Directly use fold expression in TypeListHasPixelIDValue

blowekamp:
The lambda calls are removed.

Also simplify the has_type expression.

### DIFF
--- a/Code/Common/include/Ancillary/type_list2.h
+++ b/Code/Common/include/Ancillary/type_list2.h
@@ -168,19 +168,11 @@ template <typename... Ts>
 struct visit<typelist<Ts...>>
 {
 
-  template <typename T, typename Predicate>
-  static int
-  _f(Predicate && visitor)
-  {
-    visitor.CLANG_TEMPLATE operator()<T>();
-    return 0;
-  }
-
   template <typename Predicate>
   void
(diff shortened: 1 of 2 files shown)


blowekamp:
@N-Dekker Thought you might find these fold expressions interesting.

zivy (approved):

Code/Common/include/Ancillary/type_list2.h
```diff
@@ -139,25 +139,10 @@ struct index_of<typelist<t0, Ts...>, T>
  */
 template <typename Typelist, typename T>
 struct has_type;
-#if !defined(_MSC_VER) || _MSC_VER >= 1930
-// note the following causes a syntax error with MS VS 16
 template <typename... Ts, typename T>
 struct has_type<typelist<Ts...>, T>
-  : std::integral_constant<bool, std::max<bool>(std::initializer_list<bool>{ std::is_same<Ts, T>::value... })>
+  : std::integral_constant<bool, ((std::is_same<Ts, T>::value ) || ...) >
```
N-Dekker, line 144: @blowekamp I find the original `std::max(std::initializer_list{...})` very clever 😺, but the fold expression is even better 👍

PR #2097, opened 2024-04-17
Wrapping new Image methods to check for matching image geometry.

blowekamp:


zivy (commented):
This makes sense in terms of making these methods available in SimpleITK, though I would hold off on merging it for now. 

I looked at the the [ITK implementation](https://github.com/InsightSoftwareConsortium/ITK/blob/318139a0415e34e702f7ace4fff6b20f3ac65a10/Modules/Core/Common/include/itkImageBase.hxx#L410-L413) and it is multiplying the `coordinateTolerance` with the spacing along axis 0 and using that as the tolerance. Not sure about this choice. It seems arbitrary as compared to the maximum or minimum of `GetSpacing`. More importantly, as a naive user I expect that the `coordinateTolerance` I specify is used as is. Possibly you have insight into the ITK rationale?

blowekamp:
@zivy It sounds like the ITK documentation should be improved. That can be copied back here after improvements are made and review in ITK. Can you please make a PR or open an issue in ITK?

Shall we proceed with the merge of this PR?

zivy (approved):
Please merge. Will create an issue on the ITK repository to discuss documentation and a possible PR.

PR #2107, opened 2024-04-26
Explicitly set csharp interpreter for mac packaging

blowekamp:


blowekamp:
The test are fixed by changing the VS version in #2114

PR #2112, opened 2024-04-29
Add creating draft GitHub releases for the lastest branch

blowekamp:
Create draft release from the latest branch when it is updated.

This does not yet replace the nightly trigger or build performed by Azure.


zivy (approved):
Possibly move the bash commands run in the "Create Release Notes" step into a script file and call it there? Potentially useful for more elaborate release notes in the future.

PR #2132, opened 2024-06-05
WIP: testing utf-8 MSVC flag

blowekamp:
https://learn.microsoft.com/en-us/cpp/build/reference/utf-8-set-source-and-executable-character-sets-to-utf-8?view=msvc-170


blowekamp:
This test did not detect when the uft8 string was in correctly converted to the filesystem encoding. It was a case of in correctly writing, and incorrectly reading the name would work.

Further experiments showed that the uff-8 flag would work when combined with a windows manifest: https://rap-ria.tbportals.niaid.nih.gov/swagger/#/TB%20or%20not%20TB/post%20TBorNotTB

PR #2137, opened 2024-06-24
Backport patches made in Slicer fork

blowekamp:

PR #2141, opened 2024-06-27
Address issue with refinding python in superbuild

blowekamp:
Alternative to  #2137
Adds Python version related variable to list of CMake variable propagated.

### DIFF
--- a/CMake/sitkLanguageOptions.cmake
+++ b/CMake/sitkLanguageOptions.cmake
@@ -149,6 +149,18 @@ if ( WRAP_PYTHON )
     Python_LIBRARY_DEBUG
     Python_LIBRARY_RELEASE
     )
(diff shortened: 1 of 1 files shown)


blowekamp:
@jcfr  This PR adds the Python version variable to the location where other Python variable are lists which gets propagated to CMake external projects. Also the other Python_\* variables listed in #2137 should already be in "SITK_LANGUAGES_VARS" and not needed to be added to SimpleITK_VARS.

However, I am not sure how to test if this would still fix the issues that Slicer had during building.

jcfr:
Thanks @blowekamp 

I will test this week and get back to you.

blowekamp:
I think this PR should be merged it may fix some Slicer build issues.

dave3d (approved):

PR #2145, opened 2024-07-05
Use debug build in CircleCI.

blowekamp:
This reduces build time so that a full build can occour in the aloted time.


blowekamp:
WIth no ccache, unable to build with in CircleCI allotted time.

PR #2147, opened 2024-07-05
DOC: Adding documentation for resampling.

zivy:
Providing an explanation of how resampling works, backward mapping as opposed to forward mapping.


zivy:
This PR requires data that will only be available after [this PR](https://github.com/SimpleITK/SimpleITKExternalData/pull/36) is merged into the SimpleITKExternalData repository.

dave3d:
If you rebase off the latest master, the CircleCi build-and-test ought to pass.

PR #2159, opened 2024-07-15
GHA Package print CMake configuration logs

blowekamp:


blowekamp:
PR only for debugging

PR #2160, opened 2024-07-15
Support non-annotated tags for git describe versioning

blowekamp:
Github Actions clobbers annotated tags with a tag at the triggered sha. Adding support to sitkSourceVersion for unannotated to support the GHA workflow.


blowekamp:
I believe this solution is good enough to merge and make the RC. 

However, the reason that GHA checkout globber the annotated tag is that it wants to ensure that the checkout out code and ref matches that of the tigger. That is if the tag has changed, the trigger tag and ref will still be used. However, for the case of making packages an error should be generated.

zivy (approved):

PR #2172, opened 2024-08-12
DOC: Change URLs to DOIs.

zivy:
The insight-journal now recommends using the DOIs to the papers instead of the old handles.


blowekamp (commented):
Great! Thank you!

blowekamp (approved):

PR #2174, opened 2024-08-13
WIP: Add VCPKG for dependencies management

cornking2020:
## What

- New Feature (Non-breaking change that adds functionality)

## Why

**Problem:** Managing C++ dependencies manually is time-consuming and error prone. Developers struggle with library versioning, compatibility issues, and complex build configurations.

**Solution:** Integrating [VCPKG](https://vcpkg.io/en/) simplifies dependency management by automating library acquisition and integration. It ensures consistent, reproducible builds across environments, reduces setup time for new developers, and allows easy updates of third-party libraries, improving overall project maintainability and productivity.

## How

Integrate VCPKG by adding configuration files and modifying `CMakeLists.txt` to enable dependency management.

## How to Test

Follow the official VCPKG documentation to install and set up VCPKG. Once installed, build the project using the same process as before.

<https://learn.microsoft.com/en-us/vcpkg/get_started/get-started>

## Additional Remarks

CMake options:

```shell
-G "Ninja Multi-Config"
-DBUILD_TESTING:BOOL=TRUE
-DBUILD_SHARED_LIBS:BOOL=ON
-DBUILD_DOXYGEN:BOOL=ON
-DCMAKE_TOOLCHAIN_FILE:PATH=/path/to/vcpkg/scripts/buildsystems/vcpkg.cmake
```

Tested on:

- Windows 11 ✔️
- Ubuntu 22.04 ✔️
-  Mac Studio M1 ✔️

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "vcpkg"]
+path = vcpkg
+url = https://github.com/microsoft/vcpkg.git
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,0 @@
-[submodule "vcpkg"]
-path = vcpkg
-url = https://github.com/microsoft/vcpkg.git
--- a/vcpkg.json
+++ b/vcpkg.json
@@ -1,0 +1,16 @@
+{
+  "$schema": "https://raw.githubusercontent.com/microsoft/vcpkg-tool/main/docs/vcpkg-schema-definitions.schema.json",
+  "name": "simple-itk",
+  "version-string": "2.4.0",
+  "license": "Apache-2.0",
+  "dependencies": [
+    {
+      "name": "itk",
+      "version>=": "5.4.0"
+    },
+    {
+      "name": "gtest",
+      "version>=": "1.15.0"
+    }
+  ]
+}
--- a/vcpkg.json
+++ b/vcpkg.json
@@ -1,6 +1,6 @@
 {
   "$schema": "https://raw.githubusercontent.com/microsoft/vcpkg-tool/main/docs/vcpkg-schema-definitions.schema.json",
-  "name": "simple-itk",
+  "name": "simpleitk",
   "version-string": "2.4.0",
   "license": "Apache-2.0",
   "dependencies": [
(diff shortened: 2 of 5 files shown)


vcpkg.json
```diff
@@ -0,0 +1,16 @@
+{
+  "$schema": "https://raw.githubusercontent.com/microsoft/vcpkg-tool/main/docs/vcpkg-schema-definitions.schema.json",
+  "name": "simple-itk",
```
blowekamp, line 3: We generally go by "SimpleITK" or "simpleitk"., without a hyphen.

vcpkg.json
```diff
@@ -0,0 +1,16 @@
...
+  "version-string": "2.4.0",
+  "license": "Apache-2.0",
+  "dependencies": [
+    {
+      "name": "itk",
+      "version>=": "5.4.0"
```
blowekamp, line 9: SimpleITK is relatively tightly coupled with itk, the version should match up to the minor release.

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "vcpkg"]
+path = vcpkg
+url = https://github.com/microsoft/vcpkg.git
```
blowekamp, line 3: I am not sure if this should be a submodule or a super build dependency. Not every build will need to have this resource. The GTest repo is another super build dependency that is only a clone and not complied code:
https://github.com/SimpleITK/SimpleITK/blob/master/SuperBuild/External_GTest.cmake

blowekamp (commented):
This is amazing! Thank you for the contribution.

I make a few initial comments and have not gone into how VCPKG works.

dave3d:
Very cool PR.  Thanks for doing this.

cornking2020:
Okay. I'll make some changes and add the documentation by the way.

vcpkg.json
cornking2020: Fixed

.gitmodules
cornking2020: Removed. Just follow the official VCPKG installation instructions. You don't need to add any Git submodules.

PR #2178, opened 2024-08-15
STYLE:  Fixes to the python examples for pylint

dave3d:


zivy (approved):
Looks good. 

Please make sure these examples are not referenced in the [documentation rst files](https://github.com/SimpleITK/SimpleITK/tree/master/docs/source) with line numbers that may have changed due to this PR and need to be updated.

blowekamp:
These are some old and unmaintained examples.

The "Examples/Python/Documentation.rst" does not look like it is incorporated into ReadTheDocs. Also adding the docs strings to the files may change how the examples should would be rendered.

blowekamp (approved):
These changes do look like an improvement.

PR #2181, opened 2024-08-16
Update docs

zivy:


docs/source/gettingStarted.rst
```diff
@@ -129,21 +129,27 @@ at :ref:`setup SimpleITK with Java <setup-java>`.
 R binary files
```
blowekamp, line 129: This heading seems misleading now.

blowekamp (approved):
Do you think that heading should be changed too?

docs/source/gettingStarted.rst
zivy: I don't think it's misleading. If someone is searching for R binaries they get to this location and it explicitly says there are none and gives instructions on how to easily build your own. Otherwise one may keep searching assuming they haven't found the instructions on where to download the binaries.

dave3d (approved):

PR #2182, opened 2024-08-16
Instruct on creating isolated conda-forge ecosystem

blowekamp:


zivy (approved):
Nice. 

I particularly like the `--override-channels` to ensure installation doesn't magically happen from another channel if the requested SimpleITK version is not on the conda-forge channel.

dave3d (approved):

blowekamp:
@zivy This issue is more that the installation will not work if python itself comes from the default channel: conda-forge/libitk-feedstock#98

PR #2183, opened 2024-08-16
Additional clarification on using conda-forge

blowekamp:
Further clarification about not mixing channels.

### DIFF
--- a/docs/source/gettingStarted.rst
+++ b/docs/source/gettingStarted.rst
@@ -79,19 +79,15 @@ SimpleITK Conda binaries are available for the conda-forge ecosystem. To get sta
 .. code-block :: bash
 
  conda create --name sitk python=3.11 simpleitk --channel conda-forge --override-channels
+ conda activate sitk
 
-This will create a new conda environment named ``sitk`` with Python 3.11 and SimpleITK installed from the `Anaconda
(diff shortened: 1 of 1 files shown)


docs/source/gettingStarted.rst
```diff
@@ -79,19 +79,15 @@ SimpleITK Conda binaries are available for the conda-forge ecosystem. To get sta
...
-To activate the environment, run:
-
-.. code-block :: bash
-
- conda activate sitk
-
-Conda-forge now recommends not mixing the `default` channel with the conda-forge channel, since the
-mixing different libc versions can cause conflicts. Alternatively, the
+The above ``create`` command creates an environment which only used the conda-forge channel. This follows Conda-forge's
+recommendation to not mixing the ``default`` channel with the conda-forge channel. Mixing the channels can cause package
```
zivy, line 89: "...to not mixing..." -> "...to not mix..."
"default" -> "defaults" (strange that they named the channel defaults and not default, but that is what is printed during installation).

zivy (approved):

docs/source/gettingStarted.rst
blowekamp: done.

PR #2193, opened 2024-10-01
BUG: Correct SITK_MAX_DIMENSION usage

blowekamp:
Fixes instantiation of 4D Elastix.

Addresses #2191.

### DIFF
--- a/Code/ElastixTransformixWrappers/src/sitkTransformixImageFilterImpl.cxx
+++ b/Code/ElastixTransformixWrappers/src/sitkTransformixImageFilterImpl.cxx
@@ -5,6 +5,7 @@
 #include "sitkCastImageFilter.h"
 #include "sitkImageConvert.h"
 #include "sitkInternalUtilities.h"
+#include "sitkImageConvert.hxx"
 
 namespace itk::simple
 {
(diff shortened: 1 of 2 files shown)


zivy (approved):
LGTM.

Strange that on GitHub it marks `SITK_MAX_DIMENSIONS` as deleted and `SITK_MAX_DIMENSION` as added and not just a deleted `S` (I think I am reading it correctly).

Code/ElastixTransformixWrappers/src/sitkTransformixImageFilterImpl.cxx
```diff
@@ -125,7 +126,7 @@ TransformixImageFilter::TransformixImageFilterImpl ::TransformixImageFilterImpl(
   this->m_MemberFactory->RegisterMemberFunctions<FloatPixelIDTypeList, 2>();
   this->m_MemberFactory->RegisterMemberFunctions<FloatPixelIDTypeList, 3>();
 
-#if SITK_MAX_DIMENSIONS >= 4
+#if SITK_MAX_DIMENSION >= 4
```
N-Dekker, line 129: - Looks like this bug was introduced with pull request #1892 right?

Anyway, "sitkConfigure.h.in" suggests `SITK_MAX_DIMENSION` without `S`, so that should be right  👍 
https://github.com/SimpleITK/SimpleITK/blob/2c9389c0b570cb40ddea7e7ab26f41846e475fe9/Code/Common/src/sitkConfigure.h.in#L30

N-Dekker (approved):
Thanks @blowekamp 👍

Code/ElastixTransformixWrappers/src/sitkTransformixImageFilterImpl.cxx
blowekamp: I am not sure 4D was working before that change either.

PR #2196, opened 2024-10-07
Use constexpr statements over enable_if dispatched functions

blowekamp:
The conventional if statements with "constexpr" is more readable than complicated enable_if statements. Additionally, the produced object files are smaller due to shorter symbols names.

### DIFF
--- a/Code/Common/include/sitkMemberFunctionFactoryBase.h
+++ b/Code/Common/include/sitkMemberFunctionFactoryBase.h
@@ -68,30 +68,14 @@ struct hash<std::pair<S, T>>
 template <class... TupleArgs>
 struct hash<std::tuple<TupleArgs...>>
 {
-private:
-  // recursive hashing of std::tuple from Sarang Baheti's blog
-  // https://www.variadic.xyz/2018/01/15/hashing-stdpair-and-stdtuple/
-  template <size_t Idx, typename... TupleTypes>
-  inline typename std::enable_if<Idx == sizeof...(TupleTypes), void>::type
-  hash_combine_tup(size_t &, const std::tuple<TupleTypes...> &) const
-  {}
-
-  template <size_t Idx, typename... TupleTypes>
-    inline typename std::enable_if < Idx<sizeof...(TupleTypes), void>::type
-                                     hash_combine_tup(size_t & seed, const std::tuple<TupleTypes...> & tup) const
-  {
-    hash_combine(seed, std::get<Idx>(tup));
-
-    //  on to next element
-    hash_combine_tup<Idx + 1>(seed, tup);
-  }
 
 public:
   size_t
   operator()(std::tuple<TupleArgs...> tupleValue) const
   {
     size_t seed = 0;
-    hash_combine_tup<0>(seed, tupleValue);
+    std::apply(
+      [&seed](auto... tupleElement) { (hash_combine(seed, tupleElement), ...); },tupleValue);
     return seed;
   }
 };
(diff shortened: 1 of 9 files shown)


zivy (approved):

Code/IO/include/sitkImportImageFilter.h
```diff
@@ -118,11 +118,7 @@ class SITKIO_EXPORT ImportImageFilter : public ImageReaderBase
   // components per pixel needs to be set, otherwise the method
   // does not exist. This is done with the enable if idiom.
   template <class TImageType>
-  typename std::enable_if<!IsVector<TImageType>::Value>::type
-  SetNumberOfComponentsOnImage(TImageType *)
-  {}
-  template <class TImageType>
-  typename std::enable_if<IsVector<TImageType>::Value>::type
+  void
   SetNumberOfComponentsOnImage(TImageType *);
```
N-Dekker, line 122: Looks like this member function declaration (`SetNumberOfComponentsOnImage`) may be removed as well!

Code/Common/include/sitkMemberFunctionFactoryBase.h
```diff
@@ -69,21 +69,16 @@ template <class... TupleArgs>
...
-    inline typename std::enable_if < Idx<sizeof...(TupleTypes), void>::type
-                                     hash_combine_tup(size_t & seed, const std::tuple<TupleTypes...> & tup) const
+  void
+  hash_combine_tup(size_t & seed, const std::tuple<TupleTypes...> & tup) const
   {
     hash_combine(seed, std::get<Idx>(tup));
 
-    //  on to next element
-    hash_combine_tup<Idx + 1>(seed, tup);
+    if constexpr (Idx + 1< sizeof...(TupleTypes))
```
blowekamp, line 78: @N-Dekker This was one of the most interesting replacements on a recursive variadic template function.

N-Dekker: @blowekamp Nice, but maybe it can even be further improved! Let me think...

N-Dekker: (hash_combine(seed, std::get<Idx>(tup), ...)

N-Dekker: Using a C++17 [fold expression](https://en.cppreference.com/w/cpp/language/fold) and [std::apply](https://en.cppreference.com/w/cpp/utility/apply)...?

```cpp
  operator()(std::tuple<TupleArgs...> tupleValue) const
  {
    size_t seed = 0;
    std::apply(
      [this, &seed](auto... tupleElement) { (hash_combine(seed, tupleElement), ...); },
      tupleValue);
    return seed;
  }
```

Disclaimer: uncompiled, untested code 🙀

blowekamp: Yes, I just stumble onto std::apply now too. However, I think just: `std::apply(hash_combine,tupleValue);` is what is needed 🤔

N-Dekker:
@blowekamp I'm sorry I don't have time to review the entire PR thoroughly, but you don't need to wait for me. I very much like its direction, and I'm glad I could help with a few suggestions 👍

Code/Common/include/sitkMemberFunctionFactoryBase.h
blowekamp: Thank you very much for this suggestion.

PR #2198, opened 2024-10-09
Address unused variable warning related to constexpr

blowekamp:


zivy (approved):
Nice. Was not familiar with the [[maybe_unused]] directive.

PR #2202, opened 2024-10-10
Use xcode 14.3.1 for packaging

blowekamp:


blowekamp:
The manual arm builds were built with RelMinSize before.

PR #2203, opened 2024-10-10
Create Packages with RelMinSize build type

blowekamp:
This may produce binaries 30-50% smaller.


blowekamp:
No large gains with this configuration change.

PR #2219, opened 2024-12-12
Use build-in xml package

blowekamp:
Prefer the built in xml package over the nonstandard lxml package.


blowekamp:
@dave3d Why was lxml used of xml?

dave3d (approved):
Looks good to me.  Sorry about the delay.

PR #2226, opened 2025-01-03
Update itk 4.5.2

blowekamp:


zivy (approved):
Commit message says 4.5.2 and should be 5.4.2 which is correctly specified in the file, the 5 and 4 were transposed in the commit message.

blowekamp:
And the branch name too, need a new PR.

PR #2232, opened 2025-01-15
DOC: Update R documentation.

zivy:
Update the documentation to match the change made from using the devtools package to the remotes package for GitHub based installation.


blowekamp (approved):
Thanks for keeping this up to date.

PR #2236, opened 2025-02-07
Add upto 5D support to MinimumProjectionImageFilter

blowekamp:


blowekamp:
The InsightSoftwareConsortium/ITK#5231 PR, will enable direct support for VectorImages to Min/Max projection image filter. This will enable removal of the per component feature used in the json. This will reduce the size of the object code substantially. But that is in the future.

dave3d (approved):

PR #2238, opened 2025-02-18
Update clang format

blowekamp:
The clang format file it copied from ITK.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,318 @@
+## This config file is only relevant for clang-format version 19.1.7
+##
+## Examples of each format style can be found on the in the clang-format documentation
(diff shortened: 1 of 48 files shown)


zivy (approved):
Does the new .clang-format configuration file match the ITK clang settings? I suspect this  update is intended to keep the SimpleITK C++ style consistent with ITK? If so, indicate this in the commit message or possibly in the header of the config file?

If this isn't the case, then just ignore this comment. Always good to have a consistent styling tool run on the code.

blowekamp:
Yes, the .clang-format file comes for ITK. I updates the commit message.

dave3d (approved):

PR #2240, opened 2025-02-18
Use pre-commit GHA action

blowekamp:


zivy (commented):
Not sure if moving away from explicit installation of pre-commit to the pre-commit action is the right thing to do. Specifically, the pre-commit action README says "this action is in maintenance-only mode". 

Possibly it doesn't matter as the action is doing what the deleted configuration does, so doesn't matter that it isn't updated and only maintained in working order?

dave3d (approved):

zivy (approved):

PR #2243, opened 2025-02-21
Adress failing Doxygen builds in Packaging CI

blowekamp:
Updated script to follow modern python conventions. Fixed issues with using legacy python 2 string and print syntax.


dave3d (approved):
black would reformat this file a bit.

blowekamp:
We have black running as a pre-commit and in the CI, so it shouldn't... But I'll double check

PR #2251, opened 2025-02-26
Increase ITK coverage time out for tests

blowekamp:
The itkComposeBigVectorImageFilterTest was timing out.


dave3d (approved):
LGTM!

blowekamp:
The itkComposeBigVectorImageFilterTest test is still timing out.  Maybe some profiling needs to be done on the test to see why it 's so slow.

dzenanz (approved):
LGTM

PR #2255, opened 2025-03-13
Update Build.yml to use Codebuild runners

jwj019:
This PR changes the label from the Monarch self-hosted runners to a new set of runners that use AWS Codebuild as a backend.  A Codebuild project has been set up according to the AWS documentation to [Configure a CodeBuild-hosted GitHub Actions runner](https://docs.aws.amazon.com/codebuild/latest/userguide/action-runner.html#sample-github-action-runners-config).


blowekamp:
Only using code build on nightly builds.

PR #2259, opened 2025-04-04
Replace self-hosted runners with GH hosted runners

blowekamp:


zivy (approved):

zivy (approved):
This PR is now replacing self-hosted runners with github-hosted ones. The commit message about replacing self-hosted runners with codebuild should be updated.

blowekamp:
> This PR is now replacing self-hosted runners with github-hosted ones. The commit message about replacing self-hosted runners with codebuild should be updated.

The commit message is correct. I didn't hit save on the change of the name of the PR.

PR #2261, opened 2025-04-07
Replace self-hosted runners with CodeBuild in Nightly build

blowekamp:


blowekamp:
We will not be using the code build agents.

PR #2271, opened 2025-04-18
Add optional input for initial estimated displacement field

blowekamp:
Add support for providing an optional initial estimate of the inverse displacement field to the InvertDisplacementFieldImageFilter. This can significantly improve convergence and reduce computation time when a good initial estimate is available.

Updated from JSON to YAML configuration format to work with current main branch.


blowekamp:
A test should be added which uses this optional input.

dave3d (approved):

zivy (approved):

PR #2272, opened 2025-04-18
Remove signed label image support from label intensity filter

blowekamp:
Only support unsigned label input images for the
LabelIntensityStatisticsImageFilter. This filter is a dual dispatch filter with intensity image input. The generated output object file for this filter was by far the largest. On a MAC ARM system this reduced the size of the object file from 27MB to 14MB.

Updated to work with YAML configuration files instead of JSON.

### DIFF
--- a/Code/BasicFilters/yaml/LabelIntensityStatisticsImageFilter.yaml
+++ b/Code/BasicFilters/yaml/LabelIntensityStatisticsImageFilter.yaml
@@ -4,7 +4,7 @@ template_code_filename: DualImageFilter
 template_test_filename: ImageFilter
 number_of_inputs: 0
 doc: Docs
-pixel_types: IntegerPixelIDTypeList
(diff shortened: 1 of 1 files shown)


blowekamp:
This PR reduces the memory usages during build time enough for the GH hosted Mac arm runners to complete. Unfortunately, it removes the support of some combinations of pixel types. Before this filter's supported images matched that of the LabelStatisticsImageFilter. However, the support for only signed label images corresponds to the scalar type supported by that Label Map image type.

zivy (approved):

blowekamp:
Note the LabelMaps are only instantiated with the unsigned pixel types: https://github.com/SimpleITK/SimpleITK/blob/7d731082b54fb401e5b2d3c6b48b86c37ea6ca53/Code/Common/include/sitkPixelIDTypeLists.h#L160

PR #2273, opened 2025-04-23
DOC: Remove reference to kitware hosted wiki.

zivy:
Wiki is no longer available from kitware and was only referenced for historical reasons, removed it from the Readme.md.


blowekamp (approved):
Nice!

PR #2274, opened 2025-04-23
DOC: Updated dead URLs in read-the-docs documentation.

zivy:
msysGit is no longer maintained, replaced with reference to "Git For Windows". Links to ITK doxygen updated to point to current ITK doxygen location. Links to SimpleITK doxygen updated.
Link to ImageJ website updated to current site.
Removed links to old tutorial websites that were taken down.

### DIFF
--- a/docs/source/faq.rst
+++ b/docs/source/faq.rst
@@ -127,7 +127,7 @@ programmatically hard code this header information. The suggested way is
 to create a Meta image header file (\*.mhd) which references the raw
 data file and describes the size and type of the data. The documentation
 on how to write a Meta image header can be found
-`here <https://www.itk.org/Wiki/MetaIO/Documentation#Reading_a_Brick-of-Bytes_.28an_N-Dimensional_volume_in_a_single_file.29>`__.
+`here <https://docs.itk.org/en/latest/learn/metaio.html#reading-a-brick-of-bytes-an-n-dimensional-volume-in-a-single-file>`__.
 
 The following is a sample Meta image header file, perhaps of name
 sample.mhd:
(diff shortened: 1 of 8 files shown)


docs/source/filters.rst
```diff
@@ -86,7 +86,7 @@ SimpleITK Filters
      - BinaryThreshold projection.
```
blowekamp, line 86: I believe this file may be generated from "Utilities/filters.csv" with the "CompareITKandSITKFilters.py". That script looks like it is in need of an update. Some of the incorrect filter names in the simpleITK doxygen linked may be due to issues in the script.

Additionally, I am not sure we should change the simpleitk URL from latest to different version numbers.

docs/source/faq.rst
```diff
@@ -127,7 +127,7 @@ programmatically hard code this header information. The suggested way is
 to create a Meta image header file (\*.mhd) which references the raw
 data file and describes the size and type of the data. The documentation
 on how to write a Meta image header can be found
-`here <https://www.itk.org/Wiki/MetaIO/Documentation#Reading_a_Brick-of-Bytes_.28an_N-Dimensional_volume_in_a_single_file.29>`__.
+`here <https://docs.itk.org/en/latest/learn/metaio.html>`__.
```
blowekamp, line 130: Please use the hashed URL: https://docs.itk.org/en/latest/learn/metaio.html#reading-a-brick-of-bytes-an-n-dimensional-volume-in-a-single-file

docs/source/filters.rst
zivy: All versions set to latest (force pushed). I think we can update the script at a later time.

blowekamp (approved):

PR #2280, opened 2025-05-06
Use CodeBuild runners in nightly.

blowekamp:


.github/workflows/Nightly.yml
```diff
@@ -87,6 +88,21 @@ jobs:
           enableCrossOsArchive: true
           restore-keys: |
             external-data-v1-
+      - name: Set up Python 3.11
+        uses: actions/setup-python@v5
+        id: cpy
+        with:
+          python-version: 3.11
```
zivy, line 95: Should this use Python 3.12 or 3.13? I was looking at the [status of Python versions page](https://devguide.python.org/versions/) recently and realized that Python 3.11 is starting to get old.

.github/workflows/Nightly.yml
```diff
@@ -153,6 +170,22 @@ jobs:
           repository: InsightSoftwareConsortium/ITK
           path: ITK-dashboard
           ref: dashboard
+      - name: Set up Python 3.11
+        uses: actions/setup-python@v5
+        id: cpy
+        with:
+          python-version: 3.11
```
zivy, line 177: Same question as above, with respect to Python version.

zivy (approved):

.github/workflows/Nightly.yml
blowekamp: It is good enough for a couple of years.

PR #2289, opened 2025-05-12
Update itk to 6.0a3 for development

blowekamp:


blowekamp:
@N-Dekker With updating ITK to 6.0a3 a compilation error is occurring with Elastix. Likely does to the removal of some ITK macros like ITKv5_CONST or something similar. Any recommendations with Elastix and compatibility with ITK 6.0 for development?

blowekamp:
> > With updating ITK to 6.0a3 a compilation error is occurring with Elastix. Likely does to the removal of some ITK macros like ITKv5_CONST or something similar. Any recommendations with Elastix and compatibility with ITK 6.0 for development?
> 
> Thanks for asking, @blowekamp I locally use ITK v6.0a02 when building elastix from the main branch. Do you think anything between ITK v6.0a02 and ITK v6.0a03 triggered the errors?
> 
> Which version (commit or tag) of elastix did you try?

SimpleITK currently is on Elastix "5.2.0". Looking closer at the error is seems to be related to missing trailing ";" after macros. Perhaps a patch to 5.2.0 is in order? I'll look into making a patch.

dave3d (approved):

PR #2291, opened 2025-05-13
Use large instance size for nightly build

blowekamp:


blowekamp:
This instance type runs out of disc space.

PR #2294, opened 2025-05-15
Add ForceOrthogonalDirection and ReverseOrder options - for release

blowekamp:
Add missing options to the ImageSeriesReader

### DIFF
--- a/Code/IO/include/sitkImageSeriesReader.h
+++ b/Code/IO/include/sitkImageSeriesReader.h
@@ -169,6 +169,33 @@ public:
   double
   GetSpacingWarningRelThreshold() const;
 
+  /** Set/Get/OnOff whether to force orthogonal direction cosines.
+   *
+   * Do we want to force orthogonal direction cosines? On by default. Turning it off enables proper reading of DICOM
+   * series with gantry tilt.
+   */
+  SITK_RETURN_SELF_TYPE_HEADER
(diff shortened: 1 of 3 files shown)


dave3d (approved):

blowekamp:
@zivy  What do you think about back porting these changes to 2.4?

zivy:
@blowekamp, I believe in SimpleITK 2.x we don't force the orthogonal axes, so as long as the default for this is `False` then it would be nice to have this functionality in the 2.x releases. Otherwise I would view this as a breaking change which should happen in 3.0 where changes are widely expected.

blowekamp:
> @blowekamp, I believe in SimpleITK 2.x we don't force the orthogonal axes, so as long as the default for this is `False` then it would be nice to have this functionality in the 2.x releases. Otherwise I would view this as a breaking change which should happen in 3.0 where changes are widely expected.

This was not set in SimpleITK before, the default is true in ITK:
https://github.com/InsightSoftwareConsortium/ITK/blob/4c028be3ba0ed839702893e6057c51a50a83ed03/Modules/IO/ImageBase/include/itkImageSeriesReader.h#L207

So this patch does not change behavior.

zivy:
Thanks for the clarification. Given that the behavior didn't change, then it would be great to have this functionality already in the 2.x version so that folks don't have to wait for 3.x to use it if they need this functionality.

PR #2296, opened 2025-05-16
Add key word only options to ReadImage

blowekamp:
Enables advance options like ReverseOrder, and Debug to be set with the procedural ReadImage interface


zivy (approved):
Elegant implementation with the keywords configured to invoke "Set" methods using introspection.

dave3d (approved):

PR #2302, opened 2025-05-30
Use boolean types for boolean attributes in JSON

blowekamp:
Convert data representation from string or integers to boolean value types in JSON.


dave3d (approved):

zivy (approved):
LGTM. 

There were some places where "1" was converted to 1 and some to "true". Initially thought that was an error, but second look at the variables realized that these are correct conversions. In some cases the string represented the number and in some the boolean value.
Nicely done.

PR #2306, opened 2025-05-30
Fix Docker FromAsCasing error for Doxygen - for RELEASE

blowekamp:


zivy (approved):
Is there a reason for naming this stage in the docker build? If it isn't referenced later it doesn't need to be named, `FROM ubuntu:24.04` should work. Is this a convention, or is there something else I'm missing?

blowekamp:
It gives the name of the build so that it can be run with the docker command. The documentation references this image name.

PR #2308, opened 2025-06-02
Json conformance

blowekamp:
Making step towards defining a JSON schema and having all JSON files conform.


Code/BasicFilters/json/ConnectedThresholdImageFilter.json
```diff
@@ -128,7 +127,6 @@
         {
           "parameter" : "SeedList",
           "point_vec" : true,
-          "dim" : 2,
           "value" : [
```
zivy, line 130: Is the `dim` entry deleted because it is now derived from the length of the "value" field entries now? 

Asking so that if this is an incorrect deletion it is undone (there are several instances of this change in the PR).

zivy (approved):

dave3d (approved):

Code/BasicFilters/json/ConnectedThresholdImageFilter.json
blowekamp: The fields removed are no longer used. It's not worth speculating how they were once used. Currently the values are just concatenated to form arguments for each language.

PR #2309, opened 2025-06-03
Update to windows-2022

blowekamp:
The windows-2019 image is being depricated.


dave3d (approved):
I guess I should check what version of VS Studio my PC is running.

PR #2310, opened 2025-06-03
Add formal JSON schema

blowekamp:
Add pre-commit json validation, and use script to validate json before the build process.

### DIFF
--- a/ExpandTemplateGenerator/simpleitk_filter_description.schema.json
+++ b/ExpandTemplateGenerator/simpleitk_filter_description.schema.json
@@ -1,0 +1,560 @@
+{
+  "$schema" : "http://json-schema.org/draft-07/schema#",
+  "title" : "SimpleITK Filter Description",
+  "type" : "object",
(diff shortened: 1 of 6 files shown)


ExpandTemplateGenerator/simpleitk_filter_description.schema.json
```diff
@@ -0,0 +1,560 @@
+{
+  "$schema" : "http://json-schema.org/draft-07/schema#",
+  "title" : "SimpleITK Filter Description",
+  "type" : "object",
+  "properties" : {
+    "name" : {
+      "type" : "string",
+      "description" : "Name of the filter as a SimpleITK class/"
```
zivy, line 8: "class/" -> "class."

zivy (changes requested):
In the "description" fields, some end with a period, some don't. Possibly make consistent, but can be ignored. Actual change requests are inline comments.

zivy (approved):

PR #2315, opened 2025-06-10
Add pretty-format-json to pre-commit

blowekamp:
This changes the separate from " : " to ": " removing a space. This is not configurable with this hook.


zivy (approved):
Nice. 
Now the JSON files have both good function and consistent form. I don't think it's possible to maintain indenting/spacing consistency without an automated tool, at least I can't, so good addition to pre-commit.

PR #2317, opened 2025-06-10
ENH: Address changes in Fiji installation on windows.

zivy:
Newer versions of the Fiji distribution (5/2025 and later) now install in a directory called Fiji and the executable is
fiji-windows-x64.exe. Adding to the image viewer's default-executables list.


zivy:
@dave3d, Thanks for working on this. Which version of the Fiji distribution did you install, the JDK or JRE? It shouldn't matter, as both should work, but just that we know which.

dave3d (approved):

PR #2326, opened 2025-06-16
Update JSONBeautify Python script.

blowekamp:
Change separator to match pre-commit configuration.

Add argparse interface to support multiple json file on the command line, and different verbosity levels.

### DIFF
--- a/Utilities/JSON/JSONBeautify.py
+++ b/Utilities/JSON/JSONBeautify.py
@@ -1,23 +1,85 @@
-#!/usr/bin/env python3
-import json
-import sys
-
-try:
-    from collections import OrderedDict
-except ImportError:
-    import ordereddict as OrderedDict
-
-# This script takes one argument the name of a json file.
+#!/usr/bin/env python
+#==========================================================================
 #
+#   Copyright NumFOCUS
+#
+#   Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


Utilities/JSON/JSONBeautify.py
```diff
@@ -1,23 +1,85 @@
...
+
+        try:
+            # Load JSON with OrderedDict to preserve key order
+            logging.info(f"Processing '{json_file}'")
+            with open(json_file, 'r', encoding='utf-8') as fp:
+                json_data = json.load(fp, object_pairs_hook=OrderedDict)
 
+            # Write beautified JSON back to the file
+            with open(json_file, 'w', encoding='utf-8') as fp:
+                json.dump(json_data, fp, indent=args.indent, separators=(",", ": "), ensure_ascii=False)
```
zivy, line 64: What is the reason for changing the default `ensure_ascii` from `True` to `False`? Are there non-ascii characters in the JSON files that should not be escaped?

zivy (approved):

Utilities/JSON/JSONBeautify.py
blowekamp: There are non-ascii characters in the JSON. In the documentation strings from ITK as I recall. I am unaware of a need for them to be escaped. In general we can use UTF-8 character encoding for text files.

PR #2332, opened 2025-06-19
Add gersemi cmake formatter

blowekamp:
Use [gersemi](https://github.com/BlankSpruce/gersemi) to apply automatic formatting to CMake file with pre-commit.

### DIFF
--- a/.gersemi.config
+++ b/.gersemi.config
@@ -1,0 +1,14 @@
+## Outcome configuration based on defaults,
+# yaml-language-server: $schema=https://raw.githubusercontent.com/BlankSpruce/gersemi/0.19.3/gersemi/configuration.schema.json
+
+definitions: [ "CMake/stubs" ]
+disable_formatting: false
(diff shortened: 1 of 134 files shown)


blowekamp:
I used the following prompt to generate the stubs:
```
Please look at the section called "Let's make a deal" in the following document: [https://github.com/BlankSpruce/gersemi/blob/master/README.md](vscode-file://vscode-app/Users/blowekamp/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-sandbox/workbench/workbench.html) for each cmake function defined in this file create a subs file in the CMake/stubs director yto describe the arguments for the gersemi formatter.
```

zivy (approved):
In most cases the list_expansion is an improvement, though in some cases I find the results to be less readable than the original non-expanded version. 

Still, using this linter for CMake is an improvement as it enforces uniformity. Nice addition.

PR #2336, opened 2025-06-23
DOC: Add C#, C++ and java versions to the DicomConvert example.

zivy:
Provide the example in three additional languages.

### DIFF
--- a/.github/workflows/Build.yml
+++ b/.github/workflows/Build.yml
@@ -94,6 +94,10 @@ jobs:
       uses: r-lib/actions/setup-r@v2
       with:
         r-version: 'latest'
+    - name: Install Examples R package dependencies
+      if: contains(matrix.ctest-cache, 'WRAP_R')
+      run: |
+        Rscript -e 'install.packages(c("argparser"))'
     - name: Generate External Data Hash
       shell: bash
       run: |
--- a/.github/workflows/additional_dictionary.txt
+++ b/.github/workflows/additional_dictionary.txt
@@ -1854,7 +1854,7 @@ rescale
 rescaled
 rescope
 resizable
-resizeable
+resize
 return
 reusability
 rfind
--- a/Examples/DicomConvert/DicomConvert.cxx
+++ b/Examples/DicomConvert/DicomConvert.cxx
@@ -1,0 +1,163 @@
+/*=========================================================================
+ *
+ *  Copyright NumFOCUS
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0.txt
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ *
+ *=========================================================================*/
+
+/**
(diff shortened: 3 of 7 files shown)


.github/workflows/additional_dictionary.txt
```diff
@@ -1854,6 +1854,7 @@ rescale
 rescaled
 rescope
 resizable
+resize
```
blowekamp, line 1857: Why do we have different spelling of resizable above and below this?

Examples/DicomConvert/DicomConvert.cxx
```diff
@@ -0,0 +1,181 @@
...
+                             image.GetDirection(),
+                             0,
+                             image.GetPixelID());
+    }
+    // If a single channel image, rescale to [0,255]. Also modify the
+    // intensity values based on the photometric interpretation. If
+    // MONOCHROME2 (minimum should be displayed as black) we don't need to
+    // do anything, if image has MONOCHROME1 (minimum should be displayed as
+    // white) we flip the intensities. This is a constraint imposed by ITK
+    // which always assumes MONOCHROME2.
```
blowekamp, line 83: So the goal here is that if the input in monochrome1 then to write the output as monochrome1? The important part to note here is that ITK inverts the intensity on read, but if output is monochrome1 then it does not automatically invert the intensity. (bug?)

Examples/DicomConvert/DicomConvert.cxx
```diff
@@ -0,0 +1,181 @@
...
+      if (iSize[2] == 1)
+      {
+        imageSize.push_back(0U);
+      }
+      else
+      {
+        imageSize.push_back(static_cast<unsigned int>(iSize[2]));
+      }
+    }
+    reader.SetExtractSize(imageSize);
```
blowekamp, line 59: You are always reading the whole image. All these conditional are just to convert a single size image to 3D->2D.  I would just read the image. Check is dimension, then if dim==3 then use the extract image filter to reduce the dimension. This should be more compact and skip this messiness with types.

Examples/DicomConvert/DicomConvert.cxx
zivy: The script is converting the DICOM content to another file format which already requires intensity modifications from a high dynamic range to a low one. Because of the ITK convention of inverting MONOCHROME1 when reading, we need to invert it again so that when the file is saved it is visually similar to the original DICOM when displayed. I don't think this is a bug, it is the ITK convention and the desire to have consistent appearance of the converted image with the original DICOM content.

Examples/DicomConvert/DicomConvert.cxx
zivy: Sounds good, will replace the usage of the reader's setExtractSize with the extract image filter in the strongly typed languages.

.github/workflows/additional_dictionary.txt
zivy: That is strange. No idea why it's there. I will remove and hopefully testing still passes.

Examples/DicomConvert/DicomConvert.cxx
blowekamp: I think I understand the logic less now.

.github/workflows/Build.yml
```diff
@@ -94,6 +94,10 @@ jobs:
       uses: r-lib/actions/setup-r@v2
       with:
         r-version: 'latest'
+    - name: Install Examples R package dependencies
+      if: contains(matrix.ctest-cache, 'WRAP_R')
+      run: |
```
blowekamp, line 99: This change set does not contain any R code changes. Is the a missing dependency with the R example?

blowekamp (approved):
The Examples look good. Although the change to the Build.yml looks off topic.

PR #2350, opened 2025-07-03
Disable passing filename on some local pre-commit hooks

blowekamp:


.pre-commit-config.yaml
```diff
@@ -83,6 +84,7 @@ repos:
       entry: 'Utilities/Hooks/kw-pre-commit'
       language: system
       stages: [pre-commit]
+      pass_filenames: false
```
zivy, line 87: Not clear to me why the `pass_filenames`  value is changed from its default of `true`. Is this hook run on all file names? That would be a bit strange. My expectation is that it is run just on the changed+added files, so no need to change the `pass_filenames` as the default is appropriate. What am I missing?

zivy (approved):

PR #2353, opened 2025-07-09
Adding DICOM reading failure question to FAQ.

zivy:
Added question dealing with failure to read DICOM file. Point to relevant tools for debugging and a possible solution by building SimpleITK locally and including the DCMTK IO module.

### DIFF
--- a/docs/source/faq.rst
+++ b/docs/source/faq.rst
@@ -41,6 +41,27 @@ the version of SimpleElastix included in SimpleITK, you will need to
 using the SuperBuild configuration and turn on the SimpleITK_USE_ELASTIX flag in
 addition to the relevant WRAP_LANGUAGE flags for the desired programming languages.
 
+SimpleITK failed to read my DICOM image, what can I do to resolve this issue?
+-----------------------------------------------------------------------------
+
+To understand the reason for this failure we recommend using David Clunie's
+`Dicom3tools <https://dclunie.com/dicom3tools.html>`_. Specifically, the *dcidvfy* program
+to verify standard compliance and identify issues with the file and *dcdump* to look at the
+file's meta-data contents.
+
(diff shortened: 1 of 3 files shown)


docs/source/faq.rst
```diff
@@ -41,6 +41,26 @@ the version of SimpleElastix included in SimpleITK, you will need to
...
+read DICOM images. In some cases it fails to read the image. Fortunately, there is an additional
+optional DICOM IO module, DCMTK. To use the DCMTK IO you will need to
+:ref:`build SimpleITK <building_simpleitk>` locally using the
+SuperBuild configuration. Set CMake's :ref:`advanced options <advanced_build_options>`
+to include -D Module_ITKIODCMTK:BOOL=ON in addition to turning on the relevant WRAP_LANGUAGE flag
+for your programming language of choice. Then :ref:`install from the build tree <installation_from_build_tree>`.
+
+When you attempt to read the image now, both GDCM and DCMTK
+are available. If you do not specify which one to use, the toolkit will automatically try GDCM and then
+DCMTK. Additional information describing SimpleITK's image IO is available :ref:`here <image-io>`.
```
blowekamp, line 62: Should how to explicitly use DCMTK be shown? How will a user know which IO is used?

zivy: I don't think we need to explicitly show how to use DCMTK here. The text says `When you attempt to read the image now, both GDCM and DCMTK are available. If you do not specify which one to use, the toolkit will automatically try GDCM and then DCMTK.`

The last sentence points to the documentation that provides general details on how to use the IO mechanism. 

Is the following modification of that sentence better? 
Additional information describing SimpleITK's image IO mechanism and how to select a specific IO is available :ref:`here <image-io>`.

PR #2354, opened 2025-07-09
Make the Addressor constexpr

blowekamp:
Explicitly declare methods as constexpr to support additional compile time creation of objects.


zivy (approved):
Going green. 

Shifting computations to compile time increases our compute resource usage but decreases the compute for many people around the world.

blowekamp:
I don't think there where will be a measurable increase in compile time with this line of work.

PR #2371, opened 2025-07-18
Make construction of function factories compiles time

blowekamp:
Introduce a consteval hash map container class. This new class uses an std::array, and a consteval hash function.
The member function factory variable are not marked as constexpr with an constexpr/consteval lambda function used for initialization.


blowekamp:
@N-Dekker This may be some C++ that you'd find interesting to take a look at. This code initializes a hash map of function pointers at compile time.

blowekamp:
@hjmjohnson This is an interesting PR, where I was experimenting with a "constexpr/consteval" hash_map class. It was to store all the pointers to the the template functions used in SimpleITK. Unfortunately function address are computed during linking and not compilation time so it's not portable/well formed.

PR #2382, opened 2025-08-01
Improve formatting of generated doxygen code blocks

blowekamp:
Use the "indent" filter to add "*" to the prefix of multi-line doxygen comments. Additionally remove extra new lines in doxygen blocks.


dave3d (approved):

blowekamp:
> Is there a way to have the wordwrap parameter, 117, included from another file or the environment so that it is shared in all template files where it is used? It needs to be consistent across all usages and will require changes across multiple files if we want to change it in the future.

done

PR #2387, opened 2025-08-10
DOC: correct referencing of code in the rst file.

zivy:
Harmonize the code referencing to exclude the import/library/include parts and only include the relevant operation. Also C++ referencing was off by a couple of rows so didn't show the code correctly.


zivy:
Updated to use labels and force pushed.

blowekamp (approved):
Thanks for making that additional change.

PR #2394, opened 2025-08-12
Convert JSON to YAML

blowekamp:
Use YAML for filter description. YAML supports comments, and better options of multiline string. The yaml files are formatted better for editing.


zivy (approved):
The multiple newlines confused me a bit but that was already addressed in the GitHub comment, possibly include that as part of the commit message. I think this version is still preferable over the very long single line text in JSON. 

The only slightly inconsistent aspect of the change is that the yaml files are in a directory named "json". Possibly change directory name to "templates" so that it doesn't matter if it is json or yaml.

zivy:
This should have a followup PR to update the FilterTemplates.rst documentation as those refer to the JSON information.

dave3d (approved):

PR #2397, opened 2025-08-14
Add override to SimpleLoggerAdaptor methods and update logger tests

blowekamp:
- Marked all overridden methods in SimpleLoggerAdaptor with override
- Updated logger tests and test harness for context-aware logging
- Ensured interface consistency with ITK OutputWindow API


blowekamp:
We need to update the examples for usage of this newer interface.

PR #2405, opened 2025-08-20
Update the ImageGridManipulation example.

zivy:
Enable the testing. It was previously not run because the ImageGridManipulation directory was not added via add_subdirectory to the Examples CMakeLists.txt. Modify the testing input to match the expected type. Modify the usage of VectorIndexSelectionCast to the current API and add java, C++ and C# versions.


blowekamp (commented):
I don't think the python code represents current exemplar usage of the python interface. 

This example may have intentionally not been added due to the notebooks showing better usage. There have been several improved ways to perform image grid manipulations since this was first written. Many of those are language specific for R and Python.

zivy:
This PR adds the example to the **testing** framework which was missing. The incorrect Python and R code [is currently included](https://simpleitk.readthedocs.io/en/master/link_ImageGridManipulation_docs.html) on read-the-docs. 

I can either update the Python and R code or we can remove the whole example from read-the-docs. Which option makes more sense to you?

blowekamp:
Seeing the current state explains things:
https://simpleitk.readthedocs.io/en/master/link_ImageGridManipulation_docs.html

I'm not sure that the page layout is best for what it's trying to explain, but that is another issue.

PR #2442, opened 2025-11-05
Script for analyzing GitHub release downloads.

zivy:
Using the GitHub API fetch all the information about the release assests and analyze it. The script assumes SimpleITK specific file naming conventions for the release assets to identify programming languages and operating systems.


Utilities/Statistics/environment.yml
```diff
@@ -1,11 +1,9 @@
...
 
 dependencies:
-  - matplotlib
-  - numpy
-  - cartopy
-  - pandas
-  - requests
+  - python=3.12
+  - pip
+  - pip:
```
blowekamp: Two points:
- My preference has become to use uv and uvx to run things.  And there is a new Python standard for embedding metadata in a standalone script. If you are interested in new ways to do things ( as a replacement to the yaml), you can look at this: https://docs.astral.sh/uv/guides/scripts/#creating-a-python-script
- It has become bad practice to mix conda and pip, as it can lead to binary incompatibilities. I don' think there is a need to use pip in the yaml.

PR #2447, opened 2025-11-19
Add PyBuffer interface to Image (PEP 688) and numpy array conversion

blowekamp:
Enables view and conversions to numpy arrays with the following:
```
img = sitk.Image((2,5), sitk.sitkVectorUInt8, 3)
a_view = np.asarray(img, copy=False)
a_copy = np.array(img, copy=True)
```

References to the Python Image object are maintained. 

- Remove internal GetMemoryViewFromImage function
- Add pixel_types module to Python
- Add array interface and buffer protocol to Image class
- Add a CPython ImageBuffer class to have pyBuffer interface

COMPATIBILITY:

Images of vector pixel type with only 1 component now return memory views and arrays of n+1 dimensions. ( Previously the 1 size component dimension was dropped )

### DIFF
--- a/.github/workflows/additional_dictionary.txt
+++ b/.github/workflows/additional_dictionary.txt
@@ -952,6 +952,7 @@ buildsystems
 burton
 bwh
 byteorder
+buf
 cacheline
 callables
 cardinality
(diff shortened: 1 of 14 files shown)


blowekamp:
@dave3d Is there a way to tell the spell checker to ignore a whole file? In the pre-commit config or else where?

zivy (approved):
So Pythonic and makes the usage of views safer for users. Very nice work! 

Some trivial spelling that can be ignored.

PR #2456, opened 2025-11-26
Use pip wheel for PEP 571 python build system

blowekamp:


zivy (approved):
LGTM. I did find this [reference to eggs](https://github.com/SimpleITK/SimpleITK/blob/2c95e2f05fb9174d1643aa0306d4d3445c43e782/Utilities/Distribution/manylinux/imagefiles/cmd.sh#L81
) in the code base. Possibly remove that too and include in this PR?

PR #2475, opened 2025-12-12
Add using ITK with CMake's FetchContent module

blowekamp:
Documentation on CMake's FetchContent can be found here:
https://cmake.org/cmake/help/latest/module/FetchContent.html

This is an experimental option for SimpleITK as an alternative to the Superbuild. With this change the Superbuild build of ITK still works, and the SimpleITK project still respects ITK_DIR with the FetchContent trying to use `find_package(ITK)` to determine if ITK needs to build built.

If ITK needs to be built with FetchContent, then ITK is downloaded, and configured during SimpleITK's configuration. And ITK shares the same CMake configuration cache with the SimpleITK project to enable consistent, and easier tuning of the ITK build through the SimpleITK project.

### DIFF
--- a/CMake/sitkITKFetchContent.cmake
+++ b/CMake/sitkITKFetchContent.cmake
@@ -1,0 +1,96 @@
+#-----------------------------------------------------------------------------
+# Get and build ITK using FetchContent
+
+include(FetchContent)
+
+# Set ITK Git repository and tag
+set(
+  ITK_GIT_REPOSITORY
+  "https://github.com/InsightSoftwareConsortium/ITK.git"
+  CACHE STRING
+  "URL of ITK Git repository for FetchContent"
+)
+mark_as_advanced(ITK_GIT_REPOSITORY)
+
+# The commit contains the ExternaData changes to work with SimpleITK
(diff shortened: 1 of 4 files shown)


CMake/sitkITKFetchContent.cmake
```diff
@@ -0,0 +1,96 @@
...
+  set(CMAKE_C_VISIBILITY_PRESET "hidden" CACHE STRING "" FORCE)
+  set(CMAKE_CXX_VISIBILITY_PRESET "hidden" CACHE STRING "" FORCE)
+  set(ITK_TEMPLATE_VISIBILITY_DEFAULT OFF CACHE BOOL "" FORCE)
+endif()
+
+# Set ITK build options
+set(ITK_USE_KWSTYLE OFF)
+
+# ITK automatically checks COMPILER_LAUNCHER variables to automatically enable ccache option.
+# However additional contrains and usages are added, that are not needed for
```
zivy: "ITK **automatically** checks COMPILER_LAUNCHER variables to **automatically**..." -> drop first "automatically", "ITK checks COMPILER_LAUNCHER variables to automatically..."

Second sentence is not complete, not sure about the intention ("However...").

zivy (approved):

dave3d (approved):

PR #2476, opened 2025-12-12
Fix extension for Python module on windows - for release

blowekamp:
When limited API is enabled, it was being evaluated as "..pyd". This prevented the library from being loaded.


blowekamp:
Closing as no longer needed. This backports the fix from #2474 (merged to `main`), but I verified that the bug it addresses does not manifest on the `release` branch's current Windows packaging pipeline (raw CMake/CTest via `win_build_python.sh`, distinct from `main`'s scikit-build-core path where the original issue was observed).

Downloaded the actual published Windows limited-API wheel for the latest release, `simpleitk-2.5.6-cp311-abi3-win_amd64.whl`, and confirmed the packaged module is correctly named `SimpleITK/_SimpleITK.pyd` (single dot) rather than the broken `_SimpleITK..pyd` this PR guards against. So the `release` branch has been shipping working Windows wheels without this fix.

Reopen/reapply if a future change to the `release` branch's packaging pipeline (e.g. adopting scikit-build-core) reintroduces the empty `Python_SOSABI` condition this guards against.

PR #2501, opened 2026-01-13
Update Doxygen build to use ITK FetchContent

blowekamp:
By using ITK FetchContent, the ITK dependecies don't need to be built to build the SimpleITK Doxygen.

### DIFF
--- a/Utilities/Doxygen/docker/README
+++ b/Utilities/Doxygen/docker/README
@@ -1,18 +1,109 @@
-This dockerfile has been created to facilitate the creation of SimpleITK html
-doxygen documentation.
+# SimpleITK Doxygen Documentation Builder
 
-Run the following commands:
+## Overview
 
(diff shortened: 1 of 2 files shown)


Utilities/Doxygen/docker/README
```diff
@@ -1,18 +1,109 @@
...
+docker rm sitk-dox
+```
+
+## Complete Example
+
+```bash
+# Build the image (only needed once)
 docker build -f Dockerfile -t simpleitk-doxygen .
 
-  export TAG=v1.2.0
```
zivy: I believe the common build pattern is to use the `release` branch so no need to specify the export here. If the more common usage is expected to be "tag based" then the export statement should be kept.

zivy (approved):

dave3d (approved):

PR #2522, opened 2026-03-09
Update ITK version and refactor variables

blowekamp:
This update addresses a variable scoping issue in ITK with the MINC library.

### DIFF
--- a/CMake/sitkGitProtocol.cmake
+++ b/CMake/sitkGitProtocol.cmake
@@ -1,0 +1,24 @@
+#-----------------------------------------------------------------------------
+# Set the default GIT protocol for fetching dependencies.
+
+find_package(Git)
+set(_SimpleITK_GIT_PROTOCOL_default "https")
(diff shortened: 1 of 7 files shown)


CMake/sitkGitProtocol.cmake
```diff
@@ -0,0 +1,24 @@
...
+if(GIT_VERSION_STRING VERSION_LESS "1.7.10")
+  # minimum version for https support
+  set(_SimpleITK_GIT_PROTOCOL_default "git")
+endif()
+set(
+  SimpleITK_GIT_PROTOCOL
+  ${_SimpleITK_GIT_PROTOCOL_default}
+  CACHE STRING
+  "If behind a firewall turn set this to 'https' or 'http'."
+)
```
zivy: "turn set this" -> "set this"

zivy (approved):
Shared config for both build approaches is very nice!

PR #2527, opened 2026-03-13
ENH: Use CMake TIMESTAMP

blowekamp:
Replace python code with CMake code to produce the date and time.

Inspired by InsightSoftwareConsortium/ITK#5942


zivy (approved):
Cool usage of CMake configuration time variable.

PR #2544, opened 2026-03-26
Test packaging SimpleITK with Elastix

blowekamp:


blowekamp:
This appears to be working. If anyone tests an artifact then please report if it works for you or not.

blowekamp:
Elastix has been enabled as an option in the standard distribution.

PR #2550, opened 2026-04-02
WIP: Add Jinja2 templates for generating filter examples

dave3d:
Add a new ExpandTemplateGenerator/templates/examples/ directory with Jinja2 templates that generate runnable example programs for select SimpleITK filters in all supported languages (Python, C++, Java, C#, R, Ruby, Lua), plus an RST documentation stub and CMakeLists.txt with per-language test targets.

Templates handle single-input, dual-input, named-input, image-source, and measurement-only (no_return_image) filters. Input data and settings are taken from an optional top-level 'example' key in the filter YAML or fall back to the first 'tests' entry.

Also add Examples/example_filter_list.yaml: opt-in list of ~40 filters.


blowekamp:
Please don't use git LFS, we have an existing CMake infrastructure to download binary data.

PR #2556, opened 2026-04-09
ENH: Use Ninja generator for manylinux builds

blowekamp:
Switch the manylinux Docker image builds from make to Ninja. Add ninja-build to the yum install in all four Dockerfiles and configure CMake with -GNinja in cmd.sh. Replace bare make and make dist calls with cmake --build equivalents. Remove the MAKEFLAGS export since Ninja handles parallelism automatically.


Utilities/Distribution/manylinux/imagefiles/cmd.sh
```diff
@@ -50,7 +50,7 @@ build_simpleitk() {
         -DITK_C_OPTIMIZATION_FLAGS:STRING="" \
         -DITK_CXX_OPTIMIZATION_FLAGS:STRING="" \
         ${SRC_DIR}/SuperBuild &&
-    make  &&
+    cmake --build . &&
```
zivy, line 53: Should this also include `-j ${NPROC}` on this line to match the Make build or is the default number of processes for a Ninja build desired, `${NPROC}+2`?

zivy (approved):

dave3d (approved):

Utilities/Distribution/manylinux/imagefiles/cmd.sh
blowekamp: There seems to be no issue with the default number.

PR #2558, opened 2026-04-10
Update elastix to af47e22, and use target based properties

blowekamp:
This update to elastix added target based includes. Update SimpleITK usage to only link to the elastix library targets.


blowekamp:
@N-Dekker FYI

dave3d (approved):

zivy (approved):

PR #2566, opened 2026-04-14
SWIG use doxygen flag for automatic docs

blowekamp:
Replace generated DocString files with SWIG parsed doxygen to docstrings processing.


dave3d (approved):

blowekamp:
This SWIG based passing is a little different than our current generated doctoring 'I' file. I think they are good enough in most cases. However there are some warnings related to mathtex, that may need additional work.

zivy (approved):

PR #2568, opened 2026-04-15
Don't use batch file for VS env

blowekamp:


blowekamp:
This change is not needed

Issue #2574, opened 2026-04-29
R wrapping build failures with R 4.6.0: removed non-API C functions


blowekamp:
## Description

R 4.6.0 (released April 2025) removed several non-API C functions from its public headers as part of an ongoing API cleanup effort. This breaks the SimpleITK R wrapping build with 6 compilation errors.

**CDash build:** https://open.cdash.org/builds/11240852/build

## Errors

There are three categories of errors:

### 1. `CLOENV` removed — SimpleITK source code

In `Wrapping/R/sitkRCommand.cxx` line 89:

```
error: 'CLOENV' was not declared in this scope
   this->SetCallbackREnviron(CLOENV(this->m_FunctionClosure));
```

`CLOENV` was removed from the R API in R 4.6.0. The replacement is `R_ClosureEnv()`, available since R 4.5.0. Per the [Writing R Extensions](https://cran.r-project.org/doc/manuals/r-devel/R-exts.html) manual §6.23.1:

> **`CLOENV`** — Use `R_ClosureEnv`, added in R 4.5.0.

**Fix:** Replace `CLOENV(this->m_FunctionClosure)` with `R_ClosureEnv(this->m_FunctionClosure)`, using a version guard as a backport for older R versions:

```c
#if R_VERSION < R_Version(4, 5, 0)
# define R_ClosureEnv(x) CLOENV(x)
#endif
```

### 2. `SET_S4_OBJECT` removed — SWIG-generated code

In the SWIG-generated `SimpleITKR_wrap.cxx` at lines 1133, 1151, and 1174:

```
error: 'SET_S4_OBJECT' was not declared in this scope; did you mean 'NEW_OBJECT'?
   SET_S4_OBJECT(r_obj);
```

This is an upstream SWIG issue: https://github.com/swig/swig/issues/3407

The fix is to replace `SET_S4_OBJECT(x)` with `x = Rf_asS4(x, TRUE, 0)` in SWIG's `Lib/r/rrun.swg`. Until SWIG ships a fix, this may require a local workaround.

### 3. `CHARACTER_POINTER` / `STRING_PTR_RO` now read-only — SWIG-generated code

In the SWIG-generated `SimpleITKR_wrap.cxx` at lines 2561 and 3027:

```
error: assignment of read-only location '*(STRING_PTR_RO(result) + ...)'
   CHARACTER_POINTER(result)[pos] = Rf_mkChar(((*val)[pos]).c_str());
```

[...]

PR #2576, opened 2026-04-29
BUG: Replace removed CLOENV with R_ClosureEnv for R 4.6.0 compatibility

blowekamp:
## Summary

Fixes part 1 of #2574.

R 4.6.0 removed several non-API C macros from its public headers. This PR fixes both the SimpleITK R wrapper and the bundled SWIG superbuild to compile and run against R 4.6.0.

### `Wrapping/R/sitkRCommand.cxx`

`CLOENV` was removed in R 4.6.0. The replacement `R_ClosureEnv()` was added in R 4.5.0. Line 90 used `CLOENV(this->m_FunctionClosure)` to retrieve the closure environment when setting up the R callback.

Added a version-guarded backport for R < 4.5.0:

```cpp
// R_ClosureEnv was added in R 4.5.0, CLOENV was removed in R 4.6.0
#if R_VERSION < R_Version(4, 5, 0)
#  define R_ClosureEnv(x) CLOENV(x)
#endif
```

### SuperBuild SWIG patch (`swig-r-api-r460.patch`)

R 4.6.0 also removed two other non-API macros used by SWIG's R runtime and typemaps:

- `SET_S4_OBJECT` — 3 uses in `Lib/r/rrun.swg`, replaced with `Rf_asS4(obj, TRUE, 0)`
- `CHARACTER_POINTER` — 2 uses in `Lib/r/std_vector.i`, replaced with `SET_STRING_ELT`

A patch file is applied via `ExternalProject` `PATCH_COMMAND` using `patch --forward` so the step is idempotent on CMake re-runs. The upstream SWIG fix is tracked at https://github.com/swig/swig/issues/3407.

### CI

Updated the R version used in CI from 4.5 to 4.6 to catch these issues going forward.

### Pre-commit / `.gitattributes`

Excluded `.patch` files from trailing-whitespace checks (both the `pre-commit` hook and `.gitattributes`) so that patch context lines with significant trailing whitespace are preserved correctly.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -30,6 +30,7 @@
 *.rb             whitespace=tab-in-indent,no-lf-at-eof
 *.R              whitespace=tab-in-indent,no-lf-at-eof
 *.tcl            whitespace=tab-in-indent,no-lf-at-eof
+*.patch          whitespace=-trailing-space
 
 Wrapping/Java/JavaDoc.i              hooks-max-size=2097152
 Wrapping/Python/PythonDocstrings.i   hooks-max-size=2097152
--- a/.github/workflows/BatchBuild.yml
+++ b/.github/workflows/BatchBuild.yml
@@ -219,7 +219,7 @@ jobs:
           fi
 
           if [[ "${{ contains(matrix.ctest-cache, 'WRAP_R') }}" == "true" ]]; then
-            Rscript -e 'install.packages(c("argparser"))'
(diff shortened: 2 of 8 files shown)


dave3d (approved):

zivy (approved):

blowekamp:
The PR branch was split into separate PR including #2589 and #2585

PR #2586, opened 2026-05-05
ENH: Add R wrapping CI coverage to Build and BatchBuild workflows

blowekamp:
- Addresses run-time link error on OSX related to the LoggerBase class.
- Add DEFAULT_R_VERSION and DEFAULT_PYTHON_VERSION top-level env defaults
- Add WRAP_R to to more builds
- Install argparser R package as part of build dependencies when WRAP_R

### DIFF
--- a/.github/workflows/BatchBuild.yml
+++ b/.github/workflows/BatchBuild.yml
@@ -21,6 +21,7 @@ permissions:
 
 env:
   DEFAULT_PYTHON_VERSION: "3.11"
+  DEFAULT_R_VERSION: "4.5"
 
 jobs:
   build:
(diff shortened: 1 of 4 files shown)


dave3d (approved):

.github/workflows/BatchBuild.yml
```diff
@@ -99,16 +100,19 @@ jobs:
...
               WRAP_PYTHON:BOOL=ON
               WRAP_JAVA:BOOL=ON
+              WRAP_R:BOOL=ON
               SimpleITK_USE_ELASTIX:BOOL=ON
 
           - os: ubuntu-24.04
             cmake-build-type: "Release"
             cmake-generator: "Ninja"
             apt-get-dependencies: "liblua5.3-dev lua5.3"
+            r-version: "4.0"
```
zivy, line 110: Any reason for using this old of a version? Possibly use three minor versions back from current 4.6, so 4.3?

zivy (approved):

.github/workflows/BatchBuild.yml
blowekamp: That is what swig was experimenting with as the old supported version. The most recent 3 versions is also reasonable.

PR #2588, opened 2026-05-06
BUG: Patch SWIG superbuild for R 4.6.0 API removals

blowekamp:
R 4.6.0 removed several internal C API symbols (`SET_S4_OBJECT`, `CLOENV`, etc.) that SWIG's generated R bindings depend on. This patch applies fixes to the SWIG source during the superbuild.

### Changes
- Add `swig-r-api-r460.patch` to `SuperBuild/ExternalSource/` so it is managed alongside other external sources
- Use `find_package(Patch REQUIRED)` to locate the `patch` executable portably instead of relying on `patch` being in PATH
- Use `${Patch_EXECUTABLE}` in the `PATCH_COMMAND` of the Swig ExternalProject
- Add `*.patch -whitespace` to `.gitattributes` so trailing whitespace in patch context lines does not trigger the pre-commit hook
- Update SWIG superbuild version to 4.4.1 and include related SHA512 hashes


blowekamp:
Replaced by #2589

PR #2596, opened 2026-05-14
ci: harden Package workflow against supply-chain and injection risks

blowekamp:
Addresses three security issues identified in the Package workflow and associated composite actions.

## Changes

**Prevent shell injection via context expressions** (`publish` job)
Context values (`github.ref_name`, `steps.tag.outputs.name`, etc.) were interpolated directly into bash scripts. They are now passed through intermediate `env:` variables, following [GitHub's recommended practice](https://docs.github.com/en/actions/security-for-github-actions/security-guides/security-hardening-for-github-actions#good-practices-for-mitigating-script-injection-attacks).

**Pin third-party actions to full commit SHAs**
All `actions/checkout`, `actions/setup-python`, `actions/upload-artifact`, `actions/download-artifact`, and `actions/cache` references now use immutable commit SHAs instead of mutable version tags. Covers `Package.yml` and the three composite actions.

**Fix `VCVAR_OPTIONS` interpolation in `cmd` shell** (`package_csharp`, `package_java`, `package_python`)
`${{ inputs.VCVAR_OPTIONS }}` was written directly into `cmd` shell scripts. It is now passed via `env:` and referenced as `%VCVAR_OPTIONS%`.

### DIFF
--- a/.github/actions/external-data-cache/action.yml
+++ b/.github/actions/external-data-cache/action.yml
@@ -13,7 +13,7 @@ inputs:
 runs:
   using: composite
   steps:
-    - uses: actions/cache/restore@v5
+    - uses: actions/cache/restore@27d5ce7f107fe9357f9df03efb73ab90386fccae # v5
       if: inputs.save != 'true'
       with:
         path: ${{ inputs.path }}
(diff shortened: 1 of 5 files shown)


zivy (approved):
Possibly add a comment next to each commit sha indicating what version it corresponds to and saying that the immutable commit sha is used to ensure security and that the version tag which is modifiable originally pointed to that commit?

blowekamp:
> Possibly add a comment next to each commit sha indicating what version it corresponds to and saying that the immutable commit sha is used to ensure security and that the version tag which is modifiable originally pointed to that commit?

I just checked the docs: https://github.com/marketplace/actions/dependabot-sha-comment-action?version=v0.0.4

The version comment has to be on the same line for dependapot to do its things.

PR #2619, opened 2026-07-02
Adding macos-26 build

blowekamp:


zivy (approved):
Is this an intentional explicit specification, macos-26, or is the intent macos-latest which currently is macos-26?

blowekamp:
As with all the other os name specification it is version spec'd. This ensure that the os and requirements don't change at an arbitrary time across branches and ensures stability until the platform spec is EOL'd.

Issue #2505, opened 2026-01-19
BUG: Documentation of resampling using CompositeTransforms inconsistent with results.


gdevenyi:
**Describe the bug**

Please see https://discourse.itk.org/t/simpleitk-compositetransform-versus-antsapplytransform-order-documentation-wrong/7700 for a complete description

aymuos15:
Hi @gdevenyi and @zivy,

Was looking into this. Here's a repro with my thoughts based on the above. I used AI to help me.

**`sitk.CompositeTransform` is behaving as documented. The catch is that the `-t` order on the ANTs command line is not the `AddTransform()` order**, so porting the `-t` list straight into `AddTransform()` calls gives you the opposite composition.

`antsApplyTransforms --verbose` prints the order it actually adds them in. With `reproduce.sh`:

```
$ antsApplyTransforms -d 3 -i $moving -r $fixed \
    -t transform_1Warp.nii.gz -t transform_0GenericAffine.mat -o out.nii.gz --verbose

The composite transform comprises the following transforms (in order):
  1. transform_0GenericAffine.mat (type = AffineTransform)
  2. transform_1Warp.nii.gz (type = DisplacementFieldTransform)
```

The warp was typed first but lands second, because repeated `-t` flags are stored with [`push_front`](https://github.com/ANTsX/ANTs/blob/v2.6.5/Utilities/antsCommandLineOption.cxx#L42) and the composite is [built from the front of that list](https://github.com/ANTsX/ANTs/blob/v2.6.5/Examples/itkantsRegistrationHelper.h#L1283). So ANTs applies the warp first here, then the affine, which is the opposite of what the comment in `reproduce.sh` assumes. 

## Recommended fix

Nothing to change in SimpleITK itself. To match ANTs, add the transforms in reverse of the `-t` order:

```python
composite = sitk.CompositeTransform(3)
composite.AddTransform(affine)    # listed LAST on the ANTs command line, so add FIRST
composite.AddTransform(warp_tx)   # listed FIRST on the ANTs command line, so add LAST
```

Full-volume resample vs. `antsApplyTransforms` (ANTs output mean intensity around 479):

| SimpleITK composite | max abs diff | mean abs diff | correlation |
|---|---|---|---|
| `AddTransform(warp)` then `AddTransform(affine)` (current script) | 3678 | 141.1 | 0.880 |
[...]

PR #2626, opened 2026-07-13
DOC: Note ANTs transform order when porting to CompositeTransform

aymuos15:
Adds a note to the composite transform documentation, as requested in #2505.

`CompositeTransform` behaves as documented; the catch is that the `-t` order on the `antsApplyTransforms` command line is not the `AddTransform()` order. ANTs stores repeated command line flags in reverse, so a `-t` list ported straight into `AddTransform()` calls composes in the opposite order. 

The note also records that the reversal applies to the `-t` flags themselves, not to the sub-transforms within a single composite transform file.

### DIFF
--- a/docs/source/fundamentalConcepts.rst
+++ b/docs/source/fundamentalConcepts.rst
@@ -255,20 +255,6 @@ In the context of registration, if you use a composite transform as the transfor
 that is optimized, only the parameters of the last transformation :math:`T_n` will
 be optimized over.
 
-.. note::
-
-   When porting from the `Advanced Normalization Tools (ANTs) ecosystem
-   <https://github.com/ANTsX>`_, the repeated ``-t`` transforms given to
-   ``antsApplyTransforms`` must be added in **reverse** of their command line
-   order. ANTs prepends each repeated flag to its internal option list, so it
-   builds its composite from the last command line argument to the first::
-
-    antsApplyTransforms -t warp.nii.gz -t affine.mat
-
-   is reproduced by ``CompositeTransform([affine, warp])``. The reversal applies
-   to the ``-t`` flags themselves; if a single ``-t`` names a composite transform
-   file, its sub-transforms are added in their stored order.
-
 Additional Resources
 =====================
 
--- a/docs/source/fundamentalConcepts.rst
+++ b/docs/source/fundamentalConcepts.rst
@@ -233,6 +233,27 @@ In the context of registration, if you use a composite transform as the transfor
 that is optimized, only the parameters of the last transformation :math:`T_n` will
 be optimized over.
 
+.. note::
+
+   When porting from the `Advanced Normalization Tools (ANTs) ecosystem
+   <https://github.com/ANTsX>`_, the repeated ``-t`` transforms given to
+   ``antsApplyTransforms`` must be added in **reverse** of their command line
+   order. ANTs stores repeated flags in reverse and builds its composite from
+   that reversed list, so::
+
+    antsApplyTransforms -t warp.nii.gz -t affine.mat
+
+   is reproduced by::
+
+    composite = CompositeTransform(3)
+    composite.AddTransform(affine)  # listed last on the command line, so added first
+    composite.AddTransform(warp)    # listed first on the command line, so added last
+
+   Running ``antsApplyTransforms --verbose`` prints the transforms in the order
+   ANTs adds them, which is the order to pass to ``AddTransform``. The reversal
+   applies to the ``-t`` flags themselves; if a single ``-t`` names a composite
+   transform file, its sub-transforms are added in their stored order.
+
 Additional Resources
 =====================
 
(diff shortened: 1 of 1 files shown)


docs/source/fundamentalConcepts.rst
```diff
@@ -233,6 +233,27 @@ In the context of registration, if you use a composite transform as the transfor
 that is optimized, only the parameters of the last transformation :math:`T_n` will
 be optimized over.
 
+.. note::
+
+   When porting from the `Advanced Normalization Tools (ANTs) ecosystem
+   <https://github.com/ANTsX>`_, the repeated ``-t`` transforms given to
+   ``antsApplyTransforms`` must be added in **reverse** of their command line
+   order. ANTs stores repeated flags in reverse and builds its composite from
```
blowekamp: Would it be clearer to say "ANTs stores repeated flags in a stack, and build's it composite from the last argument to the first"?

aymuos15: Thank you very much for the review. I agree with the second half, "from the last argument to the first" is clearer than my "from that reversed list". 

On the first half, can I suggest "prepends" over "stack": `AddFunction` does `push_front` into a `std::deque` (`antsCommandLineOption.cxx:42`) and the consumer walks it front-to-back (`itkantsRegistrationHelper.h:1285`). 

So the full sentence would be: "ANTs prepends each repeated flag to its internal option list, so it builds its composite from the last command line argument to the first."

gdevenyi:
Hi, I appreciate you following up on this. Rather than specifically address the ANTs confusion, I think it would be better to just improve the documentation regarding the relationship between the "Add" action, and what order the transforms will be applied.

docs/source/fundamentalConcepts.rst
blowekamp: >can I suggest "prepends" over "stack"
Yes, that is even better.

blowekamp:
> Rather than specifically address the ANTs confusion, I think it would be better to just improve the documentation regarding the relationship between the "Add" action, and what order the transforms will be applied.

Would a description of adding transform `A`, `B`, and `C` to the composite along with the formula of how they apply to a point `x` achieve your suggestion?

gdevenyi:
> > Rather than specifically address the ANTs confusion, I think it would be better to just improve the documentation regarding the relationship between the "Add" action, and what order the transforms will be applied.
> 
> Would a description of adding transform `A`, `B`, and `C` to the composite along with the formula of how they apply to a point `x` achieve your suggestion?

yes that would be fine thanks!

aymuos15:
Thank you for the suggestion. 

I've added a general composite ordering section (`A`, `B`, `C` → `composite(x) = A(B(C(x)))`) and trimmed the ANTs note down to the parts that don't follow from it. Is that trimmed enough, or would you rather drop the ANTs material entirely?

gdevenyi:
I think the ANTs material could be dropped with the clarified translation of the function calls to math.

aymuos15:
Thanks again for the clarification. Should be good to go now.

zivy:
Hello @aymuos15  and @gdevenyi,

It seems that this PR has changed completely from the original ANTs related information @aymuos15 was proposing. 

Looking at the current PR, it is not clear to me how it improves upon the existing text. If you look at the text above the addition you will see the formula and how the transforms are added into the CompositeTransform. I am admittedly too close to this, so I may be missing something (what is natural/clear to the person who wrote the text is often not so clear to others).

Please confirm that after you read the whole text you still feel that the addition helps clarify things. If yes, we'll merge. If not, then we'll close this PR.

Apologies if this seems nitpicky, we're trying our best to keep the documentation complete, yet as streamlined as possible, so that it is fit for human consumption and easy to maintain. AI won't help here unless we want to say the same thing in a much longer form, for example [this AI review](https://github.com/SimpleITK/SimpleITKRInstaller/pull/103#discussion_r3478757073) I got. Valid, but too verbose.

aymuos15:
Thanks for the guidance @zivy I leave it to the maintainers if you do not mind my saying. I was trying to start contributing to the repo (as I need to get familiar with a few things for my personal work) and hence took up a documentation related one from the start. I am not that well versed yet to take a call on something like this I feel (hopefully soon enough!). 

> AI won't help here unless we want to say the same thing in a much longer form, for example https://github.com/SimpleITK/SimpleITKRInstaller/pull/103#discussion_r3478757073 I got. Valid, but too verbose.

Agreed!

Happy to close it.

gdevenyi:
From my perspective, I knew how ANTs' order of application worked and the documentation for SimpleITK was insufficient such that I opened an issue regarding clarification of how CompositeTransform worked. Had this documentation been present at the time I would not have had this issue. I can see how this could be merged as an improved version of the documentation above.

I guess fundamentally my issue may have been that the notebooks did not properly document this which is where I had been learning the process.

zivy:
Hello @aymuos15, @gdevenyi,

I just created [this PR](https://github.com/SimpleITK/SimpleITK/pull/2628) based on @aymuos15's [original PR](https://github.com/SimpleITK/SimpleITK/pull/2626) which had the note about ANTs usage and added some changes to the example code. Hopefully this addresses everyone's concerns. 

@aymuos15 you are listed as co-author on the PR. Is that OK with you?

aymuos15:
Totally cool with me. Thank you very much for considering. Really appreciate it.

zivy:
Superseded by #2628

PR #2633, opened 2026-07-15
Update GitHub download stats script.

zivy:
Starting with release 2.5.5 we distribute R
binaries via GitHub releases on the SimpleITKRInstaller repository. Updated the download stats script to include the R binaries in the analysis.

Also, some minor naming changes (added "github" prefix to filenames and nicer naming of programming languages and OS classes).


dave3d (approved):

blowekamp (approved):
Changes look fine.

Please note that we have now uploading binaries from before 1.0 to GitHub Releases recently.

zivy:
@blowekamp Sadly, the returned release date is the date associated with the GitHub release and not the tag. This causes problems because these old releases appear after newer ones. I looked into retrieving the date based on the tag but that made the script overly complicated, so just updated the code documentation to explain why releases prior to 1.0.0 are omitted. Also, we have zero downloads from GitHub for these releases so not useful for analyzing download counts.

blowekamp:
> Also, we have zero downloads from GitHub for these releases so not useful for analyzing download counts.

In the release pages' asset section I just noticed download counts for the files. These older releases do have a few download each. But I agree these older releases are not useful for these statistics.

PR #2635, opened 2026-07-15
Wrap StructuralSimilarityImageFilter for SimpleITK

blowekamp:
Adds SimpleITK wrapping for `itk::StructuralSimilarityImageFilter`, which computes the Structural Similarity Index Measure (SSIM) between two images (Wang et al., IEEE TIP 2004).

## Summary
- New filter YAML: `Code/BasicFilters/yaml/StructuralSimilarityImageFilter.yaml`
- Wraps `GaussianSigma`, `MaximumKernelWidth`, `K1`, `K2`, `DynamicRange`, `LuminanceExponent`, `ContrastExponent`, `StructureExponent`, and the `MeanSSIM` measurement.
- `ScaleWeights` (multi-scale SSIM) is intentionally not wrapped yet, since ITK's own filter raises an exception for arrays longer than 1 element (not yet implemented upstream).
- 3 test cases: identical-image SSIM=1, two different 3D MRI volumes, and a custom-sigma/K1/K2 case; baseline images and `MeanSSIM` measurement values included.

## Notes
- `itk_module` is `StructuralSimilarity` (no `ITK` prefix), matching the module's own `itk-module.cmake` registration and the convention used by sibling new-filter branches (`FastBilateral`, `AnisotropicDiffusionLBR`).
- Verified: clean build against a local ITK checkout with `Module_StructuralSimilarity=ON`, and all 3 test cases pass, including the `MeanSSIM` measurement assertions.
- Test baseline images uploaded to SimpleITK/SimpleITKExternalData#47 (must be merged before CI can download them via S3).


zivy (approved):
Nice addition. Wasn't aware that this filter was available in ITK.

dave3d (approved):

PR #2636, opened 2026-07-16
PERF: Resolve pixel type dispatch once in Image.__iter__

aymuos15:
Iterating over an `Image` in Python yields each pixel through `__getitem__` → `GetPixel`, which calls `GetPixelIDValue()` and walks a chain of up to 24 pixel-type comparisons for every pixel, even though the pixel type is fixed for the image.

This PR adds a private, class-cached map from pixel ID to the low-level `__GetPixelAs*__` accessor name. `Image.__iter__` now resolves the bound accessor once before the loop and yields pixels through it directly. `GetPixel` and `__getitem__` are unchanged, and the iterator's error behavior is preserved (`invalid pixel type` for `sitkUnknown`, `unknown pixel type` otherwise).

## Benchmarks

Python 3.10, Linux. Old = SimpleITK 2.5.5 wheel (current implementation); new = patched method bodies calling the same SWIG accessors. Median of repeated runs.

| Image | Old | New | Speedup |
|---|---|---|---|
| 512×512 UInt8 | 1001 ms | 293 ms | 3.4x |
| 1024×1024 Float32 | 4560 ms | 1290 ms | 3.5x |
| 64×64 VectorFloat64 (3 comp.) | 19.5 ms | 5.6 ms | 3.5x |
| 512×512 ComplexFloat64 | 1133 ms | 302 ms | 3.8x |
| 128×128×128 Int16 | 8997 ms | 2546 ms | 3.5x |
| 256×256×64 Float64 | 18925 ms | 5144 ms | 3.7x |
| 48×48×48×16 Float32 (4D) | 8254 ms | 2138 ms | 3.9x |
| 4×4 UInt8, 20,000 repeated iterations | 1330 ms | 445 ms | 3.0x |

<sub>Benchmarked on an Intel Core i7-12800H, 32 GB RAM, Ubuntu (Linux 6.8), Python 3.10.12, CPython single-threaded.</sub>

<details>
<summary>Benchmark script</summary>

Runs against an installed SimpleITK ≤ 2.5.5 wheel: `list(img)` is the current implementation, `new_iter` replicates the patched `__iter__` calling the same SWIG accessors.

```python
import time, statistics
import SimpleITK as sitk

GETTERS = {
    sitk.sitkUInt8: "__GetPixelAsUInt8__",
    sitk.sitkInt16: "__GetPixelAsInt16__",
    sitk.sitkFloat32: "__GetPixelAsFloat__",
    sitk.sitkFloat64: "__GetPixelAsDouble__",
    sitk.sitkVectorFloat64: "__GetPixelAsVectorFloat64__",
    sitk.sitkComplexFloat64: "__GetPixelAsComplexFloat64__",
}

def new_iter(img):  # patched __iter__: accessor resolved once, not per pixel
    dim, size = img.GetDimension(), img.GetSize()
    idx = [0] * dim
    getter = getattr(img, GETTERS[img.GetPixelIDValue()])
    while idx[dim - 1] < size[dim - 1]:
        yield getter(idx)
        for d in range(dim):
            idx[d] += 1
            if idx[d] >= size[d] and d != dim - 1:
                idx[d] = 0
            else:
                break

def median_time(fn, reps=3):
    ts = []
    for _ in range(reps):
        t0 = time.perf_counter()
        fn()
        ts.append(time.perf_counter() - t0)
    return statistics.median(ts)

CASES = [
    ("512x512 UInt8", sitk.Image(512, 512, sitk.sitkUInt8)),
    ("1024x1024 Float32", sitk.Image(1024, 1024, sitk.sitkFloat32)),
    ("64x64 VectorFloat64x3", sitk.Image([64, 64], sitk.sitkVectorFloat64, 3)),
    ("512x512 ComplexFloat64", sitk.Image(512, 512, sitk.sitkComplexFloat64)),
    ("128^3 Int16", sitk.Image(128, 128, 128, sitk.sitkInt16)),
    ("256x256x64 Float64", sitk.Image(256, 256, 64, sitk.sitkFloat64)),
    ("48^3x16 Float32 (4D)", sitk.Image([48, 48, 48, 16], sitk.sitkFloat32)),
]

for name, img in CASES:
    assert list(img) == list(new_iter(img))  # identical results
    old = median_time(lambda: list(img))
    new = median_time(lambda: list(new_iter(img)))
    print(f"{name}: old {old:6.2f}s  new {new:6.2f}s  speedup {old/new:.2f}x")
```

</details>

### DIFF
--- a/Wrapping/Python/sitkImage.i
+++ b/Wrapping/Python/sitkImage.i
@@ -580,6 +580,54 @@
 
         # iterator and container methods
 
+        @classmethod
+        def _pixel_getter_name(cls, pixelID):
+            """Returns the name of the low-level per-type pixel getter method
+               for a pixel ID. The map is built once and cached on the class."""
+
+            if pixelID == sitkUnknown:
+              raise Exception("invalid pixel type")
+
+            try:
+              getter_map = cls._pixel_getter_name_map
+            except AttributeError:
+              suffixes = {
+                sitkInt8: "Int8",
+                sitkUInt8: "UInt8",
+                sitkLabelUInt8: "UInt8",
+                sitkInt16: "Int16",
+                sitkUInt16: "UInt16",
+                sitkLabelUInt16: "UInt16",
+                sitkInt32: "Int32",
+                sitkUInt32: "UInt32",
+                sitkLabelUInt32: "UInt32",
+                sitkInt64: "Int64",
+                sitkUInt64: "UInt64",
+                sitkLabelUInt64: "UInt64",
+                sitkFloat32: "Float",
+                sitkFloat64: "Double",
+                sitkVectorInt8: "VectorInt8",
+                sitkVectorUInt8: "VectorUInt8",
+                sitkVectorInt16: "VectorInt16",
+                sitkVectorUInt16: "VectorUInt16",
+                sitkVectorInt32: "VectorInt32",
+                sitkVectorUInt32: "VectorUInt32",
+                sitkVectorInt64: "VectorInt64",
+                sitkVectorUInt64: "VectorUInt64",
+                sitkVectorFloat32: "VectorFloat32",
+                sitkVectorFloat64: "VectorFloat64",
+                sitkComplexFloat32: "ComplexFloat32",
+                sitkComplexFloat64: "ComplexFloat64"
+              }
+              getter_map = cls._pixel_getter_name_map = {
+                id: "__GetPixelAs{0}__".format(s) for id, s in suffixes.items()
+              }
+
+            try:
+              return getter_map[pixelID]
+            except KeyError:
+              raise Exception("unknown pixel type") from None
+
         def __iter__( self ):
 
             if len(self) == 0:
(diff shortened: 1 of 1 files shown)


zivy:
Hi @aymuos15,

What was the use case that required iterating over pixel locations to obtain the intensities? 

Generally speaking, iterating in interpreted languages is discouraged, most operations can be done via SimpleITK methods that do the iteration and possibly multi-threading under the hood in C++ (e.g. adding two images via the SimpleITK operator/ Add filter vs. iterating per pixel and adding them yourself).

aymuos15:
I was scoping the codebase in general and the AI pinged that this might be a performance improvement area. Hence went ahead with it after I verified it. 

Now that you mention it, I do realise this is not the way to do things. I assumed this was used somewhere since it is still in the codebase and hence went ahead with this. But having had a broader look now (and your reply) I realise that is not the case. 

Happy to close it. Thank you for taking the time to explain it.

zivy:
Hi @aymuos15,

It seems that your PR inspired @blowekamp to look into this, see [this PR](https://github.com/SimpleITK/SimpleITK/pull/2638). 

While we discourage usage of the `GetPixel` method in an iterative fashion across the image pixels, given that it has been publicly available in the API for so long, it may be worth improving its efficiency. Please close this PR. Feel free to discuss with @blowekamp if you have concerns about his PR.

Thanks again for looking into SimpleITK (hopefully you are also learning something useful).

aymuos15:
I am learning a lot! Thank you!

Will look into that PR tomorrow then, awesome that there is something to look into here. Thanks again!

PR #2640, opened 2026-07-16
Replace buggy bash commit-msg hook with Python implementation

blowekamp:
The bash `Utilities/Hooks/kw-commit-msg` hook has had a regression since 2024-03-19 (commit 7994b76245, "Update KW hooks. Remove gerrit and hook chaining"). That commit removed Gerrit review support and, as a side effect, also deleted the only state transition (`else: state=rest`) that allowed a commit message body after the mandatory blank second line. Since then, the hook has rejected **any** commit message with real body content beyond that blank line — every line after it gets re-validated by the "must be empty" check forever.

This explains why so many historical commits with proper multi-paragraph bodies exist in this repo's history: they were made via GitHub's server-side squash/merge (which never runs local hooks), with `--no-verify`, or in checkouts where pre-commit wasn't installed — never through this hook actually passing.

## Fix
Replace the bash script with `Utilities/Hooks/commit-msg.py`, adapted from ITK's own Python commit-message hook implementation:
- Validates subject line length (8-78 chars, configurable via `git config hooks.commit-msg.maxLength`) and no leading/trailing whitespace
- Requires a blank second line
- Allows a free-form body afterward (the actual fix)
- Drops ITK's mandatory subject-line prefix convention (`BUG:`/`ENH:`/etc.), which SimpleITK does not use
- Fixes a missing `return` in `get_max_length()` present in ITK's version
- Drops the `kw-` prefix from the hook id/name/filename

`.pre-commit-config.yaml` is updated to invoke it with `language: python` instead of `language: system`.

## Testing
Manually verified against sample commit messages (multi-paragraph body, subject-only, too-short subject, too-long subject, missing blank second line) — all pass/fail as expected. This commit's own message is a multi-paragraph body, validated by the new hook.


dave3d (approved):

Utilities/Hooks/commit-msg.py
```diff
@@ -0,0 +1,154 @@
...
+#   Unless required by applicable law or agreed to in writing, software
+#   distributed under the License is distributed on an "AS IS" BASIS,
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+#
+# ==========================================================================
+#
+# Validate the layout of a commit message: a non-empty subject line
+# (8-78 characters, no leading/trailing space) followed by a blank line
```
zivy, line 21: 8 characters seems a bit too short. Possibly check for more than x=4 words? `len(first_line.split(" "))>4`

zivy (dismissed):

Utilities/Hooks/commit-msg.py
blowekamp: I don't see a need to diverge from ITK for this.

PR #2662, opened 2026-08-04
CI: Enable dependabot for GitHub Actions on release branch

blowekamp:
Add a second dependabot update entry targeting the `release` branch so GitHub Actions dependencies are kept current there too, matching the existing entry for `main`.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,3 +4,8 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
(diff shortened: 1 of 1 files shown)


zivy (approved):

blowekamp:
Applied review suggestion: Removed explicit `target-branch: "main"` from the first entry to rely on the default branch behavior. This avoids hard-coding the branch name and makes the config more flexible if the repository's default branch changes in the future. Only the `release` branch entry now has an explicit target-branch.

PR #2670, opened 2026-08-10
Change TWeights in MultiLabelSTAPLE to double

thomas-albrecht:
The WeightsType `TWeights` of the `itk::MultiLabelSTAPLEImageFilter` is set to the default `float`. 
There is one big problem with this. 

In [itkMultiLabelSTAPLEImageFilter.hxx](https://github.com/InsightSoftwareConsortium/ITK/blob/c01976d239314ab9b9251410232cf35ea569021c/Modules/Segmentation/LabelVoting/include/itkMultiLabelSTAPLEImageFilter.hxx#L209)
the frequency of the labels is computed like this: 
``` 
      for (in.GoToBegin(); !in.IsAtEnd(); ++in)
      {
        ++(this->m_PriorProbabilities[in.Get()]);
      }
```
where this->m_PriorProbabilities is of course an `Array<WeightsType>`. 
Ultimately, the code is doing this: 
```
float frequency = 0;
for (int i = 0; i < num_voxels; i++) {
  frequency = frequency + 1
}
```
At some point, precisely at $frequency = 2^{24} = 16,777,216$, the float representation of frequency and frequency + 1 is the same and the counter stops increasing. This number easily reached in even a modest 3D images, and the whole `m_PriorProbabilities` will be completely wrong. 

Short of changing the ITK code, a simple fix is to use double as the weights type, which raises this problematic threshold to $2^{53} = 9,007,199,254,740,992$ (over 9 quadrillion).

### DIFF
--- a/Code/BasicFilters/yaml/MultiLabelSTAPLEImageFilter.yaml
+++ b/Code/BasicFilters/yaml/MultiLabelSTAPLEImageFilter.yaml
@@ -3,7 +3,7 @@ template_code_filename: MultiInputImageFilter
 template_test_filename: ImageFilter
 number_of_inputs: 1
 pixel_types: UnsignedIntegerPixelIDTypeList
-filter_type: itk::MultiLabelSTAPLEImageFilter<InputImageType, OutputImageType, float>
+filter_type: itk::MultiLabelSTAPLEImageFilter<InputImageType, OutputImageType, double>
 members:
 - name: LabelForUndecidedPixels
   type: uint64_t
@@ -43,8 +43,8 @@ members:
   detaileddescriptionGet: >-
     Set maximum number of iterations.
 - name: PriorProbabilities
-  type: std::vector<float>
-  default: std::vector<float>()
+  type: std::vector<double>
+  default: std::vector<double>()
   briefdescriptionSet: ''
   custom_itk_cast: |-
     if (!m_PriorProbabilities.empty()) filter->SetPriorProbabilities(typename FilterType::PriorProbabilitiesType(&this->m_PriorProbabilities[0],this->m_PriorProbabilities.size()));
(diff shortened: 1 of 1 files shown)


blowekamp:
Improvement to ITK proposed here: InsightSoftwareConsortium/ITK#6759

thomas-albrecht:
FYI: I also made a pull request in ITK to fix the problem at the source: https://github.com/InsightSoftwareConsortium/ITK/pull/6759 

But that would probably take quite a while to percolate into the SimpleITK python package.

blowekamp:
> FYI: I also made a pull request in ITK to fix the problem at the source: [InsightSoftwareConsortium/ITK#6759](https://github.com/InsightSoftwareConsortium/ITK/pull/6759)
> 
> But that would probably take quite a while to percolate into the SimpleITK python package.

SimpleITK's main is underdevelopment for a 3.0 release against ITK 6.0. We are hopeful that will be by the end of the year. There should be a beta sooner, and there are the nightly builds too. However, I am inclined to wait for the fix from ITK for the main branch.

I have back ported this file to the release branch to the 2.5 patch releases. See #2673 That may be published sooner, at the end of the month.

Please provide feedback as what might work for you to guide our patching.

thomas-albrecht:
I actually think the best approach may be to implement this fix in SimpleITK 2.5 and 3.0, and also implement the fix in ITK. 

The root cause is that incrementing a floating point number using `++` is just the wrong thing to do. This PR here in SimpleITK mitigates the problem by using a double instead of a float, whereas the fix in ITK fixes it at the core by counting with integer numbers. 

When both fixes come together in SimpleITK 3.0 there can be no harm. When the float values will be correct with the ITK fix in place, the double values will also be correct. The only downside is that the filter would use a little more memory, but this is about a few vectors and matrices of size `num_labels x (num_lables + 1)` and not entire images with pixel type `double`.

blowekamp (approved):
LGTM

Issue #2683, opened 2026-08-24
Feature request: an outputImage argument to GetImageFromArray


vboussot:
This is the feature request asked for in [Is it possible to get a writable array view of a SimpleITK image?](https://discourse.itk.org/t/is-it-possible-to-get-a-writable-array-view-of-a-simpleitk-image/1585) (8 February 2019). @blowekamp wrote there:

> I think we could very easily take an image as an argument and then "fill" it with the nparray. This would both be more efficient and very safe.

and then:

> Can you please create a feature request on GitHub to add an "outputImage" to the `GetImageFromArray` method.

I could not find one, so here it is. I am happy to send the PR.

### What it would change

```python
def GetImageFromArray(arr, isVector=None, outputImage=None) -> Image:
```

With `outputImage`, the array is written into the image the caller passes instead of into a newly constructed one. Everything else is unchanged, and `outputImage=None` keeps today's behaviour exactly.

### Why it helps

`GetImageFromArray` constructs a zero-initialised image and then overwrites all of it:

```python
img = Image(shape, id, number_of_components)   # allocated and zero-filled
_SetImageFromArray(z, img)                     # entirely overwritten
```

For a caller converting one array after another **of the same shape and pixel type**, that is a fresh mapping plus a zero-fill per array, both discarded. The zero-fill is a wasted pass; the fresh mapping costs more, because it pays first-touch page faults rather than running at memory bandwidth.

Measured on a 384x384x384 uint16 array (113 MB), 11 repetitions, Linux, SimpleITK 2.5.5, Python 3.13:

| | min | median | rate |
|---|---|---|---|
| `GetImageFromArray(arr)` | 45.2 ms | 48.7 ms | 2.5 GB/s |
| `_SetImageFromArray` into a reused image, with shape and pixel-type validation | 7.9 ms | 8.3 ms | 14 GB/s |

5.7x on the minimum, 5.9x on the median.

The gain is entirely in reusing the buffer, so the scope is a loop over arrays of one shape and pixel type: slice-by-slice work, a batch of equal-size patches, a series over time. A loop whose shapes vary gets nothing from it, since it needs a new image either way.

### Why it is small

`_SetImageFromArray` already fills an existing image, so the work is in the Python wrapper.

One thing worth building in: `_SetImageFromArray` validates the **byte length** only, so an image whose buffer happens to be the same size is accepted and the bytes are reinterpreted silently. On one 240-byte source array of shape (4, 5, 6) uint16:

[...]

PR #2686, opened 2026-08-25
ENH: Add outputImage argument to GetImageFromArray

vboussot:
Implements #2683.

`GetImageFromArray(arr, isVector=None, outputImage=None)`. When `outputImage` is given, the array is copied into that image and that image is returned. Without it, nothing changes.

The image has to match what the array converts to: size, pixel ID, and number of components per pixel. Anything else raises `ValueError`. `_SetImageFromArray` checks the total number of bytes only, so an image with the same byte count but a different pixel type, a different shape, or transposed axes gets filled silently today.

The image keeps its origin, spacing, direction and metadata. Copy-on-write still holds: if the image shares its buffer with another `Image`, that other image keeps its values and the call allocates a new buffer.

### Numbers

384³ uint16 (113 MB), best and median of 11 to 21 repetitions, with this patch applied to the `3.0.0b1.post145` dev wheel, on a Linux box under load (1-minute average 7 to 10):

| run | fresh | outputImage | speedup |
|---|---|---|---|
| 1 (11 reps) | 52.3 / 64.2 ms | 6.1 / 8.1 ms | 8.5x / 7.9x |
| 2 (21 reps) | 59.6 / 67.7 ms | 8.4 / 11.3 ms | 7.1x / 6.0x |
| 3 (21 reps) | 61.6 / 67.7 ms | 8.5 / 8.9 ms | 7.3x / 7.6x |

That lines up with the 7.9x measured in the issue. Running a `Resample` on the image between two fills does not force a reallocation: those fills stay at 5.9 to 7.7 ms.

### Tests

Ten cases in `sitkNumpyArrayConversionTest.py`: the image is filled and returned as the same object, can be filled again, and keeps its geometry; the result matches a fresh conversion for `isVector` None, False and True; a mismatched image is rejected for a different size, transposed axes, a different dimension, a different pixel type, a pixel type of the same byte count, and a scalar against a vector image; vector images work, including a wrong component count; and an image sharing its buffer is left alone.

They fail without the change and pass with it, and the whole `Wrapping/Python/tests` suite is 390 passed, 1 skipped against the dev wheel.

### DIFF
--- a/Wrapping/Python/SimpleITK/extra.py
+++ b/Wrapping/Python/SimpleITK/extra.py
@@ -281,12 +281,23 @@ def GetArrayFromImage(image: Image) -> "numpy.ndarray":
     return numpy.array(image, copy=True)
 
 
-def GetImageFromArray(arr: "numpy.ndarray", isVector: Optional[bool] = None) -> Image:
+def GetImageFromArray(
+    arr: "numpy.ndarray",
+    isVector: Optional[bool] = None,
+    outputImage: Optional[Image] = None,
+) -> Image:
     """Get a SimpleITK Image from a numpy array.
 
     If isVector is True, then the Image will have a Vector pixel type, and the last dimension of the array will be
     considered the component index. By default when isVector is None, 4D arrays
     are automatically considered 3D vector images, but 3D arrays are 3D images.
+
+    If outputImage is given, the array is copied into that image and that image is returned, instead of a newly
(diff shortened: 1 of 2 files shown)


vboussot:
Closing: the discussion here landed on the PEP 688 buffer protocol (memoryview / np.asarray(img, copy=False)) as the better route than an outputImage kwarg, which this PR implements. Superseded by that.

PR #2698, opened 2026-08-28
CI: PyPI test-publish target and rerun-safety fixes

blowekamp:
## Summary
- Switch `publish-pypi` to the `pypi-test` GitHub environment and parameterize the repository URL via `vars.PYPI_REPOSITORY_URL` instead of hardcoding test.pypi.org.
- Add `skip-existing: true` to the PyPI publish step so a workflow rerun after a transient failure doesn't fail on files already uploaded to the index.
- Remove stale `# vN` version comments after action hashes, since dependabot updates the pinned hash but not the trailing comment, leaving it out of sync.

## Test plan
- [ ] Trigger the `Package` workflow via `workflow_dispatch` or a test tag push and confirm `publish-pypi` runs against the `pypi-test` environment/URL.
- [ ] Confirm action steps still resolve correctly with the version comments removed (no functional change expected, hash is unchanged).

### DIFF
--- a/.github/workflows/Package.yml
+++ b/.github/workflows/Package.yml
@@ -401,7 +401,8 @@ jobs:
   publish-pypi:
     if: github.event_name != 'pull_request' && startsWith(github.ref_name, 'v')
     environment:
-      name: release
+      name: pypi-test
+      url: ${{ vars.ENVIRONMENT_URL }}
     permissions:
       id-token: write  # for PyPI publish
     needs:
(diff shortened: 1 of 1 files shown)


zivy (approved):
I'm not sure removal of the trailing comment in the uses action is the correct thing to do. I dug a bit into this and it looks to me like depenedabot is expecting a full version vMAJOR.MINOR.PATCH. See [this issue](https://github.com/dependabot/dependabot-core/pull/13007).

PR #2711, opened 2026-09-02
BUG: Add regression test for HDF5 vector-pixel-type read failure

### LINKED ISSUE #2702: Reading a vector-pixel-type image from HDF5 fails: Unknown PixelType
**Describe the bug**

Writing a multi-component (vector) image to HDF5 succeeds, but reading it back fails with `Unknown PixelType: float(11)`. Round-tripping a scalar image through the same `HDF5ImageIO` works fine, so this is specific to vector/multi-component pixel types.

Root cause (traced in `ITK/Modules/IO/HDF5/src/itkHDF5ImageIO.cxx`, `HDF5ImageIO::ReadImageInformation()`): when the on-disk dataset has more dimensions than the image's spatial dimensions, the code infers multiple components and calls `SetNumberOfComponents(Dims[nDims - 1])`, but it never calls `SetPixelType(...)` to mark the image as `IOPixelEnum::VECTOR`. `PixelType` is left at its default (`SCALAR`).

Downstream, `SimpleITK`'s `ImageReaderBase::GetPixelIDFromImageIO()` (`Code/IO/src/sitkImageReaderBase.cxx:206-237`) dispatches purely on `(numberOfComponents, pixelType)`:
- `numberOfComponents == 1 && pixelType == SCALAR` → scalar path
- `pixelType == VECTOR || RGB || ...` → vector path
- otherwise → throws `Unknown PixelType: <componentType>(<n>)`

Since `HDF5ImageIO` reports `numberOfComponents == 2` (or more) together with `pixelType == SCALAR`, it matches neither branch and hits the exception. The error message is also a bit misleading — it prints the *component* type/enum value (`float(11)`), not the actual problem (a `pixelType`/`numberOfComponents` mismatch), which made this harder to diagnose than it should be.

**To Reproduce**

1. Operating system, version, and architecture
   - OS: macOS 26 (Darwin 25.6.0)
   - Architecture: Apple ARM64 (M-series)
2. Programming language: Python 3.13.14
3. SimpleITK version: `3.0.0b1.post184-gfd9de (ITK 6.0)` (built from source at ITK main, commit `7fae0b606b`) — but this is a structural bug in `itk::HDF5ImageIO`, not version-pinned to a specific recent change; likely reproduces on any ITK 5.x/6.x with HDF5 support enabled.
4. Built from source (SuperBuild / FetchContent against ITK main).
5. Minimal repro:

```python
import SimpleITK as sitk

img = sitk.Image([4, 4], sitk.sitkVectorFloat32, 2)
sitk.WriteImage(img, "repro.hdf5")   # succeeds
back = sitk.ReadImage("repro.hdf5")  # raises
```

Output:
```
RuntimeError: Exception thrown in SimpleITK ImageFileReader_Execute: .../sitkImageReaderBase.cxx:237:
ITK ERROR: sitk::ERROR: Unknown PixelType: float(11)
```

For comparison, the scalar case round-trips fine:
```python
scalar = sitk.Image([4, 4], sitk.sitkFloat32)
sitk.WriteImage(scalar, "repro_scalar.hdf5")
sitk.ReadImage("repro_scalar.hdf5")  # OK
```

**Expected behavior**

`ReadImage` should successfully round-trip a vector-pixel-type image written to `.hdf5`/`.h5`, the same way it does for `.mha`, `.nrrd`, and `.nii.gz`.

**Additional context**

- This surfaced while investigating whether HDF5 would be a good alternative container for a vector-float (displacement field) test baseline image. Aside from this round-trip bug, HDF5's compression was also noticeably weaker than deflate-based formats for this kind of data (best case ~2.3% smaller than uncompressed; MHA/NRRD/NIfTI all achieve ~11-12% smaller via zlib), so it wasn't going to be adopted regardless — but the round-trip failure itself seems like a genuine, independent bug worth fixing.
- The fix likely belongs in `itk::HDF5ImageIO::ReadImageInformation()` (call `SetPixelType(IOPixelEnum::VECTOR)` — or similar — alongside the existing `SetNumberOfComponents()` call when a trailing dataset dimension is detected), i.e. upstream in ITK rather than in SimpleITK itself. Filing here since it's user-visible through SimpleITK's `ReadImage`/`WriteImage`; happy to help verify a fix or file the corresponding ITK issue if that's preferred.

blowekamp:
## Summary

Fixes #2702: reading a vector/multi-component pixel image back from HDF5 raised `Unknown PixelType: float(11)`, because `itk::HDF5ImageIO::ReadImageInformation()` inferred multiple components from a trailing HDF5 dataset dimension but never called `SetPixelType(VECTOR)`, leaving `PixelType` at its default `SCALAR`. `ImageReaderBase::GetPixelIDFromImageIO()` dispatches on `(numberOfComponents, pixelType)` together and can't handle that inconsistent state.

- Adds `TEST(IO, HDF5_ReadWriteVectorPixelType)` in `Testing/Unit/sitkImageIOTests.cxx`, round-tripping a `sitkVectorFloat32` image (2 components) through `.hdf5` and checking pixel type, component count, and hash equality. The test skips gracefully if `HDF5ImageIO` isn't registered in the build.
- The actual fix belongs upstream in ITK: see [InsightSoftwareConsortium/ITK#6829](https://github.com/InsightSoftwareConsortium/ITK/pull/6829), which adds the missing `SetPixelType(IOPixelEnum::VECTOR)` call.

This test currently requires the linked ITK PR to pass; it will fail against ITK versions without that fix.

## Test plan

- [x] Confirmed the test fails with the exact reported error (`Unknown PixelType: float(11)`) against unpatched ITK.
- [x] Confirmed the test passes once built against the patched ITK from InsightSoftwareConsortium/ITK#6829.
- [x] Full `IO.*` SimpleITK unit test suite (20 tests) still passes with the patched ITK — no regressions.

### DIFF
--- a/Testing/Unit/sitkImageIOTests.cxx
+++ b/Testing/Unit/sitkImageIOTests.cxx
@@ -463,6 +463,40 @@ TEST(IO, ReadWrite)
 }
 
 
+// Regression test for https://github.com/SimpleITK/SimpleITK/issues/2702
+// Round-tripping a vector-pixel-type image through HDF5 raised
+// "Unknown PixelType" on read.
+TEST(IO, HDF5_ReadWriteVectorPixelType)
+{
+  namespace sitk = itk::simple;
+
+  sitk::ImageFileWriter writer;
+
+  const std::vector<unsigned int> size{ 5u, 7u };
+  sitk::Image                     image(size, sitk::sitkVectorFloat32, 3u);
+
+  for (unsigned int y = 0; y < size[1]; ++y)
+  {
+    for (unsigned int x = 0; x < size[0]; ++x)
+    {
+      image.SetPixelAsVectorFloat32({ x, y }, { static_cast<float>(x), static_cast<float>(y), 1 });
+    }
+  }
+
+  const std::string filename = dataFinder.GetOutputFile("IO.HDF5_ReadWriteVectorPixelType.hdf5");
+
+  ASSERT_NO_THROW(sitk::WriteImage(image, filename));
+  ASSERT_TRUE(dataFinder.FileExists(filename));
+
+  sitk::Image result;
+  ASSERT_NO_THROW(result = sitk::ReadImage(filename));
+
+  EXPECT_EQ(result.GetPixelID(), sitk::sitkVectorFloat32);
+  EXPECT_EQ(result.GetNumberOfComponentsPerPixel(), 3u);
+  EXPECT_EQ(sitk::Hash(image), sitk::Hash(result));
+}
+
+
 TEST(IO, 2DFormats)
 {
   itk::simple::HashImageFilter hasher;


Testing/Unit/sitkImageIOTests.cxx
```diff
@@ -463,6 +463,46 @@ TEST(IO, ReadWrite)
...
+TEST(IO, HDF5_ReadWriteVectorPixelType)
+{
+  namespace sitk = itk::simple;
+
+  sitk::ImageFileWriter    writer;
+  std::vector<std::string> ios = writer.GetRegisteredImageIOs();
+  if (std::find(ios.begin(), ios.end(), "HDF5ImageIO") == ios.end())
+  {
+    GTEST_SKIP() << "HDF5ImageIO is not registered in this build.";
+  }
```
zivy: I'm not sure about this defensive approach, checking that the HDF5 IO is available. HDF5 is expected to be available in all ITK/SimpleITK default builds, so something is probably wrong if it isn't and we wouldn't want to silently skip this. I'd remove this code. Am I missing something here?

Testing/Unit/sitkImageIOTests.cxx
```diff
@@ -463,6 +463,46 @@ TEST(IO, ReadWrite)
...
+
+  const std::vector<unsigned int> size{ 4u, 4u };
```
zivy: AI seems to be fixated on 4x4 images? That's an arbitrary magic number that a human would suspect means something. 

We should use 1x1 which a human understands as "smallest possible image" or 42x42 which a human understands as "geek developers, no real meaning to size".

At least on the vector size the AI used the smallest possible number, 2.

blowekamp: I am not sure this was worth writing or continuing, but this test file is full of arbitrary sized images. Looks like 10x10 was used frequently. I don't see 4 much differently as an arbitrary number. I don't think it Is worth documenting each number as arbitrary either.

Testing/Unit/sitkImageIOTests.cxx
blowekamp: removed.

Testing/Unit/sitkImageIOTests.cxx
blowekamp: Changed to arbitrary prime numbers, which can be useful at detecting errors in indexing.

zivy (approved):

PR #2712, opened 2026-09-02
ENH: Improve exception message for PixelType/NumberOfComponents mismatch

blowekamp:
## Summary

`ImageReaderBase::GetPixelIDFromImageIO()` previously reported any unhandled `(numberOfComponents, pixelType)` combination from an `ImageIO` as:

```
Unknown PixelType: float(11)
```

This only prints the *component* type, not the actual problem: an inconsistent `PixelType`/`NumberOfComponents` state reported by the `ImageIO`. This came up directly while diagnosing #2702 (HDF5 vector-pixel-type round-trip failure, fixed upstream in ITK: [InsightSoftwareConsortium/ITK#6829](https://github.com/InsightSoftwareConsortium/ITK/pull/6829)) — the error message gave no hint that a mismatch, rather than the component type, was the cause, which made the bug harder to diagnose than necessary.

The exception now reports the `PixelType`, `NumberOfComponents`, `ComponentType`, and the offending `ImageIO`'s class name, e.g.:

```
Unable to determine PixelType from ImageIO: PixelType="scalar" (1), NumberOfComponents=3, ComponentType="float" (9). This is an inconsistent state reported by the ImageIO, likely indicating a bug in "HDF5ImageIO".
```

## Changes

- `Code/IO/src/sitkImageReaderBase.cxx`: expand the diagnostic exception message in `GetPixelIDFromImageIO()`.
- `Testing/Unit/sitkImageIOTests.cxx`: add `TEST(IO, ImageFileReader_InconsistentPixelTypeExceptionMessage)`, which registers a minimal fake `itk::ImageIOBase`/`ObjectFactoryBase` reporting a deliberately inconsistent `(NumberOfComponents=3, PixelType=SCALAR)` state (via `itk::ObjectFactoryBase::RegisterOverride`), so the diagnostic path is exercised directly without depending on a real `ImageIO` having a matching bug.

## Test plan

- [x] New test `IO.ImageFileReader_InconsistentPixelTypeExceptionMessage` passes, verifying the message contains the pixel type, component count, component type, and IO class name.
- [x] Full `IO.*` SimpleITK unit test suite (20 tests) passes, no regressions.

### DIFF
--- a/Testing/Unit/sitkImageIOTests.cxx
+++ b/Testing/Unit/sitkImageIOTests.cxx
@@ -26,8 +26,131 @@
 #include <sitkExtractImageFilter.h>
 #include <sitkRegionOfInterestImageFilter.h>
 
+#include <itkImageIOBase.h>
+#include <itkObjectFactoryBase.h>
+#include <itkCreateObjectFunction.h>
+#include <itkVersion.h>
+
 #include <cmath>
 #include <itksys/SystemTools.hxx>
+
+namespace
+{
+// A minimal ImageIOBase that reports an inconsistent (numberOfComponents>1,
+// pixelType==SCALAR) state, to exercise ImageReaderBase::GetPixelIDFromImageIO's
+// diagnostic exception for that case, without relying on a real ImageIO having a bug.
+class InconsistentPixelTypeImageIO : public itk::ImageIOBase
+{
+public:
+  using Self = InconsistentPixelTypeImageIO;
+  using Superclass = itk::ImageIOBase;
+  using Pointer = itk::SmartPointer<Self>;
+
+  itkOverrideGetNameOfClassMacro(InconsistentPixelTypeImageIO);
+  itkNewMacro(Self);
+
+  bool
+  CanReadFile(const char *) override
+  {
+    return true;
+  }
+  bool
+  CanWriteFile(const char *) override
+  {
+    return false;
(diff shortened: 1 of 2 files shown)


Testing/Unit/sitkImageIOTests.cxx
```diff
@@ -26,9 +26,132 @@
...
+  {
+    return false;
+  }
+
+  void
+  ReadImageInformation() override
+  {
+    this->SetNumberOfDimensions(2);
+    this->SetDimensions(0, 4);
+    this->SetDimensions(1, 4);
```
zivy: The choice of setting it to a 4x4 "image" is a bit strange to me, causes me to think that there must be some hidden reason for this selection. I know 1x1 is as arbitrary, but it makes more sense to me as a human, clearer that it is arbitrary and there is no hidden reason here. Possibly change? Conversely, keep as is and add comment saying "arbitrary sized 4x4 image"?

zivy (approved):

dave3d (approved):

Testing/Unit/sitkImageIOTests.cxx
blowekamp: Changed to 1x1 — agreed that's clearer there's no hidden significance to the size. Amended in 5d7898281.

PR #2714, opened 2026-09-04
CI: gated production PyPI publish via pypi-release environment

blowekamp:
## Summary

- Fix `concurrency` so superseded pull request runs are actually cancelled (grouped by PR number, which is stable across pushes, instead of head SHA which changes every push). Tag/release runs are never auto-cancelled.
- Consolidate the release-tag classification (final vs pre-release vs `latest`) into a new `check-release-tag` job, and have `publish-github` reuse its `is_prerelease` output instead of duplicating the regex inline.
- Rename `publish-pypi` to `publish-testpypi` (unchanged behavior).
- Add a new `publish-pypi` job that publishes to production PyPI via [trusted publishing](https://docs.pypi.org/trusted-publishers/) (OIDC), gated by the `pypi-release` GitHub environment (requires manual approval), and only runs for final release tags (`vX.Y[.Z]`, no `latest`/pre-release suffix).

## Requires (outside this PR, in repo/environment settings)

- A PyPI trusted publisher configured for this repo referencing workflow `Package.yml` and environment `pypi-release`.
- The `pypi-release` GitHub environment created with required reviewers, and its own `PYPI_REPOSITORY_URL`/`ENVIRONMENT_URL` variables set to the real PyPI endpoint (environment-scoped vars override repo-level ones with the same name, so these must differ from `pypi-test`'s values).

## Test plan

- [ ] Push a PR touching this workflow and confirm a second push cancels the first run.
- [ ] Verify workflow YAML lints cleanly (e.g. actionlint).
- [ ] Tag a pre-release and confirm `publish-pypi` (production) is skipped while `publish-testpypi` still runs.
- [ ] Tag a final release and confirm `publish-pypi` pauses for environment approval, then succeeds after approval.

### DIFF
--- a/.github/workflows/Package.yml
+++ b/.github/workflows/Package.yml
@@ -14,7 +14,9 @@ on:
   workflow_dispatch:
 
 concurrency:
-  group: '${{ github.workflow }}@${{ github.event.pull_request.head.sha || github.head_ref || github.ref }}'
+  # PR number is stable across pushes so superseded PR runs get cancelled; tag/release runs never are.
+  group: '${{ github.workflow }}@${{ github.event.pull_request.number || github.ref }}'
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
 
 # Set minimal permissions for all jobs (read-only)
 permissions:
--- a/.github/workflows/Package.yml
+++ b/.github/workflows/Package.yml
@@ -429,7 +429,7 @@ jobs:
             gh release edit "${TAG_NAME}" --draft=false
           fi
 
-  publish-pypi:
+  publish-testpypi:
     if: github.event_name != 'pull_request' && startsWith(github.ref_name, 'v')
     environment:
       name: pypi-test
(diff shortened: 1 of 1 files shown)


.github/workflows/Package.yml
```diff
@@ -288,6 +290,34 @@ jobs:
...
+      is_prerelease: ${{ steps.check.outputs.is_prerelease }}
+    steps:
+      - name: Check release tag type
+        id: check
+        shell: bash
+        env:
+          REF_NAME: ${{ github.ref_name }}
+        run: |
+          if [[ "${REF_NAME}" == "latest" ||
+              "${REF_NAME}" =~ ^v[0-9]+\.[0-9]+(\.[0-9]+)?((-.)?[a-zA-Z]+[0-9]*)$ ]]; then
```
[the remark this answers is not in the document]
blowekamp: done

dave3d (approved):

.github/workflows/Package.yml
```diff
@@ -288,6 +290,34 @@ jobs:
...
+        id: check
+        shell: bash
+        env:
+          REF_NAME: ${{ github.ref_name }}
+        run: |
+          if [[ "${REF_NAME}" == "latest" ||
+              "${REF_NAME}" =~ ^v[0-9]+\.[0-9]+(\.[0-9]+)?([-.]?[a-zA-Z]+[0-9]*)$ ]]; then
+            is_prerelease=true
+            is_final=false
+          elif [[ "${REF_NAME}" =~ ^v[0-9]+\.[0-9]+(\.[0-9]+)?$ ]]; then
```
zivy, line 310: This will not match a tag like [v2.1.1.1](https://github.com/SimpleITK/SimpleITK/releases/tag/v2.1.1.1). Possibly let it get through without this check. Anything that ends with letters and numbers is a prerelease and everything else is an official release?

blowekamp: We shouldn't be using the fourth component any more. It discouraged by a couple packagers before. A draft of the release process is here: #2716

PR #2715, opened 2026-09-04
Update swig 4.5.1

blowekamp:
Update SWIG to 4.5.1 which include fix for ruby wrapping related to auto-rename.


blowekamp:
Awaiting update of pypi SWIG to complete.

PR #2720, opened 2026-09-11
ENH: Add ImageFileReader::GetCanStreamRead

### LINKED ISSUE #2709: Expose ImageIOBase::CanStreamRead() / CanStreamWrite() on ImageFileReader and ImageFileWriter
`ImageFileReader.SetExtractIndex/SetExtractSize` silently degrades to a whole-volume decode when the underlying ImageIO cannot stream, and there is no way to ask beforehand. ITK knows the answer (`ImageIOBase::CanStreamRead()`), but SimpleITK does not expose it.

For an application that reads large volumes region by region this matters twice: performance (a per-region whole-volume decode is easily 100x the intended cost, with no error and no signal) and memory planning (we size buffers against what a read is supposed to cost, and a silent whole decode breaks the plan).

Our current workaround is a hand-maintained mirror of the streaming IO list: sniff the MetaImage header for compression, check NIfTI gzip magic, and assume everything else does not stream. That list drifts with ITK releases by construction.

Would you take a PR wrapping `CanStreamRead()` (and `CanStreamWrite()` on the writer) on `ImageFileReader`/`ImageFileWriter`, in the way `GetImageIOFromFileName` is already exposed? After `ReadImageInformation()` the reader holds the IO, so a bool method seems natural.

blowekamp:
## Summary
- Adds `ImageFileReader::GetCanStreamRead()`, exposing `itk::ImageIOBase::CanStreamRead()` so callers can check up front whether a file can be read as a sub-region without decoding the whole image, instead of relying on a hand-maintained, drift-prone list of streamable formats.
- Sets the flag inside `UpdateImageInformationFromImageIO()` so it is correct after either `ReadImageInformation()` or a direct `Execute()` call.
- Adds `CanStreamRead` to `ImageFileReader::ToString()` for consistency with the other image-information fields.

Closes #2709.

## Test plan
- Added `TEST(IO, ImageFileReader_GetCanStreamRead)` in `Testing/Unit/sitkImageIOTests.cxx`, table-driven over the formats currently known to support streamed reading (MetaImageIO, VTKImageIO, HDF5ImageIO, MRCImageIO), verifying both the `ReadImageInformation()` and `Execute()` code paths. Formats that currently do not support streaming are intentionally not asserted `false`, so that ITK adding streaming support for them later doesn't break this test.
- Also covers the default-constructed-reader case and MetaImageIO's compressed-vs-uncompressed behavior.
- Built `SimpleITKUnitTestDriver0` and ran the `IO.*` suite locally — all 21 tests pass.


zivy (approved):
Naming with the "Get" prefix is a bit strange in this case (assuming this is intentional to be consistent with the rest of SimpleITK). I guess we accept the "weirdness" of the method name GetCanStreamRead`.

blowekamp:
Good catch — renamed `GetCanStreamRead()` to `CanStreamRead()` in cfe7ee8a9.

`itk::ImageIOBase::CanStreamRead()` itself is bare (no `Get`), and SimpleITK's own convention reserves the `Get`/`Set` pair for settable properties (`GetUseCompression()`/`SetUseCompression()`, `GetDebug()`/`SetDebug()`, etc.). Read-only boolean queries with no setter already drop the prefix elsewhere in this same class, e.g. `HasMetaDataKey()`. Since there's no `SetCanStreamRead`, the bare form is the better fit, and there's no back-compat concern since this hasn't shipped in a release yet.

PR #2721, opened 2026-09-11
ENH: Add Transform::ApplyToImageMetadata

blowekamp:
## Summary
- Wraps ITK's `itk::Transform::ApplyToImageMetadata` as `Transform::ApplyToImageMetadata(Image &)` in SimpleITK.
- Updates an image's origin, spacing, and direction in place (mapped through the inverse of the transform), without resampling pixel data. Throws if the transform has no inverse or its dimension does not match the image's.
- Follows the existing pimpl dispatch pattern used by `TransformPoint`/`TransformVector` (`PimpleTransformBase` virtual method implemented in `PimpleTransform<TTransformType>`).

## Test plan
- [x] Added `TransformTest.ApplyToImageMetadata` in `Testing/Unit/sitkTransformTests.cxx`, covering a `ScaleTransform` (origin + spacing change), a `TranslationTransform` (origin-only change), and a dimension-mismatch exception case.
- [x] Built `SimpleITKCommon` and `SimpleITKUnitTestDriver0` locally; full `TransformTest.*` suite (40/40) passes.


Testing/Unit/sitkTransformTests.cxx
```diff
@@ -601,6 +602,100 @@ TEST(TransformTest, TransformVector)
...
+
+  tx6.ApplyToImageMetadata(img6);
+
+  EXPECT_EQ(img6.GetOrigin(), v3(-1.0, 2.0, -3.0));
+  EXPECT_EQ(img6.GetSpacing(), v3(2.0, 3.0, 4.0));
+  EXPECT_EQ(img6.GetDirection(), v9(1.0, 0.0, 0.0, 0.0, 1.0, 0.0, 0.0, 0.0, 1.0));
+
+  // Transform's dimension must match the image's dimension.
+  sitk::Image img3 = sitk::Image(10, 10, 10, sitk::sitkUInt8);
+  EXPECT_ANY_THROW(tx.ApplyToImageMetadata(img3));
```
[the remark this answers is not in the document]
blowekamp: Good catch, added. A singular (zero-scale) `AffineTransform` now confirms `ApplyToImageMetadata` throws when the transform has no inverse, in addition to the dimension-mismatch case. See commit 18a5f84bc.

Code/Common/include/sitkTransform.h
```diff
@@ -214,6 +214,22 @@ class SITKCommon_EXPORT Transform
...
 
+  /** \brief Apply the inverse of this transform to an image's origin, spacing and direction.
+   *
+   * This changes the image's origin, spacing, and direction such
+   * that physical points are mapped to the same voxel index as
+   * before, but as if the image itself had been transformed by this
+   * transform's inverse. This is a metadata-only operation, no
+   * resampling is performed and the pixel data is not modified.
+   *
+   * This transform must have an inverse and its dimension must match
```
zivy, line 225: Per [ITK's documentation](https://github.com/InsightSoftwareConsortium/ITK/blob/fd3b0b4ae9bdda33e878389b34364ad09ffeb1f2/Modules/Core/Transform/include/itkTransform.h#L547), please add the limitation that the method only works for linear transforms.

I am not sure why ITK [only issues a warning](https://github.com/InsightSoftwareConsortium/ITK/blob/fd3b0b4ae9bdda33e878389b34364ad09ffeb1f2/Modules/Core/Transform/include/itkTransform.hxx#L467-L471) for non linear transforms. Possibly because we can encode an affine transformation as a displacement transform in which case it should work. Still problematic and we need to point this to the users.

dave3d (approved):

Code/Common/include/sitkTransform.h
blowekamp: done

blowekamp:
@dzenanz FYI

dzenanz (approved):
Main additions look good. I have not looked at the test.

PR #2723, opened 2026-09-16
CI: Add reusable install-build-dependencies composite action

blowekamp:
## Summary
- Add `.github/actions/install-build-dependencies`, a composite action that consolidates the duplicated pip-install steps for build tooling (previously repeated inline across `Build.yml`, `BatchBuild.yml`, `Package.yml`, and `Nightly.yml`).
- The action takes an `extra-packages` input (space-separated pip specs) and an optional `install-pyproject-build-requires` input that installs the packages listed in `pyproject.toml`'s `[build-system] requires` (falling back from `tomllib` to `tomli` on Python < 3.11).
- `Build.yml` and `BatchBuild.yml` now use `install-pyproject-build-requires` instead of hardcoding `jinja2`/`jsonschema`/`pyyaml`/`swig` versions separately, so those versions only need to be maintained in `pyproject.toml`. This also fixes a drift where `Build.yml` pinned `swig~=4.4.0` while `BatchBuild.yml` and `pyproject.toml` pinned `swig~=4.5.0`.
- `Package.yml` and `Nightly.yml` are updated to call the new action for their `ninja`/`cmake` install but are otherwise behaviorally unchanged.
- Removed `.github/workflows/requirements-build.txt`, which is no longer referenced now that its contents are passed via the action's `extra-packages` input.

## Test plan
- [ ] CI runs green on `Build.yml`, `BatchBuild.yml`, `Package.yml`, and `Nightly.yml` for this branch
- [ ] Spot-check that `swig~=4.5.0` (bumped from `4.4.0`) does not break the `Build.yml` matrix


.github/actions/install-build-dependencies/action.yml
```diff
@@ -0,0 +1,40 @@
+name: 'Install Build Dependencies'
+description: 'Upgrade pip and install extra pip packages, and/or the [build-system] requires from pyproject.toml'
+inputs:
+  extra-packages:
+    description: 'Pip package specifiers to install (space-separated)'
+    required: false
+    default: ''
+  install-pyproject-build-requires:
+    description: 'Also install the packages listed in [build-system] requires of pyproject-path'
```
dave3d: I don't understand what this means.  'requires'?  Sames goes for the other descriptions further down.

blowekamp: 'requires' is the field under [build-system] in the yaml file. I'll see if I can clarify.

dave3d (approved):

